PR #76, opened 2015-01-07
Fix potential StackOverflowError retracting previous accumulated items

mrrodriguez:
Found a similar issue to what I discovered in both https://github.com/rbrush/clara-rules/pull/59 and https://github.com/rbrush/clara-rules/pull/74.
There is another use of `concat` that is subject to StackOverflowError.  This time it is in the `clara.rules.engine` namespace.  Previously I was only focusing in on the `clara.rules.memory` namespace for issues like this to fix.

I have fixed it the same way that I fixed the previous pull requests listed here.  I have changed the usage of the lazily evaluated `concat` to eagerly evaluated `into`.  Again I do not believe there are any unwanted side effects for going from lazy to eager in this scenario.

In `clara.rules.engine` the special accumulate node type `AccumulateWithJoinFilterNode` had a use of `concat` in the `right-activate-reduced` protocol function implementation.  In this function the previously accumulated "candidates", `previous-candidates`, are tracked in order to retract any tokens that were propagated due to their accumulation.

This issue is only exposed if there are many, nested `previous-candidates`.  It is easy enough to recreate by performing a single insert at a time to trigger this type of accumulate node to have many separate right activations.  Batch insertions will not likely be subject to this issue.  However, I've observed "in the wild" a stack overflow occurring for more obscure reasons due to the truth maintenance ultimately causing the accumulate node to accumulate in a non-batch-style manner.  So this can sneak up even when primarily using batch insertions.

### DIFF
--- a/src/main/clojure/clara/rules.clj
+++ b/src/main/clojure/clara/rules.clj
@@ -164,7 +164,7 @@
 
    If no sources are provided, it will attempt to load rules from the caller's namespace.
 
-   The caller may also specify keyword-style options at the end of the parameters. Currently two
+   The caller may also specify keyword-style options at the end of the parameters. Currently four
    options are supported:
 
    * :fact-type-fn, which must have a value of a function used to determine the logical type of a given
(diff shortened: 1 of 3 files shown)


rbrush:
Merged. Many thanks for tracking down another one of these, Mike.

PR #79, opened 2015-01-15
Token retractions may not occur when an accumulator reduces to nil

mrrodriguez:
This is a pull request to deal with what was discussed @ https://github.com/rbrush/clara-rules/issues/78.

- Put together a few test cases showing where previously reduced accumulations that return nil cause retractions of tokens to be missed
- Fixing this centers around disambiguating the cases of no-accum-reduced-ever vs nil-accum-reduced-value when calling clara.rules.memory/get-accum-reduced fn
- Added test case for both AccumulateNode and AccumulateWithJoinFilterNode to verify the issue is fixed

### DIFF
--- a/src/main/clojure/clara/rules/engine.clj
+++ b/src/main/clojure/clara/rules/engine.clj
@@ -568,13 +568,13 @@
 
       ;; If the accumulation result was previously calculated, retract it
       ;; from the children.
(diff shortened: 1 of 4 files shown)


rbrush:
Merged. Thanks for the fix!

PR #85, opened 2015-02-08
Proper truth maintenance when retracting equal facts

mrrodriguez:
This is a follow up to the issue discussed at https://github.com/rbrush/clara-rules/issues/84.  Facts in working memory with equal values had a potential
to not be properly retracted due to sets being used to group together facts to retract.

The biggest changes here are to the `clara.rules.memory/remove-first-of-each` function.  The former set of items to remove is now an arbitrary seq instead to account for duplicates.
The changes are a bit verbose in order to preserve the efficiency of this function via low-level looping and the use of transients.

----
I noted a few points to consider:

--
(1)

The most common usage of this function is in clara.rules.memory itself and it has the form:

```clj
filtered-elements (remove-first-of-each elements previous-elements)

```

Followed shortly by logic to return the removed elements.

This is doing redundant work and probably (previously at least) would be putting the previous-elements into a set repeatedly.  This could potentially have non-trivial overhead.  Also, to be safe on the consistency of duplicates items being removed and accurately reported, I thought it made sense to shift the responsibility of reporting what was removed to the `remove-first-of-each` function itself.

I have changed this function to return a tuple of `[items-removed coll-with-items-removed]` and I updated the docs.
I realize this makes the return value a little less "clean", but I think it is justified here.

--
(2) 

The return value from functions `remove-elements!` and `remove-tokens!` are a LazySeq with this implementation, where they used to be a set.  I looked around where they are being used at and I do not think that this detail matters.

--
(3)

I noticed that `remove-first-of-each` does not necessarily (not likely actually) preserve the order of the given coll since it rebuilds a seq from two other seq adding to the efficient end, which is the front.  I do not think this is an important detail, but if there I am curious if there is any advantage in preserving the order or not.  I'm thinking that it is not important at all, so I just left it alone for now.

For future reference if it comes up something like this could work:

```clj
(into (vec before-it) after-it)
```

### DIFF
--- a/src/main/clojure/clara/rules/engine.clj
+++ b/src/main/clojure/clara/rules/engine.clj
@@ -759,8 +759,7 @@
 
             ;; Compute the new version with the retracted information.
             :let [previous-result (do-accumulate accumulator join-filter-fn token previous-candidates)
-                  remove-set #{fact}
-                  new-result (do-accumulate accumulator join-filter-fn token (mem/remove-first-of-each remove-set previous-candidates))]]
+                  new-result (do-accumulate accumulator join-filter-fn token (second (mem/remove-first-of-each [fact] previous-candidates)))]]
 
       ;; Add our newly retracted information to our node.
       (mem/add-accum-reduced! memory node join-bindings new-result bindings)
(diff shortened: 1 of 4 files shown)


rbrush:
Merged. Thanks for yet another great contribution, Mike!

Issue #81, opened 2015-01-19
Working Memory Event Notifications


kahunamoore:
### Description

It would be helpful to have changes to working memory be announced to external listeners/code via an officially supported API in Clara.

### Requirements

1. Clara should provide notifications for changes to working memory.

1. Notifications should be symmetrical so that both insertions and retractions would trigger the notification(s.)

1. Notifications should be in the form of a collection (sequence or vector of facts), possibly only one, never empty.

### Example use case

Using Clara in CLJS for DOM related pattern matching. Rules/code can create DOM elements, insert them into working memory and pattern match against them. When necessary, DOM elements may be removed from working memory and, upon receiving a wm notification, removed from the DOM tree by a notification handler.

#### Background (from a Clara ML conversation)

kahunamoore wrote:

>I am needing to run some code whenever certain fact types are retracted (either explicitly or via truth maintenance.) I was reading through the code and found the :listeners option.

>What is the recommended usage for listeners? It seems the clojure defsession/mk-session macro/function passes along the options but the clojurescript versions do not AFAIK.

>I can see the clara.tools.tracing/with-tracing might be just the ticket I need - is this the suggested way to do tracing, or in my case, executing some code with side-effects whenever facts are inserted/removed? If so, do you have a small example that uses this feature?

rbrush wrote:

>Tracing and listeners are intended for debugging and troubleshooting rules and for truth maintenance. I don't think this is the mechanism we want to detect retractions in the long run.

>One enhancement that i had considered was the ability for callers to pass an on-retract function when doing an insertion from a rule. The function would simply be called by the truth maintenance engine if and when the support for that rule is retracted.

>Here is an (untested) example. If FireDetected matches, the rule triggers the alarm and inserts the corresponding fact. If FireDetected is retracted, the FireAlarm fact is retracted via truth maintenance, and the on-retract function is called to cancel the alarm.

>(defrule fire-alarm
 [FireDetected]
  =>
  (do
   (println "Starting the alarm!")
   (insert-facts! [(->FireAlarm)] {:on-retract (fn [facts]  (println "Stopping the alarm!"  ))})))

>If this type of functionality is interesting to you, please log an issue over on Clara's github site and we can follow up on it. I don't think it will be a huge effort to make this happen.

rbrush:
This sort of thing probably can be built into a specialized listener. To this point I've only used listeners for debugging and tracking the history of a session, but there really isn't a reason it couldn't be expanded for other uses. 

[...]

PR #92, opened 2015-02-24
Fix bug that caused options to be dropped in cljs

levand:
Also added a test case that fails on the old code, demonstrating that
the fix works.

### DIFF
--- a/src/main/clojurescript/clara/macros.clj
+++ b/src/main/clojurescript/clara/macros.clj
@@ -155,6 +155,7 @@ use it as follows:
 
     `(let [beta-network# ~beta-network
            alpha-nodes# ~alpha-nodes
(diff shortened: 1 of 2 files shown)


rbrush:
Merged. Thanks, Luke!

kahunamoore:
Sweet! I should now be able to use :listeners in cljs!

From issue #81:

"What is the recommended usage for listeners? It seems the clojure defsession/mk-session  macro/function passes along the options but the clojurescript versions do not AFAIK."

Maybe we can close #81 now that we have the :listeners option available in both clj/cljs.

rbrush:
If this meets your needs, Alan, then feel free to close #81. I'm not sure
if there is more you needed from it or not.

PR #94, opened 2015-02-26
Allow explicitly specifying productions in CLJS

levand:
Formerly, Clara/cljs could only load productions from a namespace. This
commit allows `defsession` to take an arbitrary expression, which is
evaluated *in clojure*, and may return a collection of productions that
will be used.

This can be useful in one of two modes:

- As part of a normal CLJS build process using (e.g.) lein-cljsbuild,
  rules can now be read from (for example) data files instead of
  ClojureScript source files.

- It is possible to set up a server which can compile and serve up new
  rule-containing JavaScript files *on demand*, generating sets of rules
  dynamically from (for example) a database.


rbrush:
Looks great, Luke! Merged.

PR #95, opened 2015-02-28
Support insert-all! in CLJS

levand:
It wasn't there for some reason; now it is.


rbrush:
Merged. On oversight on my part...I added this later on and didn't get it applied to the ClojureScript version. (This potential for sort of thing will go away once we jump to feature expressions or equivalent.)

PR #108, opened 2015-05-06
Clara DSL throws NPE for `..` macro symbol

mrrodriguez:
clara.rules.dsl mistakenly treats the `..` macro symbol as a Java/host constructor call, which causes a NPE to throw.

- Adding a special case to avoid the `..` macro symbol in the constructor call conditional


rbrush:
Merged. Thanks, Mike!

PR #109, opened 2015-05-13
Ensure compiler condition comparison nil-safe

mrrodriguez:
- Comparing arbitrary conditions in clara.rules.compiler (i.e. `gen-compare`) has potentially to throw NPE.  Implemented handling and reproduction case.


rbrush:
Merged. Thanks again!

PR #111, opened 2015-06-19
Conditional readers & cojure/clojurescript cohesion

lprefontaine:
Hi Ryan

here it is.

I did not tackle rules.clj/cljs yet. Will do so next week and integrate whatever comments you may have about this at the same time.

A+
Luc


src/main/clojure/clara/rules/engine/impl.cljc
```diff
@@ -0,0 +1,45 @@
+(ns clara.rules.engine.impl)
```
rbrush: It looks like this is all protocol definitions. A convention I've seen elsewhere is to simply name this namespace "clara.rules.engine.protocols", making it clear that's what it is. Putting protocols into their own namespace is a good pattern since it makes interactive reloading easier, so I like that these are separated out.

src/main/clojure/clara/rules/engine/helpers.cljc
```diff
@@ -0,0 +1,66 @@
...
+  (every? (partial contains? m) keys))
+
+(defn matches-some-facts?
+  "Returns true if the given token matches one or more of the given elements."
+  [token elements join-filter-fn condition]
+  (some (fn [{:keys [fact]}]
+          (join-filter-fn token fact (:env condition)))
+        elements))
+
+(defn send-accumulated
```
rbrush: This and other functions in this namespace are _only_ used in <snip>/engine/nodes/accumulators.cljc. This isn't so much of a reusable function as an implementation detail of the accumulator nodes. Therefore I think it might be best to simply keep these as private functions in that accumulators namespace so they remain a private implementation detail.

src/main/clojure/clara/rules/engine/impl.cljc
lprefontaine: Ok,

I use this convention mainly because most of the time there are more than just protocols.
Could not guessed that we would end up there :)

It's a global rename not a big deal to do. Will do it Wednesday.

src/main/clojure/clara/rules/engine/helpers.cljc
lprefontaine: The purpose of these seemed generic enough to put them in open space.

This was the criteria I used.

I can fold them back into the accumulators name space until the need arises later eventually.

rbrush: Understood. The main ones I think of as implementation details are send-accumulated and retract-accumulated, which are pretty specific to that namespace. I don't mind exposing things like has-keys here.

rbrush:
This looks great, Luc! I'll take some time to go through this in more detail but I think we're on the right track here.

lprefontaine:
You will have plenty of time until I can get my fingers back on my keyboard again :)
Just shut down Godjira this morning (I nicknamed my laptop :).

I'll still track your comments however while away and respond to it.
Wednesday I then should be able to push an update based on our discussions
and resume work on rules.clj/cljs plus adding ClojureScript tests.

Have a nice weekend,
Luc

rbrush:
I think the needs in this pull request have been satisfied in others over the last several months. I used some of the initial pass here in the initial transition to cljc reader conditionals and referenced this quite a bit for other changes, which was tremendously helpful. (Thank you!) Now that the cljc transition is complete, the APIs are largely unified, and everything seems to be working well.

Therefore I'm going to close this out, although if there is anything I'm missing we can create new pull requests and merge functionality in there. Thanks for your contributions in this space!

lprefontaine:
that's ok with me. 😂
Luc P.

PR #114, opened 2015-07-30
Keys in the :requires map of ns-info are symbols.

mischov:
Fix for issue 113: Aliased symbols not resolved in cljs.


rbrush:
Merged. Thanks for the contribution!

PR #115, opened 2015-07-30
Ancestors-fn support for clojurescript

infracanophile:
The ancestors-fn option isn't handled on clojurescript side.

The fix is very simple changes to create-get-alphas-fn and assemble-session as far as I can see.

PR to follow.

### DIFF
--- a/src/main/clojurescript/clara/rules.cljs
+++ b/src/main/clojurescript/clara/rules.cljs
@@ -8,7 +8,7 @@
 (defn- create-get-alphas-fn
   "Returns a function that given a sequence of facts,
   returns a map associating alpha nodes with the facts they accept."
-  [fact-type-fn merged-rules]
+  [fact-type-fn ancestors-fn merged-rules]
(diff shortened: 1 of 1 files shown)


rbrush:
Thanks for the PR! Assuming this is working in your project I'll go ahead and merge it. 

Normally I'd like to add unit tests, but we're pretty weak there on the ClojureScript side in general. Hopefully we can improve test coverage here as part of the jump to reader expressions in Clojure 1.7...but until then I'm willing to merge this as it is assuming it's gone through at least some test.

infracanophile:
I've tested it in my code and it works and is lightly tested but I'd understand wanting more thorough testing.

rbrush:
The change is simple enough and doesn't break anything on my end, so I'm happy to go ahead and merge it. We're going to take a pass through the ClojureScript code and try to port most of the Clojure tests over, at which point this will get covered.

Thanks for the contribution!

Issue #118, opened 2015-08-14
Right-hand side must be single expression


jonsmock:
I noticed this rule does not insert the B fact:

```
(defrule calculate-b
  [A]
  =>
  (println "inserting b")
  (clara/insert! (->B)))
```

However, this one does insert it:
```
(defrule calculate-b
  [A]
  =>
  (do (println "inserting b")
      (clara/insert! (->B))))
```

This could very well be intentional; I don't know what the implications are of allowing multiple s-expressions on the right-hand side. From a new-user perspective (as in, I am a brand new user, as of yesterday), I didn't expect this behavior. I wonder if either an exception could be thrown or if multiple s-expressions could be run.

I see there's already a warning about not including a right-hand side [here](https://github.com/rbrush/clara-rules/blob/master/src/main/clojure/clara/rules.clj#L218-L221). This seems like the place to throw for multiple rhs expressions, except that `split-lhs-rhs` only captures the first expression after the separator, [due to destructuring](https://github.com/rbrush/clara-rules/blob/master/src/main/clojure/clara/rules/dsl.clj#L24). Obviously, if you want to run all the s-expressions, you'd just have to destructure with `[lhs [sep & rhs]]` instead.

Anyway, just figured I'd give some feedback. Really liking things so far!

mrrodriguez:
I always thought it'd be more intuitive if Clara would just wrap the RHS in a "do" implicitly. 

I'm fairly sure there is no good reason to force the rule writer to have to wrap all their s expressions in their own "do" block. 

The error is sneaky too. I've been bitten by it before. I'm glad you reported it because I forgot about it. :)

rbrush:
I'm fine with implicitly wrapping the RHS in a do. I personally like the explicitly do block, but that's just a style preference and producing unexpected behavior like it does today is definitely a bug.

I'm going to be traveling over the weekend but can get it in early next week. It would be relatively straightforward to add if someone is interest in taking this on as well.

mrrodriguez:
I guess I was coming from the Drools background where it is just a sequence of statements on the RHS. If there was an explicit parse error thrown for having more than one form on the RHS, I'd still find that reasonable though.

rbrush:
Let's just add an implicit do around the RHS. It seems like the least
surprising behavior and I don't see a good reason not to.

jonsmock:
Thanks for the responses! I can take a shot at this, unless one of you would rather do it. I can probably get to it later this week?

[...]

PR #119, opened 2015-08-21
RHS captures all s-expressions using implicit `do`

jonsmock:
Regarding issue #118

### DIFF
--- a/src/main/clojure/clara/rules/dsl.clj
+++ b/src/main/clojure/clara/rules/dsl.clj
@@ -21,10 +21,11 @@
 (defn split-lhs-rhs
  "Given a rule with the =>, splits the left- and right-hand sides."
  [rule-body]
- (let [[lhs [sep rhs]] (split-with #(not (separator? %))  rule-body)]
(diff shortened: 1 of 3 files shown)


jonsmock:
This was the smallest change I could make to get the desired functionality. Probably better would be to simply return rhs as the captured collection, but doing this caused some problems in parse-rule*, so I backed it out. Let me know if you want me to make that change, and I can take another shot at it. I wanted to at least get this to you for review, and I just had my lunch break to do it.

rbrush:
Yes, I think the minimal approach is the way to go here. We could update parse-rule*, but that's used elsewhere in test code and in a (not publicly available) library that generates rules.

rbrush:
I went ahead and merged this. Thanks for the contribution! We can probably do a fix release with this (and a couple other recent fixes) if desired.

jonsmock:
No problem - thanks for the lib! No rush on the release. We're happily using it as is; I just figured I'd ask/fix it, if ya'll were interested. Seriously, rule engines rock. How did I not know this?! ;-)

PR #120, opened 2015-09-05
More general clara.rules.schema for s-expression forms

mrrodriguez:
When generating Clara rules a case was encountered where the s-expression predication of `list?` was too restrictive in clara.rules.schema.  This patch generalizes it to `seq?` to allow to accept forms such as a clojure.lang.Cons which can easily creep up in code forms generated with syntax quote in Clojure.

e.g.
```clj
(type `(= 1 2))
;= clojure.lang.Cons

(list? `(= 1 2))
;= false

(seq? `(= 1 2))
;= true
```


rbrush:
I went ahead and merged this. Thanks!

PR #124, opened 2015-09-10
enhanced checking of unbound variables to all types of conditions

mrrodriguez:
This is the patch for https://github.com/rbrush/clara-rules/issues/122.

Let me know if you have any issues with it.  It passes all the new and existing tests.  I ran it against my larger project and it seems to be working fine.  I even retested a "real world" example where this got me before and now the unbound variable is caught instead of ignored.

---

Notes:

I removed `clara.rules.compiler/is-equals?` and extracted this common sort of check on an expression to `c.r.c/equality-expression?`.

When doing this, I noticed the commit https://github.com/rbrush/clara-rules/commit/a1e3a729b4af15a1608c9ce05cbbc98dafa10b20 actually may have made the `is-equals?` check in `c.r.c/compile-constraints` irrelevant/handled elsewhere since it looks like `is-equals?` isn't actually called, it is just evaluated to the function which is always truthy.

I noticed that "late in the game", so for now I have reused the `equality-expression?` function correctly in its place instead to avoid any more changes with this PR.  We may want to circle back on this one, but I think it is functionally stable.

--
 
I made a test helper macro for testing clojure.lang.ExceptionInfo types via `ex-data`.  I often like this sort of testing of these data-centric exceptions rather than parsing strings.

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -8,6 +8,7 @@
             [clara.rules.listener :as listener]
             [clara.rules.platform :as platform]
             [clojure.string :as string]
+            [clojure.walk :as walk]
             [clara.rules.schema :as schema]
             [schema.core :as sc]
             [schema.macros :as sm])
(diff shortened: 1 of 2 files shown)


rbrush:
Looks good, Mike. We can follow up on changes to compile-constraints independently, so I went ahead and merged.

PR #126, opened 2015-09-23
produce correct dnf for consecutively nested conjunctions

mrrodriguez:
For a certain style of complex rule/query conditions it is possible to get an improperly formed Disjunctive Normal Form (DNF) created by the `clara.rules.compiler/to-dnf` function for the logical expression being compiled.

The issue I observed was that consecutively nested conjunctions when distributed over disjunctions were not properly combined/flattened when they should have been in the final DNF form.  I have added a test case exposing this issue in a "raw" form passed directly to the `c.r.c/to-dnf` function.  I also have a test demonstrating how this could manifest itself when compiling a multiple condition rule/query.

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -377,18 +377,22 @@
         (condp = (expr-type expression)
 
           :and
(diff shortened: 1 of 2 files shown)


rbrush:
Merged. Thanks!

PR #135, opened 2015-10-23
Small test fix for munged field names

mrrodriguez:
When doing https://github.com/rbrush/clara-rules/pull/121 I wrote a test that mistakenly didn't test what it was supposed to.  This just fixes that test to correctly run the assertion it was intended to.
The enhancement to allow "munged" field names to be used directly in rules still works as expected.

### DIFF
--- a/src/test/clojure/clara/test_rules.clj
+++ b/src/test/clojure/clara/test_rules.clj
@@ -2755,17 +2755,18 @@
                              (= ?x+y x+y)
                              (= ?bang! bang!)]])
 
-        s (-> (mk-session [q])
-              (insert ff)
-              fire-rules
-              (query q)
-              set)]
+        res (-> (mk-session [q])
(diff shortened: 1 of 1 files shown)


rbrush:
Haha, somehow I also overlooked that that comparison wasn't doing anything when reviewing the test. I see the fix and the tests pass, so I went ahead and merged it.

mrrodriguez:
Thanks!
@WilliamParker pointed that out to me because I didn't notice it before either.

WilliamParker:
FYI I found it because I use a Leiningen plugin that causes = to blow up when it is the first argument to the clojure.test is macro.

https://github.com/venantius/ultra/blob/d4b8ae0e0a499842ab77a84f607666b49d0f2517/src/ultra/test.clj#L21

The assert-expr multimethod is then called in clojure.test.

PR #139, opened 2015-10-28
Issue 134: Add ability to find seqs of rules when using namespaces as…

WilliamParker:
… rule sources


rbrush:
Merged. Thanks for the contribution!

PR #145, opened 2015-12-06
Only compile/eval unique beta node expressions

mrrodriguez:
I discovered that the truth maintenance can fail to produce predicted working memory states along certain control paths.  In this PR I have a test that exposes this issue in as small of a reproduction as I could devise right now.

This issue shows up here because the the use of an `:or` disjunction in the example test rule.  The nested conditions involved in the disjunction both have a common child ProductionNode.  When ProductionNodes have their activations added and retracted into Clara memory, the production Activation includes the entire ProductionNode under the `:node` key.  This means for one Activation to be considered equal to another Activation in memory the representation of the ProductionNode must be value-based equals (i.e. via `=`).

The problem is that ProductionNodes have a key, `:rhs`, that holds a compiled Clojure function object.  In general, it is likely a good idea that the Clara compiler should avoid compiling expressions that are equal more than one time.  For one, this has a potential benefit for both time and memory consumption.  More importantly, this is necessary behavior when the compiled Clojure function is associated to a Rete node that is used in its equality comparisons later (i.e. ProductionNodes).

What happens in the test scenario is that two nested, "sibling" conditions of the disjunction condition both have a reference to a ProductionNode that is actually the same `:id` and same rule LHS and RHS.  However, the function `clara.rules.compiler/compile-beta-tree` does not avoid compiling equalivalent expressions multiple times.  This PR fixes this specifically.

I have changed `c.r.c/compile-beta-tree` to maintain local state for "memoizing" the expression compilation (done via `eval`) calls when recursively compiling beta nodes.  This does not only fix the issue with ProductionNodes.  I have covered all compiled expressions in this process.  I have only found an issue in ProductionNodes though because of the way that an Activation refers to them and is keyed off of in Clara memory.
I do think it is a good idea to avoid any extra compilation and a good idea to ensure all nodes are completely equal, via `=`, when they represent the same node `:id` (e.g. for AccumulateWithJoinFilterNode, ProductionNode, etc).

The code changes to `c.r.c/compile-beta-tree` effected the way the forms are nested and therefore made the diff harder to see.  I think this restructure makes the function public signature clearer though and adds this memoized beta node expression compilation.  In case it is hard to see, the parts I changed were:
* Nesting the inner function recursion target.
* The inner function for memoizing compiling beta node expressions.
* All former direct uses of `eval` have been put behind this memoized variant.
* The public signature of the function no longer has `is-root` or `parent-bindings` passed in because I don't think that any consuming functions would need to be aware of those details anyways.

I will note that we could probably take this a step further (later) to have `c.r.c/compile-beta-tree` just reuse entire beta nodes if it has already compiled before along a different branch of the network.  This would be a step beyond just sharing the compiled/`eval`ed expressions beta nodes reference.  I didn't do that here just because it was potentially more disruptive.
Also, there this same sort of work may be beneficial when compiling the alpha nodes in the network.

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -873,138 +873,148 @@
 
 (sc/defn compile-beta-tree
   "Compile the beta tree to the nodes used at runtime."
-  ([beta-nodes  :- [schema/BetaNode]
-    parent-bindings]
-     (compile-beta-tree beta-nodes parent-bindings false))
-  ([beta-nodes  :- [schema/BetaNode]
-    parent-bindings
-    is-root]
-     (vec
-      (for [beta-node beta-nodes
-            :let [{:keys [condition children id production query join-bindings]} beta-node
+  [beta-nodes  :- [schema/BetaNode]]
+  (let [;; A local, memoized function that ensures that the same expression of
+        ;; a given node :id is only compiled into a single function.
+        ;; This prevents redundant compilation and avoids having a Rete node
+        ;; :id that has had its expressions compiled into different
+        ;; compiled functions.
+        compile-expr (memoize (fn [id expr] (eval expr)))
+        
+        compile-beta-tree
+        (fn compile-beta-tree [beta-nodes parent-bindings is-root]
+          (vec
+           (for [beta-node beta-nodes
+                 :let [{:keys [condition children id production query join-bindings]} beta-node
 
-                  ;; If the condition is symbol, attempt to resolve the clas it belongs to.
-                  condition (if (symbol? condition)
-                              (.loadClass (clojure.lang.RT/makeClassLoader) (name condition))
-                              condition)
+                       ;; If the condition is symbol, attempt to resolve the clas it belongs to.
+                       condition (if (symbol? condition)
+                                   (.loadClass (clojure.lang.RT/makeClassLoader) (name condition))
+                                   condition)
 
-                  constraint-bindings (variables-as-keywords (:constraints condition))
+                       constraint-bindings (variables-as-keywords (:constraints condition))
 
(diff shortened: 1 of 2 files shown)


rbrush:
Wow, nice find. I'll take a bit of a closer look and plan on merging it today.

mrrodriguez:
There is a potential issue with the approach I have here.  I'd hold off on merging this.  I didn't account for metadata on or within the expressions that we are compiling.  This would cause metadata to be lost.  
I'm going to revert back to an earlier approach I had where I take the use the id along with the expression form itself.  I'll respond back before long with an update to this PR.

WilliamParker:
You can get a branch with the example at https://github.com/WilliamParker/clara-examples/tree/duplicate-productions too.

WilliamParker:
So to elaborate a bit on the issue: Basically, the issue is that when you just use equality of the unevaluated code you don't use reader literal metadata in the equality comparison since it is evaluated in the reader phase prior to the compilation phase started by eval.  Thus, another block of unevaluated code with different metadata in internal forms would hit a cached entry from a prior equal but not identical code block.  

````
clara.examples=> (= `(conj ^{:a 1} [] 1) `(conj ^{:a 2} [] 1))
true
````

I discussed this some with @mrrodriguez ; we believe this can be solved by having the memoized compilation function cache on the node id as well as the expression.

mrrodriguez:
I've updated the PR to include the changes from the discussion above.

WilliamParker:
The updated changes look good to me.

rbrush:
The changes look good to me. If there is nothing else you want to do here, I'll go ahead and merge it.

mrrodriguez:
It is good to go.  It looks like I was able to run it against my larger codebase too without an issue.

rbrush:
Merged. Thanks for finding and fixing this! Looks like a tough edge case to track down.

PR #148, opened 2015-12-15
Use seq? instead of list? when looking for non-equality unifications

mrrodriguez:
In `clara.rules.compiler/non-equality-unification?` the criteria to find complex non-equality unifications is too restrictive and can miss "complex" unifications.  The issue is that `list?` is used which specifically only works for clojure.lang.IPersistentList impl's.  Often enough, Clojure s-expressions are other data types that are not covered by `list?`.  It seems that the safest bet for this sort of test is to stick with `seq?`.  This doesn't come up necessarily when rules are created via the Clara DSL macros and functions, however, if other methods of rule generation are used, this is likely to happen.

I have added a test demonstrating how this compiler issue could cause what seems to be valid rules to not compile.  I have also changed the other occurrence of `list?` I found in an error checking scenario of clara.rules.dsl.

Note: This situation is related to what was fixed in https://github.com/rbrush/clara-rules/pull/120.

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -416,7 +416,7 @@
    indicating it can't be solved by simple unification."
   (let [found-complex (atom false)
         process-form (fn [form]
-                       (when (and (list? form)
+                       (when (and (seq? form)
                                   (not (equality-expression? form))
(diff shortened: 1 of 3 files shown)


rbrush:
Merged. Thanks for the fix!

PR #152, opened 2015-12-31
Improve efficiency by using the result of computation

ylando2:


ylando2:
I found a way to improve it.

Issue #153, opened 2016-01-01
Improve the efficiency of compiling the expressions.


ylando2:
Hello Ryan,

You have a function to-dnf, that "Convert a lhs expression to disjunctive normal form." the problem with this is that the normal form could get exponential large consider the expression:
(and (or a1 a2 a3) (or b1 b2 b3) (or c1 c2 c3) ...)
Will turn into:
(or (and a1 b1 c1) (and a1 b2 c1) (and a1 b3 c1)  (and a1 b1 c2) ...)
If you have n subexpression with m element in every group you will get n^m expression in the disjunctive normal form. The real reta network only needs n*m nodes:
a1,a2,a3 connect to b1,b2,b3 and b1,b2,b3 connect to c1,c2,c3.
There are two possible ways to improve the efficiency of compiling the expressions.
1) We can remove this function and create the reta network directly.
2) My pull request about "cartesian-join" turn the exponential expression return from to-dnf into polynomial expression by reusing the expressions. We can use memoization between expression and  node on the tree to reduce the computation.

Regards,
Yuval

rbrush:
Hey, thanks for looking into this.

The explosion of expressions created in to-dnf should be unified when building the rete network itself. ```to-dnf``` just focuses on the single goal of converting the expression to disjunctive normal form, but there is an additional step to unify those expressions when building the rete network. If you call clara.rules.compiler/to-beta-tree on a sequence of rules, it will return a tree structure representing the beta side of the network. As part of creating that structure it will merge matching expressions, eliminating redundancy. At least some of the explosion of expressions in ```to-dnf``` is removed there.

There are probably further opportunities to optimize the beta side of the network. However, I don't think this logic belongs in to-dnf, because we want to merge logic from separate rules as well. So if someone has separate rules of the form (and a1 b1 c1) (and a1 b2 c1), etc, then these would not be merged in to-dnf, but in an optimization of to-beta-tree (or a function that it calls, like add-to-beta-tree).

I often use to-beta-tree to inspect and debug the structure of the actual generated rete network. A test that calls to-beta-tree for some set of rules and demonstrates a surprising or poorly optimized structure is probably a good start to expose an optimization bug here. If you have an example test that shows a surprising beta tree and are able to post it, I'd be happy to look for ways we can improve it.

ylando2:
Maybe you optimize the beta tree, but the question is do you optimize the work to create the beta tree?
[...]

PR #154, opened 2016-01-01
Improve efficiency by using the result of computation

ylando2:


ylando2:
We previously create a tree structure to describe the dnf but miss the opportunity to connect to the same node from different location, In this way we can describe the exponential structure created by cartesian-join using polynomial number of nodes.
For example we previously turn:
(and (or a1 a2 a3) (or b1 b2 b3) (or c1 c2 c3)) into
(or (and a1 b1 c1) (and a2 b1 c1) ...) one (b1 c1) subexpression for a1,a2,a3.
We could instead connect a1,a2,a3 to (b1 c1) and improve the size and number of step we need to create dnf from exponential to polynomial.

mrrodriguez:
I think we are still having some open discussion/investigation on ways to improve the overall efficiency of worst-case exponential size increase issues.

I just want to chime in that I saw a very little change here that doesn't tackle that whole issue, but does make a big time improvement in `clara.rules.compiler/cartesian-join`.  I think the original PR I saw here hinted at this improvement I think it is a pretty quick win and I could make (if needed) a PR for at least this.

----

`c.r.c/cartesian-join` as implemented:

```clj
(defn- cartesian-join [lists lst]
  (if (seq lists)
    (let [[h & t] lists]
      (mapcat
       (fn [l]
         (map #(conj % l) (cartesian-join t lst))) ; Lots of redundant recursive calls here.
       h))
    [lst]))


;; Perf test it
(time (let [colls (partition 3 (range 66))
            c (range 33)]
        (cartesian-join colls c)
        ;; don't spam
        nil))

;; Has heap space issues - doesn't complete for me with 2gb heap max.

```
-----

Very little change in the interest of time/space wasted:

```clj
(defn- cartesian-join [lists lst]
  (if (seq lists)
    (let [[h & t] lists
          c (cartesian-join t lst)] ; Do the recursion once.
      (mapcat
       (fn [l]
         (map #(conj % l) c)) ; Reuse it each iteration.
       h))
    [lst]))

;; Perf test it
(time (let [colls (partition 3 (range 66))
            c (range 33)]
        (cartesian-join colls c)
        ;; don't spam
        nil))

"Elapsed time: 0.685425 msecs"
;= nil
```

-----

Explanation:

The issue that I noticed here was that the `cartesian-join` recursive call is being made in the body of a `map` nested within a `mapcat`.  However, this could be done a single time and used for all iterations.  This causes significantly more recursive calls and recalculations of the same information.

-----

Side note: 

I think Ryan mentioned it elsewhere, but I think the `cartesian-join` argument names are pretty confusing.  I was definitely confused what it was expecting until doing some prints around it.

rbrush:
I made an attempt to make this a bit easier to understand. The implementation below performs as well as Mike's benchmarks and hopefully is a bit more readable. (I tend to prefer using list comprehensions over nested mapcat and map operations since the flow and variable names seem easier to understand to me.)

I'll probably put this into head since it gives the perf boost Mike was benchmarking and puts us in a better place to continue to improve it. 

Of course, there is a significant opportunity to optimize things outside of this function. At least with this optimization in place I see a lot more time spent in building the beta tree itself rather than converting to disjunctive normal form so that seems like the next place to focus.

```clj
(defn- cartesian-join
  "Performs a cartesian join to distribute disjunctions for disjunctive normal form.,
  This distributing each disjunction across every other disjunction and also across each
  given conjunction. Returns a sequence where each element contains a sequence
  of conjunctions that can be used in rules."
  [disjunctions-to-distribute conjunctions]

  ;; For every disjuction, do a cartesian join to distribute it
  ;; across every other disjuction. We also must distributed it across
  ;; each conjunction
  (reduce
   (fn [distributed-disjunctions disjunction-to-distribute]

     (for [expression disjunction-to-distribute
           distributed-disjunction distributed-disjunctions]
       (conj distributed-disjunction expression)))

   ;; Start with our conjunctions to join to, since we must distribute
   ;; all disjunctions across these as well.
   [conjunctions]
   disjunctions-to-distribute))
```

rbrush:
I did a bit more profiling and with this in place we are spending the bulk of our time in:

1. Comparing and sorting the nodes to generate consistent node IDs across rule executions. I think this might be unnecessary since re-loading a persisted session should produce nodes in the same order, so we can drop this. (Keeping node IDs across sessions only matters anyway for storing and reloading session state, which I don't think is happening much.) We can probably remove this.

2. Building the rete network itself recursively works through and modifies the immutable network structure. (see code in clara.rules.compiler/to-beta-tree and add-to-beta-tree). This will require a bit more effort to optimize, but I think changing to-beta-tree to build a temporary, transient network structure then converting it to an immutable form in a single pass would significantly improve performance.

I think we can add the cartesian-join change and item 1 above pretty easily to a fix release for a quick performance gain if that will be helpful. Re-working the creation of the beta network itself will be more involved, so that seems worth splitting into another item.

Any thoughts are welcome, of course.

ylando2:
Your version of cartesian-join look good. I still think that if we want to remove the exponential part of to-beta-tree, we have two options:
1) Remove the to-dnf, build the beta tree directly from the rule conditions.
2) to-dnf will add a unique id (serial number) for every expression. When we build beta tree we will update a map between id to node in a tree structure that we build before.
If 1 is possible I prefer 1 because it cleaner.

rbrush:
@ylando2 I'm not opposed to removing to-dnf, but converting an arbitrarily complex, nested structure to a beta tree and identifying all commonality introduces a lot of edge cases that don't have clear solutions...and even if they did it's not clear we could handle them faster than building the disjunctive normal form. From what I can tell from the literature and other engines, conversion to disjunctive normal form appears to be very common, presumably for these reasons.

However, there are definitely ways we can optimize this. Between the cartesian-join optimization and another optimization on issue #153, creating the disjunctive normal form is more than 20 times faster for the very "exponential" rule you posted on that issue. I think the next step is to do something similar to your suggestion of a unique id/serial number for every expression when building the beta tree. We already have a unique id of a sort: the identity of the expression itself (in terms of Clojure's identical? function, or a reference comparison in Java). The to-dnf results may include thousands of expressions, but many of these are "identical?" so we could use that when converting it to a network.

Hopefully this seems reasonable! I'm going to track progress on this on #153 and am happy to kick around ideas there.

ylando2:
@rbrush correct me if I am wrong, but Java GC moves the pointers around. So you can check if two expressions are equal by comparing their reference, but you cannot use the reference as the key for a map because they must be constant.

rbrush:
@ylando2 Java can move physical memory locations, but I don't think Java references can change or it would break thinks like java.util.IdentityHashMap (or anyone holding onto a reference value, for that matter). Seems like that IdentityHashMap exists for this type of use case.

devn:
@rbrush I am confident you're correct on your second point regarding beta-tree to/add, and agree that it ought to be split as separate work. Based on some simpleton profiling a few weeks back, I recall seeing beta-related fns stick out in terms of performance. This is partly what spawned a brief discussion in Slack about caching. I simply could not stomach 30-60sec, and so we needed to get fancier with our caching. If that were closer to 5sec than 50sec, it would be easier to justify dropping some reasonable amount of caching complexity.

> "Keeping node IDs across sessions only matters anyway for storing and reloading session state, which I don't think is happening much."

Could you describe in a bit more detail an example scenario that will no longer be possible if node IDs across sessions are no longer kept around across sessions? I'm not clear on the tradeoffs between having and not having it.

mrrodriguez:
Do you happen to know which beta related functions you saw in profiling? I often find paying attention to the type of node it is where you have performance bottlenecks is useful and insightful. Also it is nice to understand if the performance is leaning more towards the (alpha) "insert" or (alpha) "retract" side of things. 

Most often for me, my performance has had a good amount of time within accumulate nodes for example. However there are other cases. Also do you look all the way down the stack and see if you have any reflection warnings or that sort of thing? 

I have been interested is the feasibility of serializing and deserializing whole session state. That is interesting to me for some potential use cases. However the deserialization would need to be pretty quick to be worth it. 

Anyways I don't have the context of your slack conversation. But just curious because I watch performance quite a bit in some of my production usage of Clara as wel. I plan to start listening in on that channel. Sounds like some good conversation is happening there! :P

mrrodriguez:
Or maybe I misread this and the issue is on the time it takes to build the rule session rather than time spent running rules within it.

rbrush:
Consistent node ids across sessions only matters if you are using
clara.rules.durability to serialize and deserialize sessions, since it uses
node ids internally. But I don't think anyone is using this, as the
namespace is marked as still experimental. I think we need to improve our
durability mechanism so this is a non issue, but that's independent of this
issue.

I would encourage taking the code from this branch for a spin in your use
case to see if it helps performance. It makes a huge difference in the
worst case scenario, and I suspect it would help many other cases as well.

rbrush:
A different version of this and a number of optimizations have been included as part of #153. Therefore I'm going to close this one, but we can open another for any further optimizations.

PR #155, opened 2016-01-05
Performance enhancements

mrrodriguez:
I was looking at performance across a few fairly large rulebases and large dataset I had to work with.  While profiling, I noticed a few easy to spot performance time sinks coming out of the reflection warnings in clara.rules.memory.  I also saw a few places where using `clara.rules.platform/tuned-group-by` would be preferable than Clojure's built-in `group-by`.


rbrush:
Looks good to me. Merged and thanks!

PR #158, opened 2016-01-14
Accumulate right-retract may fail to update node memory

mrrodriguez:
## The problem

The below should not produce a query result binding for Cold facts:

```clj
(let [accum-q (dsl/parse-query []
                               [[Temperature (= ?t temperature)]
                                ;; Using simple AccumulateNode
                                [?cs <- (acc/all) :from [Cold (= ?t temperature)]]])
      accum-join-filter-q (dsl/parse-query []
                                           [[Temperature (= ?t temperature)]
                                            ;; Using AccumulateWithJoinFilterNode
                                            [?cs <- (acc/all) :from [Cold (<= temperature ?t)]]])
      c10 (->Cold 10)
      t10 (->Temperature 10 "MCI")
      s1 (mk-session [accum-q])
      s2 (mk-session [accum-join-filter-q])
      validate-query (fn [s q msg]
                       (is (= #{{:?t 10 :?cs []}}
                              (-> s
                                  (insert c10)
                                  (retract c10)
                                  (insert t10)
                                  fire-rules
                                  (query q)
                                  set))
                           msg))]
  (validate-query s1
                  accum-q
                  "AccumulateNode")
  (validate-query s2
                  accum-join-filter-q
                  "AccumulateWithJoinFilterNode"))
```

However, it fails with query results like:

```clj
({:?t 10, :?cs [#clara.rules.testfacts.Cold{:temperature 10}]})
```

Cold fact with `:temperature` 10 was retracted prior to inserting the join matching Temperature and running the query.

## The reason

When an AccumulateNode or AccumulateWithJoinNode have `clara.rules.engine/right-retract` called, the retraction only updates the accumulated value stored in node memory (or candidates to accumulate in the join filter case) if there are existing tokens that have been stored in the accumulate node memory under the same join bindings + bindings.  When there are no previous tokens, neither accumulate node type updates the previously accumulated value in node memory.  If a `left-activate` happens after the `right-retract` on an accumulate node, tokens will propagate with a potentially invalid previously accumulated value.

The changes in this PR fix this issue by ensuring `right-retract` always updates the previously accumulated value in node memory with the new value after the retracted elements have been removed.

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -762,14 +762,14 @@
       (if (not= :clara.rules.memory/no-accum-reduced previous)
 
         ;; A previous value was reduced, so we need to retract it.
-        (doseq [token (mem/get-tokens memory node join-bindings)]
+        (doseq [token matched-tokens]
           (retract-accumulated node accum-condition accumulator result-binding token
                                previous bindings transport memory listener))
 
         ;; The accumulator has an initial value that is effectively the previous result when
         ;; there was no previous item, so retract it.
         (when-let [initial-value (:initial-value accumulator)]
-          (doseq [token (mem/get-tokens memory node join-bindings)]
+          (doseq [token matched-tokens]
             (retract-accumulated node accum-condition accumulator result-binding token initial-value
                                  {result-binding initial-value} transport memory listener))))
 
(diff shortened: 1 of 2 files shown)


rbrush:
Merged. Many thanks for tracking this down.

PR #159, opened 2016-01-15
Batch right-retracts in accumulate nodes

mrrodriguez:
I have identified what looks like a quick and easy performance optimization for Clara accumulator node retraction logic.
As implemented, I think that `clara.rules.engine/right-retract`s are not retracting facts in a batch/collection-based style.  This is in contrast to how the `c.r.e/right-activate` does perform accumulation logic in batches.

I'd like to have stronger benchmarks for some future changes, but I think for this one I can make a pretty brute-force example to demonstrate the speed to be gained.

Before change:

```clj
(let [r1 (dsl/parse-rule [[?ts <- (acc/all) :from [Temperature]]]
                         (insert! (->TemperatureHistory ?ts)))
      r2 (dsl/parse-rule [[TemperatureHistory (= ?ts temperatures)]]
                         (insert-all! (map (comp ->Cold :temperature) ?ts)))
      q (dsl/parse-query []
                         [[?cs <- (acc/all) :from [Cold]]])

      n 10000
      temps (vec (for [i (range n)]
                   (->Temperature i "MCI")))

      s (-> (mk-session [r1 r2 q])
            (insert-all temps)
            fire-rules)]
  (is (= n
         (-> (query s q) set first :?cs count)))

  (time (fire-rules (apply retract s temps))))



"Elapsed time: 46181.982236 msecs"
#object[clara.rules.engine.LocalSession 0x3adee9be "clara.rules.engine.LocalSession@3adee9be"]
```

Same thing after:

```clj
"Elapsed time: 24143.068935 msecs"
#object[clara.rules.engine.LocalSession 0x67c327b9 "clara.rules.engine.LocalSession@67c327b9"]
```

So this trimmed 20 seconds off the time.  There are some other notable spots where time is spent now when I look at the before vs after profiling results, but I'd like to address each concern separately where appropriate.

I have ran these several times repeatedly in the same process as well to ensure there was some "JVM warmup" done.  The results seem consistent.

The PR here is very simple.  It just ensures that we do retraction logic in batch for clara.rules.engine.AccumulateNode and c.r.e.AccumulateWithJoinNode.

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -800,14 +800,15 @@
   (right-retract [node join-bindings elements memory transport listener]
 
     (doseq [:let [matched-tokens (mem/get-tokens memory node join-bindings)]
-            {:keys [fact bindings] :as element} elements
+            [bindings elements] (platform/tuned-group-by :bindings elements)
             :let [previous (mem/get-accum-reduced memory node join-bindings bindings)]
 
             ;; No need to retract anything if there was no previous item.
             :when (not= :clara.rules.memory/no-accum-reduced previous)
 
             ;; Compute the new version with the retracted information.
-            :let [retracted ((:retract-fn accumulator) previous fact)]]
+            :let [facts (mapv :fact elements)
+                  retracted (r/reduce (:retract-fn accumulator) previous facts)]]
 
       ;; Add our newly retracted information to our node.
       (mem/add-accum-reduced! memory node join-bindings retracted bindings)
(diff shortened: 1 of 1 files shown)


rbrush:
Love it! Merged and thanks!

As for future benchmarks, it might be worth adding them to the clara-benchmark project at https://github.com/rbrush/clara-benchmark. I had put that together when doing some initial profiling and comparison to Drools, but it makes sense to add more to it as we find things. It uses the criterium library to warm up the JVM, which I've found to be consistent and reliable.

mrrodriguez:
I messed around with clara-benchmark a while back, so I am a little familiar with it.  I will look into getting something set up there that demonstrates this accumulator path.  I have 1 or 2 other accumulator optimizations ideas I'm working through, so it would be good to have those laid out there.  

+1 to criterium.  It seems to be a commonly used lib for this sort of thing.

PR #160, opened 2016-01-16
Remove potential reflection

mrrodriguez:
I have noticed some new reflection warnings creeping up from the work done with https://github.com/rbrush/clara-rules/issues/149 where `clara.rules.compiler/extract-negations` was added.

The issue is that Clara rules conditions schema for clara.rules.schema.FactType `:type` is open to be more than just Class instances https://github.com/rbrush/clara-rules/blob/0.9.2/src/main/clojure/clara/rules/schema.clj#L29 .  The compiler appropriately allows the type of `:type` to not be important.  However, when the type is a Class instance, it is used as `:tag` metadata so that Clojure doesn't need to do reflection when accessing fields of objects in rule conditions.

When `extract-negations` was added, it used a Symbol rather than a Class for the `:type` of the condition.  This could just be switched to a Class, but that only fixes this issue.  I think a more general approach is to just extend `c.r.c/add-meta` to try to resolve Symbols to a Class, and if they resolve to one, use that as `:tag` metadata.  
this is consistent with Clojure in many ways, since Clojure's compiler often blurs the line of whether `:tag`-type metadata is an actual Class instance, or still a Symbol that resolves to a Class.

With this PR applied, all reflection warnings I saw related to this went away.  I always build Clara with reflection warnings on.  Also, my projects that use Clara are no longer seeing the reflection warnings coming from the extracted negations anymore.

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -262,9 +262,19 @@
 (defn- add-meta
   "Helper function to add metadata."
   [fact-symbol fact-type]
-  (if (class? fact-type)
-    (vary-meta fact-symbol assoc :tag (symbol (.getName ^Class fact-type)))
-    fact-symbol))
+  (let [fact-type (if (symbol? fact-type)
+                    (try
+                      (resolve fact-type)
+                      (catch Exception e
+                        ;; We shouldn't have to worry about exceptions being thrown here according
+                        ;; to `resolve`s docs.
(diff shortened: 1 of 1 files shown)


rbrush:
Merged. I'll probably also turn on reflection warnings by default on Clara so these will be more obvious.

mrrodriguez:
Yeah, I like to have reflection warnings on.

I put this in my .lein/profiles.clj
```clj
:global-vars {*warn-on-reflection* true}
```

Perhaps it would be a good idea to incorporate it into the Clara project.clj in some way.

PR #162, opened 2016-01-20
Issue 161: Add fact-to-rule map in clara.tools.inspect/inspect

WilliamParker:


src/main/clojure/clara/tools/inspect.clj
```diff
@@ -121,7 +124,17 @@
...
+
+     :fact->explanations (reduce (fn [previous [insertion info]]
+                                   (update previous insertion (fnil conj []) info))
+                                 {}
+                                 (mapcat (fn [[rule rule-node]]
+                                           (for [token (mem/get-tokens-all memory rule-node)
+                                                 insertion (mem/get-insertions memory rule-node token)]
+                                             [insertion {:rule rule
+                                                         :explanation (first (to-explanations session [token]))}]))
+                                         rule-to-nodes))}))
```
rbrush: Overall this looks good. Just one thing to consider: I'd be tempted to structure this a bit differently unless there are performance reasons not to. I think the above code could be accomplished like this:

```clj
     :fact->explanations (apply merge-with into
                                (for [[rule rule-node] rule-to-nodes
                                      token (mem/get-tokens-all memory rule-node)
                                      insertion (mem/get-insertions memory rule-node token)]

                                  {insertion [{:rule rule
                                               :explanation (first (to-explanations session [token]))}]}))
```

We can go with your implementation if this is too expensive or there is some other disadvantage. It just seems a bit more readable to me.

WilliamParker: I agree; I think this is cleaner.  It is probably a little slower but I doubt this is significant, particularly for a function that is primarily for debugging rather than routine use (although it is possible that others have a different use case).

I submitted a new pull request with the change at https://github.com/rbrush/clara-rules/pull/163

WilliamParker:
Closing in favor of https://github.com/rbrush/clara-rules/pull/163

PR #164, opened 2016-01-21
Add grouping-by accumulator

devn:
This PR adds a `grouping-by` accumulator.

Basic example:

```clojure
(defrecord Foo [num])

(def grouping-accum (clara.rules.accumulators/grouping-by :num))

(defquery q:grouped-foos []
  [?grouped-foos <- grouping-accum :from [Foo]])

(-> (mk-session [q:grouped-foos])
    (insert-all [(->Foo 1)
                 (->Foo 2)
                 (->Foo 3)])
    (query q:grouped-foos))

;; ({:?grouped-foos {1 [#clara.rules.testfacts.Foo{:num 1}], 2 [#clara.rules.testfacts.Foo{:num 2}], 3 [#clara.rules.testfacts.Foo{:num 3}]}})
```

An example of using the optional `convert-return-fn` argument:

```clojure
(defrecord Foo [num])

(defn max-foos
  [m]
  (if (seq m)
    (->> (keys m)
         (map #(Long/parseLong %))
         (apply max)
         (str)
         (get m))
    m))

(def grouping-accum (clara.rules.accumulators/grouping-by :num max-foos))

(defquery q:max-foos []
  [?max-foos <- grouping-accum :from [Foo]])

(-> (mk-session [q:max-foos])
    (insert-all [(->Foo "10")
                 (->Foo "20")
                 (->Foo "30")
                 (->Foo "30")
                 (->Foo "45")])
    (retract (->Foo "45"))
    (query q:max-foos))

;; ({:?max-foos [#clara.rules.testfacts.Foo{:num "30"} #clara.rules.testfacts.Foo{:num "30"}]})
```

### DIFF
--- a/src/main/clojure/clara/rules/accumulators.cljc
+++ b/src/main/clojure/clara/rules/accumulators.cljc
@@ -107,6 +107,27 @@
        :combine-fn wrapped-combine-fn
        :convert-return-fn wrapped-convert-return-fn}))))
 
+(let [grouping-fn (fnil conj [])
+      combine-fn (fn [a b]
+                   (merge-with (comp vec into) a b))]
+  (defn grouping-by
+    "Return a generic grouping accumulator. Behaves like clojure.core/group-by.
+
+  * `field` - required - The field of a fact to group by.
+  * `convert-return-fn` - optional - Converts the resulting grouped
+  data. Defaults to clojure.core/identity."
+    ([field]
+     (grouping-by field identity))
+    ([field convert-return-fn]
+     {:pre [(ifn? convert-return-fn)]}
+     (reduce-to-accum
+      (fn [m x]
+        (let [v (field x)]
+          (update m v grouping-fn x)))
+      {}
+      convert-return-fn
+      combine-fn))))
+
 (defn- comparison-based
   "Creates a comparison-based result such as min or max"
   [field comparator returns-fact supports-retract]
--- a/src/test/clojure/clara/test_accumulators.clj
+++ b/src/test/clojure/clara/test_accumulators.clj
@@ -284,3 +284,49 @@
     (is (= 1 (count temp-history)))
     (is (= [{:?his (->TemperatureHistory [temp-10-mci])}]
             temp-history))))
+
+(deftest test-grouping-accum
+  (let [grouping-accum (acc/grouping-by :temperature)
+        grouping-query (dsl/parse-query [] [[?t <- grouping-accum
+                                             :from [Temperature]]])
+
+        convert-return-fn (fn [m]
+                            (if (empty? m)
+                              []
+                              (val (apply max-key key m))))
+        grouping-convert-accum (acc/grouping-by :temperature
+                                                convert-return-fn)
+        grouping-convert-query (dsl/parse-query [] [[?t <- grouping-convert-accum
+                                                     :from [Temperature]]])
+
+        session (-> (mk-session [grouping-query
+                                 grouping-convert-query])
+                    (insert-all [(->Temperature 30 "MCI")
+                                 (->Temperature 10 "MCI")
+                                 (->Temperature 80 "MCI")
+                                 (->Temperature 80 "MCI")]))
+
+        retracted-session (-> session
+                              (retract (->Temperature 80 "MCI")
+                                       (->Temperature 80 "MCI")))]
+
+    (testing "grouping-accum"
+      (is (= {:?t {30 [(->Temperature 30 "MCI")]
+                   10 [(->Temperature 10 "MCI")]
+                   80 [(->Temperature 80 "MCI")
+                       (->Temperature 80 "MCI")]}}
+             (first (query session grouping-query)))))
+
+    (testing "grouping-accum with retraction"
+      (is (= {:?t {30 [(->Temperature 30 "MCI")]
+                   10 [(->Temperature 10 "MCI")]}}
+             (first (query retracted-session grouping-query)))))
+
+    (testing "grouping-accum with custom convert-return-fn (max temp group)"
+      (is (= {:?t [(->Temperature 80 "MCI")
+                   (->Temperature 80 "MCI")]}
+             (first (query session grouping-convert-query)))))
+
+    (testing "grouping-accum with custom convert-return-fn and retraction (max temp group)"
+      (is (= {:?t [(->Temperature 30 "MCI")]}
+             (first (query retracted-session grouping-convert-query)))))))


rbrush:
Looks great! I'm going to review it a bit more thoroughly tomorrow (Friday) but will probably merge it unless something pops up.

devn:
One thing I wanted to bring up is that this forces a user into using the grouping function: `(fnil conj [])`

It also forces a combining function of: `(fn [a b] (merge-with (comp vec into) a b))`

I wonder what you'd think of having a more general grouping accumulator generator:
```clojure
(defn group-accum
  "Return a generic grouping accumulator. It behaves like
  clojure.core/group-by with the exception of specifying the grouping
  step."
  [{:keys [value-fn grouping-fn combine-fn convert-return-fn]
    :or {value-fn identity
         grouping-fn (fnil conj [])
         combine-fn (fn [a b] (merge-with (comp vec into) a b))
         convert-return-fn identity}}]
  {:pre [(ifn? grouping-fn)]}
  (clara.rules.accumulators/reduce-to-accum
   (fn [m x]
     (let [v (value-fn x)]
       (update m v grouping-fn x)))
   {}
   convert-return-fn
   combine-fn))
```

Suppose for instance, you wanted to group, but use sets instead of vectors. So instead of `(fnil conj [])`, you could provide `(fnil conj #{})`, and instead of `(fn [a b] (merge-with (comp vec into) a b))`, you chose `(fn [a b] (merge-with (comp set into) a b))`.

An example of using this would be:

```clojure
(defn max-by-date  [current-value candidate-value]
  (if current-value
    (max-key (comp time.coerce/to-long :date) current-value candidate-value)
    candidate-value))

(defn latest-by
  "Return an accumulator which constructs a map of the return value of
  `f` to the most temporally maximal value passed to `f`."
  [f]
  (group-accum
   {:value-fn f
    :grouping-fn max-by-date}))
```

rbrush:
+1 for the more general accumulator. If the documentation would include a brief description of what each keyed parameter means, that would be nice as well. 

I like this addition a lot....it's the best way to express things like "max for each day" that I've seen yet.

devn:
Hey Ryan,

I've added the more general `grouping-accum`. Hopefully I didn't go _too_ crazy for your tastes on the docstring. It's sort of verging on a wiki page (which I am happy to author).

In any event, please let me know what you think!

Cheers,
Devin

devn:
A couple of things I'm interested in your opinion on specifically:

Does it make sense for `grouping-accum` to accept 0 arguments, or should it be required to supply a `:value-fn`? Thinking maybe `[value-fn & [{...}]]`...

Thoughts on the name `:value-fn`? I suspect that the vast majority of the time, users will pass `:foo` in, which makes it feel more like the other accumulator functions' `field` argument. It is, I suppose, interesting to know that it can be a more complex function, but I wonder if that would be better expressed in documentation.

In general, I'm trying to be sensitive to the 90% use case. It can be a little tricky to think through how all of these fn args behave. One thought might be to treat `grouping-accum` (or whatever we decide to call it) similar to `accum`, and use it to build a more opinionated `grouping-by`. My gut is that most people are going to be looking for a way to do something like you mentioned above: "max for most recent day". I wonder if there's a simple enough way of allowing someone to express "by coercing :date to longs, and grouping by that value, take the max key of the resulting map, and then return the fact from that collection with the max :cost value" (again with the option of supplying a function to coerce :cost values).

```clojure
(grouping-by [:date clj-time.coerce/to-long :max] [:cost sale/calculate-cost :max])
```

I'll stop there. Perhaps this is getting into DSL territory, but I'm curious to know if you or anyone else reading has any ideas for a very simple, but slightly more built-for-purpose grouping construct. If not, maybe what I'm describing is something that could/should live in a `clara.tools`.

rbrush:
Apologies for the slow response. I think we should keep things simple and target the 90% case here, and if there are more specialized needs users can write their own accumulators (or functions to create them).

In that spirit, I'd prefer to leave out coercion of the types...and come to think of it, allowing the user to specify the combine and grouping functions doesn't make a lot of sense to me (and adds complexity to the API), since grouping-by suggest the groups should just be sequences. 

+1 for making this look like the other APIs in clara.rules.accumulators, so perhaps something like this:

```clj
(defn grouping-by [field & {:keys [:convert-return-fn] :or {:convert-return-fn identity}}])
```

Where, like in the other APIs, the ```field``` parameter can be a keyword or a function that extracts a field from the given structure. This keeps thing simple and (hopefully) intuitive to new years, and is consistent with the other pre-defined accumulators.

mrrodriguez:
I think there is probably utility in having an accumulator for grouping in cases where you want to perform some calculation based on the group as an aggregate. 

I just want to mention that Clara already gives you support for "grouping" facts by their field-level bindings.  Along with this, you can do a transformation on these values as well, so that it ends up working like a `:value-fn` from the examples above:
```clj
(defrecord Foo [num])
(defrecord FooGroup [foos])

;; Basic grouping

(defrule group-foos 
  [?group <- (acc/all) :from [Foo (= ?num num)]]
  =>
  (insert! (->FooGroup ?group)))

(defquery find-groups []
  [?fg <- FooGroup])

(-> (mk-session [group-foos find-groups])
    (insert (->Foo "10")
            (->Foo "20")
            (->Foo "30")
            (->Foo "30")
            (->Foo "45"))
    (retract (->Foo "45"))
    fire-rules
    (query find-groups))

;;=
;; ({:?fg #clara.test_rules.FooGroup{:foos (#clara.test_rules.Foo{:num "30"} #clara.test_rules.Foo{:num "30"})}}
;;  {:?fg #clara.test_rules.FooGroup{:foos [#clara.test_rules.Foo{:num "20"}]}}
;;  {:?fg #clara.test_rules.FooGroup{:foos [#clara.test_rules.Foo{:num "10"}]}})

;; :value-fn style grouping

(defrule group-foos 
  [?group <- (acc/all) :from [Foo (= ?num (first num))]]
  =>
  (insert! (->FooGroup ?group)))

(defquery find-groups []
  [?fg <- FooGroup])

(-> (mk-session [group-foos find-groups])
    (insert (->Foo "10")
            (->Foo "20")
            (->Foo "30")
            (->Foo "30")
            (->Foo "35"))
    (retract (->Foo "35"))
    fire-rules
    (query find-groups))

;;=
;; ({:?fg #clara.test_rules.FooGroup{:foos [#clara.test_rules.Foo{:num "10"}]}}
;;  {:?fg #clara.test_rules.FooGroup{:foos [#clara.test_rules.Foo{:num "30"} #clara.test_rules.Foo{:num "30"} #clara.test_rules.Foo{:num "35"}]}}
;;  {:?fg #clara.test_rules.FooGroup{:foos [#clara.test_rules.Foo{:num "20"}]}})
```

So this has been the style of grouping I've typically needed.  Where I just want the rule/query conditions in a partitioned/grouped way.

I'm just mentioning this in case people weren't aware of this behavior.

devn:
@mrrodriguez Thanks for that. I would love to see this added to documentation somewhere.

I guess my only comment would be that there's an expressivity difference primarily because an intermediate record type is now required. I could see this being really helpful in some cases, and less so in others where the intermediate name of the grouping increases the distance between programmers and the domain being expressed.

I think it's worth adding a group-by accumulator either way, mostly based on the number of mailing list posts I've seen asking about similar functionality in Jess and Drools, but I don't know that I have a clear set of instructions for when one method of grouping is more idiomatic than another.

Either way, my apologies for letting this sit dormant. Hoping to find some time before EOD today with Ryan's suggestions applied.

mrrodriguez:
@devn I do not disagree that there are times when a grouping accumulator could be useful.

I think there are basically 2 major grouping flows I can come up with and the accumulator is simpler for one of them.

From my example above, you do not need a grouping accumulator really if all you are concerned with is getting a query result or firing a rule that is partitioned into groups.

However, if what you want is to group facts into some sort of partitions and then make further "decisions" on the groups in aggregate, it may be more convenient to have an accumulator so that you could group the facts and reason about all of the groups at once in a single rule.
You could accomplish this with 2 rules and an intermediate "grouped" fact, but that is more clunky and the accumulator is probably nicer there.

I can't speak for Jess much, but in Drools you tend to have to write a grouping accumulator for most all cases since variable bindings in Drools accumulators are only available within the accumulator "from" scope, so it new variable bindings there do not form partitions over the rule/query activations.  This feature in Clara can be a nice convenience when you want to activate a rule/query in a partitioned way.

I agree that this is not really documented in a way where it makes things obvious.  Perhaps once we get a grouping accumulator like this available, we can document its use-cases, but also mention how Clara's accumulators by default do have some grouping/partitioning behavior across the facts accumulated.

src/test/clojure/clara/test_accumulators.clj
```diff
@@ -284,3 +284,48 @@
...
     (is (= [{:?his (->TemperatureHistory [temp-10-mci])}]
             temp-history))))
+
+(deftest test-grouping-accum
+  (let [grouping-accum (acc/grouping-by :temperature)
+        grouping-query (dsl/parse-query [] [[?t <- grouping-accum
+                                             :from [Temperature]]])
+
+        convert-return-fn (fn [m]
+                            (if (seq m)
```
devn: This needs a guard otherwise it blows up. I find this kind of unfortunate. Curious to know what others think.

devn:
I made one comment on the code above, otherwise this should be good to go.

src/test/clojure/clara/test_accumulators.clj
rbrush: Looks like either the guard or a default value is needed here since max-key requires is not defined for zero arguments. We could change the structure a bit to make that more obvious (one option is shown in the following snippet), but I don't think it's a big difference.

```clj
        convert-return-fn (fn [m]
                            (if (empty? m)
                              ;; No max temperature to convert, so return an empty sequence.
                              []
                              (val (apply max-key key m))))
```

src/main/clojure/clara/rules/accumulators.cljc
```diff
@@ -107,6 +107,27 @@
...
        :convert-return-fn wrapped-convert-return-fn}))))
 
+(let [grouping-fn (fnil conj [])
+      combine-fn (fn [a b]
+                   (merge-with (comp vec into) a b))]
+  (defn grouping-by
+    "Return a generic grouping accumulator. Behaves like clojure.core/group-by.
+
+  * `field` - required - The field of a fact to group by.
+  * `:convert-return-fn` - optional - Converts the resulting grouped
```
rbrush: Extraneous colon in front of :convert-return-fn.

rbrush:
Looks good, and I'm happy to merge this unless there is anything else others would like to see here. (I had couple nits in the comments, but not enough to block this.)

src/test/clojure/clara/test_accumulators.clj
devn: I guess I was more curious about whether or not `convert-return-fn` could be a bit smarter and do this by default. Too much?

src/main/clojure/clara/rules/accumulators.cljc
devn: Fixed in d66476c79c76d9ab21254c721293e783e692b08a

src/test/clojure/clara/test_accumulators.clj
devn: Changed this to your snippet above in 32ebcb352c4ee5b74e5f85a67db121953dcfadc1

rbrush:
Looks great! 

Just one thing before merging: would you mind squashing these into a single commit?  I can squash and merge these myself (and don't mind doing so), but then you wouldn't get credit in the change history. 

This way this change shows up as a single commit in master, which keeps the history clean and readable. 

I really appreciate the contribution!

devn:
@rbrush Squashed.

rbrush:
Merged into master. Thanks, @devn!

PR #167, opened 2016-02-01
Issue 157: Protect against duplicate rules and cache sessions with th…

WilliamParker:
…e rules as the key, rather than the rule sources


rbrush:
Looks good, and merged to master. Thanks for doing this!

PR #168, opened 2016-02-02
providing compile context during compilation (take 2)

mrrodriguez:
A few error messages I have seen during testing newer changes in the current snapshot have lost one part of the changes from https://github.com/rbrush/clara-rules/pull/156.

The `:production` is not showing up when `clara.rules.compiler/try-eval` fails like it was intended to.  This was lost in the clara.rules.compiler rework done with https://github.com/rbrush/clara-rules/commit/f9537c7540e1a8baa5cf07f0f88b0116125d31f9.
I have added it back again and added a few tests to attempt to validate that this isn't overlooked again in the future as easily.


rbrush:
Merged. My mistake for omitting this during refactoring. Thanks for fixing it.

PR #175, opened 2016-03-04
Add listeners to the exception data map when the RHS throws an exception

WilliamParker:
I had a case recently when I wanted to look at the trace to debug some rules problems that caused an exception in the RHS of a rule and discovered that the trace leading up to the exception was unavailable.   Specifically, I was experimenting with a custom domain-specific salience sort function in our codebase to potentially improve performance, but we have a few rules (fortunately few) that depend on salience for functional correctness.  Correctness of some of the facts that these rules produce is enforced by rules that throw exceptions in the RHS if some logically nonsensical conditions occur.  To debug this I wanted to look at the order in which rules were firing.  I ended up making this change in order to debug my problem.  Side note: the problem turned out to be in my salience function; I didn't find anything incorrect in how Clara handles salience ordering.

This seems like something that would be generally useful to have available, and since this branch of the code would only run in the error case I don't see any performance concerns.  I also added a try-catch around the code to make the listener persistent and retrieve its now-persistent children in case a listener itself had a problem so that such an error wouldn't cause problems with generating other parts of the ExceptionInfo data map.

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -1108,24 +1108,31 @@
                       (flush-insertions! batched false))
                     (when-let [batched (seq @batched-rhs-retractions)]
                       (flush-rhs-retractions! batched))
-                    (catch #?(:clj Exception
-                                   :cljs :default) e
-
-                                   ;; If the rule fired an exception, help debugging by attaching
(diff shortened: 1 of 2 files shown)


rbrush:
Looks good to me! Merged, and thanks for another nice pull request.

PR #176, opened 2016-03-14
Issue 174: Assign higher salience to extracted complex negation rules…

WilliamParker:
… than their source rules

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -898,6 +898,8 @@
                            ;; Propagate properties like salience to the generated production.
                            (:props production) (assoc :props (:props production))
 
+                           true (assoc-in [:props :clara-rules/internal-salience] :extracted-negation)
+
                                ;; Propagate the the environment (such as local bindings) if applicable.
                            (:env production) (assoc :env (:env production)))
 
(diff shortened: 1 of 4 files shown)


src/main/clojure/clara/rules/compiler.clj
```diff
@@ -898,6 +898,8 @@
                            ;; Propagate properties like salience to the generated production.
                            (:props production) (assoc :props (:props production))
 
+                           true (assoc-in [:props :clara-rules/internal-salience] :extracted-negation)
```
rbrush: This won't block a merge, but for future reference it might be cleaner to put this property in the original map (the first parameter to cond->) rather than as an additional assoc.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1253,3 +1253,60 @@
...
+            forward-result)
+          (let [backward-result (user-activation-group-sort-fn (nth salience2 0)
+                                                               (nth salience1 0))
+                forward-bool (boolean forward-result)
+                backward-bool (boolean backward-result)]
+            ;; Since we just use Clojure functions, for example >, equality may be implied
+            ;; by returning false for comparisons in both directions rather than by returning 0.
+            ;; Furthermore, ClojureScript will use truthiness semantics rather than requiring a
+            ;; boolean (unlike Clojure), so we use the most permissive semantics between Clojure
+            ;; and ClojureScript.
```
rbrush: Thanks for putting these explanations in the code. Makes the code much easier to understand and maintain when the reasoning behind the logic isn't obvious.

rbrush:
Merged. Thanks!

PR #179, opened 2016-04-07
No auto qualifying vars in rule RHS

mrrodriguez:
This address the issue from https://github.com/rbrush/clara-rules/issues/178.

I didn't deal with issues of locals shadowing in LHS rule conditions right now.  They are theoretically possible, but probably much less common due to the fact that I'd not expect it to be "good practice" to do things like let-bindings in the constraints of LHS conditions.  We can address this later if needed, but qualified symbols do have some value in our static, "semantic" analysis done on LHS conditions within the Clara compiler.

If anyone is generated rules without using the clara.rules.dsl (we have cases) then the contract has not changed.  The generated rule RHS would have had to use "manually" qualified vars before and still will now.  The only addition is that the generated rules could also choose to set the `:ns-name` instead of qualifying the vars if they wanted the RHS form to be eval'ed in a specific `*ns*` valued binding.

I also added some fairly primitive support to use figwheel in Clara.  This looks to be the state-of-the-art in running REPLs in CLJS and it works pretty easily right out-of-the-box.  I'd still like to get it working to launch an nREPL server, but I'm leaving that as a task for another time.  I stumbled across CLJS due to the changes at clara.macros/gen-beta-network where I wanted to preserve the expected CLJS behavior around eval of the RHS.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /lib
 /classes
 /checkouts
+/resources/public/js/*
 pom.xml
 pom.xml.asc
 *.jar
@@ -10,4 +11,5 @@ pom.xml.asc
 .lein-failures
 .lein-plugins
 .lein-repl-history
+figwheel_server.log
 *.*~
--- a/project.clj
+++ b/project.clj
@@ -6,27 +6,34 @@
   :dependencies [[org.clojure/clojure "1.7.0"]
                  [org.clojure/clojurescript "1.7.170" :scope "provided"]
                  [prismatic/schema "1.0.1"]]
-  :plugins [[lein-codox "0.9.0"]
-            [lein-javadoc "0.2.0"]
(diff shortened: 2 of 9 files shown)


WilliamParker:
+1 to the locals shadowing changes (I didn't look at the Figwheel part much).

rbrush:
Merged! Thanks for doing this, it looks great.

(For future reference, normally I'd prefer independent things like fig wheel support to come in a separate commit/PR, but this was small enough that it wasn't worth going back to break up.)

mrrodriguez:
Yes.  Sorry for coupling in the figwheel.  I was just frustrated when I wanted to try to use the RELP some to troubleshoot the cljs tests when I had some failures.  It was a plugin so I piggy-backed it.  Next time I make a change of that nature though, I'll just separate it out.

Thanks for getting it looked over and merged so quickly.

PR #180, opened 2016-04-15
Only accessed used fields in constraints

mrrodriguez:
- The clara.rules.compiler should only access used fields of the fact type in in production condition constraints


rbrush:
Nice! Merged.

PR #185, opened 2016-05-02
Fix ClojureScript test build for whitespace optimizations

WilliamParker:
It looks like [1] changed the "simple" build for the ClojureScript tests and moved the file that this build depends on. [2] The build is now failing [3] when [4] fails to resolve the JavaScript variable clara.test.run [5].

This PR adds back the original simple build and moves the Figwheel build into a new ClojureScript build.  It also adds the source directory "src/test/common" to the advanced build source paths
to be the same as the simple build.  Lastly, if my understanding of the way ClojureScript
namespace forms is correct (which it might not be) it looks like [6] is relying on 
its transitive dependencies to require the ClojureScript version of cljs.test 
(as opposed to the .cljc version with macros). I went ahead and added an explicit require for the
cljs.test namespace there.

Ideally the ClojureScript tests failing like this would cause the main build to actually
fail instead of just printing failure messages to standard out.  I'll log a separate issue
for that; I don't see immediately see a quick fix for that.

1. https://github.com/rbrush/clara-rules/pull/179/files
2. https://github.com/rbrush/clara-rules/blob/0.11.1/src/test/html/simple.html#L3
3. https://travis-ci.org/rbrush/clara-rules/builds/125594217
Particularly
````
Running ClojureScript test: phantom-simple
ReferenceError: Can't find variable: clara
  phantomjs://webpage.evaluate():2
  phantomjs://webpage.evaluate():3
  phantomjs://webpage.evaluate():3
````
4. https://github.com/rbrush/clara-rules/blob/0.11.1/src/test/js/runner.js#L24
5. https://github.com/rbrush/clara-rules/blob/0.11.1/src/test/clojurescript/clara/test.cljs#L13
6. https://github.com/rbrush/clara-rules/blob/0.11.1/src/test/clojurescript/clara/test.cljs#L8

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -23,7 +23,7 @@
   :clean-targets ^{:protect false} ["resources/public/js" "target"]
   :hooks [leiningen.cljsbuild]
   :cljsbuild {:builds [;; Simple mode compilation for tests.
-                       {:id "simple"
+                       {:id "figwheel"
                         :source-paths ["src/test/clojurescript" "src/test/common"]
                         :figwheel true
                         :compiler {:main "clara.test"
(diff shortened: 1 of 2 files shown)


mrrodriguez:
+1
Nice catch that things were failing, but not showing up.

WilliamParker:
For some reason GitHub is rendering my footnotes as "1-2-3 1-2-3" even if I type them as "1-2-3-4-5-6".  I'm not sure why that is happening, but please consider the numbering to be the latter.

rbrush:
Looks good. Merged, and thanks!

PR #187, opened 2016-05-04
Mutable JVM colls in local memory for efficiency

mrrodriguez:
These changes are explained more in https://github.com/rbrush/clara-rules/issues/184.

High-level overview:
- Mutable Java collection types used in transient local memory when items retracted
- Memory collections remain persistent and immutable unless there is a need for items to be retracted from them
- Memory is converted to completely persistent, immutable clj structures when `clara.rules.memory/to-persistent!` is called
- No changes to memory that doesn't directly interact with `clara.rules.memory/remove-first-of-each` (our potential hot spot we are fighting here)
- Flush all updates during `clara.rules.engine/flush-updates` (See [1] for more)
-- Remove (now superfluous) `retract-facts!` call for empty insertions in ProductionNode

Implementation notes:

- Added `clara.rules.memory/add-all!` to avoid Collection.addAll(Collection) of Java.  It performs a copy of the incoming collection as an array.  This is unnecessarily slow in some cases, especially when I was looking at time spent in clj data structures to array, and it isn't valuable.
- `clara.rules.memory/remove-first-of-each!` is used for memory removals on the JVM side.  The non-destructive `remove-first-of-each` is still used by JS side, as well as in a few places in clara.rules.engine.
- There has been a slight change sometimes in the order facts are propagated, so I had to fix a few order-dependent tests in clara.tools.test-inspect.  It looks like they had no reason to assume the order they were assuming though.  I made them order independent now.

Note I had, but didn't address with this change @ [2].

[1] 
- There used to be a fragile and hard to understand situation here.  `flush-updates` did not flush updates that resulted from flushing the current updates.  This should be a recursive process.  Instead the ProductionNode happened to stage an "empty" activation when performing the `clara.rules.memory/remove-activations!`.
- Counterintuitively, `remove-activations!` was able to add new activations.  They didn't do anything other than cause the agenda group to possibly change, which resulted in additional (and necessary!) `flush-updates` calls.  I couldn't break the existing behavior, but I'm not confident that it doesn't have edge cases that this wouldn't flush all updates prior to switching to a lower priority agenda group.
- `clara.test-rules/test-flush-all-updates-before-changing-activation-group` is an example to test this odd case to make sure it is working both before and after these changes.

[2] 
- Applies to https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/memory.cljc#L499
- There cannot be a previous when building and iterating from the incoming map right?  Keys have to be unique.  I don't think there is any reason to check for previous here.  Also, activations probably cannot be nil here anyways, but it doesn't necessarily hurt to check...  Adding and removing activations should just never associate nil to a key.  That isn't useful.

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -152,26 +152,34 @@
 (def ^:dynamic *rule-context* nil)
 
 (defn- flush-updates
-  "Flush pending updates in the current session. Returns true if there were some items to flush,
-  false otherwise"
+  "Flush all pending updates in the current session. Returns true if there were
+   some items to flush, false otherwise"
   [current-session]
+  (letfn [(flush-all [current-session flushed-items?]
+            (let [{:keys [rulebase transient-memory transport insertions get-alphas-fn listener]} current-session
+                  pending-updates @(:pending-updates current-session)]
 
-  (let [{:keys [rulebase transient-memory transport insertions get-alphas-fn listener]} current-session
-        pending-updates @(:pending-updates current-session)]
+              ;; Remove the facts here so they are re-inserted if we flush recursively.
+              (reset! (:pending-updates current-session) [])
 
-    ;; Remove the facts here so they are re-inserted if we flush recursively.
-    (reset! (:pending-updates current-session) [])
+              (if (empty? pending-updates)
+                flushed-items?
+                (do
+                  (doseq [partition (partition-by :type pending-updates)
+                          :let [facts (mapcat :facts partition)]
+                          [alpha-roots fact-group] (get-alphas-fn facts)
+                          root alpha-roots]
 
-    (doseq [partition (partition-by :type pending-updates)
-            :let [facts (mapcat :facts partition)]
-            [alpha-roots fact-group] (get-alphas-fn facts)
-            root alpha-roots]
+                    (if (= :insert (:type (first partition)))
+                      (alpha-activate root fact-group transient-memory transport listener)
+                      (alpha-retract root fact-group transient-memory transport listener)))
 
-      (if (= :insert (:type (first partition)))
-        (alpha-activate root fact-group transient-memory transport listener)
-        (alpha-retract root fact-group transient-memory transport listener)))
+                  ;; There may be new :pending-updates due to the flush just
+                  ;; made.  So keep flushing until there are none left.  Items
+                  ;; were flushed though, so flush-items? is now true.
+                  (flush-all current-session true)))))]
 
-    (not (empty? pending-updates))))
+    (flush-all current-session false)))
 
 (defn insert-facts!
   "Place facts in a stateful cache to be inserted into the session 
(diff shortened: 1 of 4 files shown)


rbrush:
This looks great! I'm excited to hear about the perf improvements this leads to.

I think you're correct on [1], and things were working just by happenstance of the artificial activation that is created. Recursively handling this as done in your change is much better. The test you added appears to cover the main edge case, flushing everything from one activation group before moving onto the next. Thanks for tracking that down.

As for item [2] on your list: you're right, there can't be a previous, so that logic is superfluous. To be honest I'm not sure what I was thinking there, or if it was cruft from some previous representation of the working memory. But that's not really relevant to this change, so I can fix it independently.

I'm happy to merge this as it stands now, but let me know if there is anything else you wanted to do here.

mrrodriguez:
Thanks for looking it over.  I don't have anything more to do with this particular change, so to me it is good to go.  It'd probably be good to get @WilliamParker to take a look over it.  He looked at an earlier pass I had, but I don't think he has looked at this one yet.

On the performance topic, I completely forgot to mention that in the actual PR.  It's difficult for me to get any solid numbers yet, but the overall time has some improvements, but where the big pay off has been is when we encounter the datasets that started taking much, much longer than the average.  I'm not longer seeing any "long running" datasets as I described in the main testbed I've been using for this so far that I described in https://github.com/rbrush/clara-rules/issues/184.  

I'm thinking of this performance benefit as giving us more stability/robustness to dealing with datasets that trigger bad, perhaps edge-case, scenarios.  This is certainly an important benefit to our use case though!

Another interested note is back in https://github.com/rbrush/clara-rules/pull/183 I made a small microbenchmark looking like:
```clj
(let [inserted (insert-all s temps)
      retracted (time (apply retract
                             inserted
                             temps))]
  [(-> inserted
       fire-rules
       (query q)
       count)
   (-> retracted
       fire-rules
       (query q)
       count)])

```
Which previously returned 
```clj
"Elapsed time: 2390.955789 msecs"
[10000 0]
```

Running that same exact example with these changes in place, I get
```clj
"Elapsed time: 116.660136 msecs"
[10000 0]
```
So it went from over 2 seconds to less than 200 ms.  The time being spent there after that change was all in this `remove-first-of-each` path.

Also, I forgot to note in this PR overview, I only am causing memory to go into mutable-collections when items are removed because that is the expensive path.  This likely helps to avoid the degradation of `to-transient` and `to-persistent!` with these changes.  The idea being a small percentage of memory should be having items removed from it from one batch of changes to the next.

As for [2], I'm glad you agree it is superfluous and I'm not crazy. :)

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -152,26 +152,34 @@
...
+                      (alpha-activate root fact-group transient-memory transport listener)
+                      (alpha-retract root fact-group transient-memory transport listener)))
 
-      (if (= :insert (:type (first partition)))
-        (alpha-activate root fact-group transient-memory transport listener)
-        (alpha-retract root fact-group transient-memory transport listener)))
+                  ;; There may be new :pending-updates due to the flush just
+                  ;; made.  So keep flushing until there are none left.  Items
+                  ;; were flushed though, so flush-items? is now true.
+                  (flush-all current-session true)))))]
```
WilliamParker: I'm concerned that we are vulnerable to a StackOverflow here; depending on the rules and data I don't see any bound to the recursion here.  I suggest implementing this with recur instead.

mrrodriguez: We talked about this some in person, but basically it is not a very realistic scenario to get into a situation where this `flush-all` recursion would go beyond a small number of iterations.  Due to this, not fixing it right now.  The only new `:pending-updates` that can occur on each iteration is new retractions being added.  The way to get this flushing cycle to go on a long time would be to have many "chained" logical dependencies among facts.
Example:
```clj
(let [n 10
      ss (mapv str (range n))
      rhss (mapv (fn [s] `(insert! ~s))
                 ss)
      rules (mapv (fn [s rhs]
                    {:lhs (if (= ::none s)
                            []
                            [{:type `String
                              :constraints [`(= ~'this ~s)]}])
                     :rhs rhs})
                  (cons ::none ss)
                  rhss)
      q (dsl/parse-query [] [[?ss <- (acc/all) :from [String]]])
      s (mk-session rules [q])
      before (-> s
                 fire-rules)
      after (-> before
                (retract "0")
                fire-rules)]
  [(-> (query before q) first :?ss count)
   (-> (query after q) first :?ss count)])

;;= [10 0]

```

Changing `n` to 1000 here, does indeed give a StackOverflowError however, it is not due to this change (happens in master and 0.11.1 etc).  It is due to lazy collections being used in a few places of the engine.

Examples being:
1. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L343
2. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L390
3. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L271

1 and 2 are very common and making them eager could have adverse side effects to performance - so that would warrant its own issue.  3 is a bit easier, but again it'd be worthwhile to assess where laziness is helping and where it is not.

Also, this rule situation I just described is not something I'd expect to see in practice.  It would not be likely that rules structured in this way would be very coherent or even be able to perform very well with such a deeply linked dependency chain among the rules.

So for now, keeping this `flush-all` as is for now unless there are objections.  The lazy collections comes up before we even get to this flushing loop having issues.  Changing to loop-recur here would be fairly trivial at a later time.

WilliamParker: @mrrodriguez  and I discussed this offline; it turns out that there are some other places that would overflow before this would, and the range of cases that would be vulnerable is more limited than I initially thought.  I'm OK with leaving this as is.

WilliamParker:
+1 to merging this.

rbrush:
Merged. Thanks @mrrodriguez and @WilliamParker!

PR #191, opened 2016-05-13
Issue 182: Prevent nil accumulator results from propagating and avoid…

WilliamParker:
… propagation from accumulators when the Rete operation does not change the final result

Some notes here:

- If the retracted or reduced value from an accumulator is nil, this assumes that the converted return
value is also nil.  Currently, because a nil initial value does not propagate and we don't propagate nils that come from retractions (as opposed to propagation) an accumulator that has a nil initial value and takes on a non-nil value for any actual data can have a convert-return-fn that is not nil safe.  Making such a non-passive change to support a pattern that is probably not that common anyway (taking a nil value and converting it to something truthy) didn't seem worth the complication/risk.  Furthermore, such convert-return functions would be unsafe anyway now since any retraction, even one that didn't change the final value, would cause retraction of the relevant tokens. [1, 2] This could be revisited in [3, 4, 5].  Does this approach seem correct to you?
- Since a nil value will be ineligible for propagation, I don't think we need to special-case ::no-accum-result in do-accumulate anymore. [6] We can just return nil there and the output will be suppressed.
- I don't think [7] or [8] needs
````
{result-binding initial-value}
````
since the result-binding will be added (and overridden) anyway by [9].  Furthermore, this binding is invalid because it is the initial-value, not the initial value after application of the convert-return-fn.  I removed this bindings map and replaced it with an empty map.

Let me know if you can think of additional test cases that you'd like to have and I'd be happy to add them.  The various pieces of accumulator logic can interact in a lot of different ways, which makes figuring out which sorts of test cases will provide the best benefit versus cost a difficult calculation.  This is particularly true for tests that are targeted at a particular implementation detail, even if they don't depend on that implementation detail to pass (they might just lose value if that implementation detail changed).  Overall, I think generative tests for accumulators, even if the generated part was just data and/or insert/retract patterns, rather than the actual rules or accumulators themselves, are a idea worth exploring to help cope with the large set of possible interactions.  However, that also seems out of scope here, and such tests would probably fail right now anyway due to [4, 5].

@mrrodriguez will probably want to take a look at this as well before merging.  I think there is enough complication and edge cases here to merit more than one pair of eyes over it.

1. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L821
2. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L1004
3. https://github.com/rbrush/clara-rules/issues/190
4. https://github.com/rbrush/clara-rules/issues/189
5. https://github.com/rbrush/clara-rules/issues/188
6. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L837
7. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L772
8. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L946
9. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L659

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -654,9 +654,8 @@
 
 (defn- retract-accumulated
   "Helper function to retract an accumulated value."
-  [node accum-condition accumulator result-binding token result fact-bindings transport memory listener]
-  (let [converted-result ((:convert-return-fn accumulator) result)
-        new-facts (conj (:matches token) [converted-result (:id node)])
+  [node accum-condition accumulator result-binding token converted-result fact-bindings transport memory listener]
+  (let [new-facts (conj (:matches token) [converted-result (:id node)])
         new-bindings (merge (:bindings token)
                             fact-bindings
                             (when result-binding
@@ -668,9 +667,8 @@
 
 (defn- send-accumulated
   "Helper function to send the result of an accumulated value to the node's children."
-  [node accum-condition accumulator result-binding token result fact-bindings transport memory listener]
-  (let [converted-result ((:convert-return-fn accumulator) result)
-        new-bindings (merge (:bindings token)
+  [node accum-condition accumulator result-binding token converted-result fact-bindings transport memory listener]
+  (let [new-bindings (merge (:bindings token)
                             fact-bindings
                             (when result-binding
                               { result-binding
@@ -687,7 +685,7 @@
               (= previous-result converted-result))
 
       (send-tokens transport memory listener (:children node)
-                 [(->Token (conj (:matches token) [converted-result (:id node)]) new-bindings)]))))
+                   [(->Token (conj (:matches token) [converted-result (:id node)]) new-bindings)]))))
 
 (defn- has-keys?
   "Returns true if the given map has all of the given keys."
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: src/main/clojure/clara/rules/engine.cljc [hunk 1]
```diff
@@ -687,58 +685,75 @@
...
+                ;; Note that we check that the :initial-value is truthy above, which is why we
+                ;; don't need (when initial-value (convert-return-fn initial-value)) here.
+                initial-converted (convert-return-fn initial-value)]
+            
+            (l/add-accum-reduced! listener node join-bindings initial-value fact-bindings)
+
+            ;; Add it to the working memory.
+            (mem/add-accum-reduced! memory node join-bindings initial-value fact-bindings)
+
+            (when initial-converted
```
rbrush: Should we be using a not nil check, (e.g., some?) for these values, since accumulating something to a boolean false seems valid?

WilliamParker: I should have mentioned this in the PR notes.  I was conflicted on this.  The current implementation just uses when checks, so false will be filtered (with the inconsistency caveats discussed in the parent issue), and boolean false is a falsey Clojure value, so maybe it makes sense to suppress it.  On the other hand, I can definitely imagine accumulators that would want to use boolean false.  If you think we should allow boolean false I can change this to use a some? check; honestly my view on this at any given time has been pretty much a coin flip.

mrrodriguez: I have tended to prefer Clojure "falsey" instead of purely nil.  However, there are reasons to debate that.

I just think it may be a stretch to use "false" as a return in an accumulator and I think anyone doing that should probably be questioning whether or not that makes sense that the rules would propagate that.

mrrodriguez:
I've finished looking through this. +1 from my perspective. It looks good to me.

src/main/clojure/clara/rules/engine.cljc
rbrush: Agreed that an accumulator returning false is a stretch, but I can't think of a good reason to use it rather than _some?_. It seems like the less surprising behavior if someone does write some kind of "is-there-any-of-under-these-conditions" type of accumulator that could just return false.

rbrush:
Other than the comment on the nil vs. falsey accumulator checks, I'm +1 on this as well.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: I can understand that some cases propagating false may be "expected" to work.  I think this would be a larger deviation from the current behavior though.  I know in our use-cases of accumulators we have always used false/nil interchangeably as "falsey" and not expected to propagate.  So if false was propagated, I think it'd be something that had to be explained to others to avoid using.

I'm not entirely opposed, I just wonder how broad of an impact that would be to change it.  I also, think it is hard to understand what people would expect.  I think it can be surprising either way depending on what the goal is at the time.

stephenbrady: Accumulating to false seems very reasonable to me and makes sense. Agreed that it's common to see falsey in Clojure, false and nil are definitely not the same, and I think this is a case where nil reifies void/nothing, which is a distinction this code needs, and false means the opposite of true.

WilliamParker: I am working on changing this so boolean false will propagate.

WilliamParker:
I tried to update this, but GitHub automatically closed the PR on me when I deleted the branch and pushed a replacement.  I have a new PR at https://github.com/rbrush/clara-rules/pull/193 with the changes to propagate boolean false.

PR #193, opened 2016-05-18
Issue 182: Prevent nil accumulator results from propagating, allow bo…

WilliamParker:
…olean false accumulator results to propagate, and prevent propagation when a right activation or right retraction does not change the final propagated result or bindings


rbrush:
These changes look good time (and thanks for the nil change and corresponding tests). I'll give any others who might be interested in a review a chance to look but can merge tomorrow.

mrrodriguez:
+1 from me.

These look good and I'm fine with the behavior after discussing it more.  Boolean false can propagate and just be something to keep in mind.

PR #195, opened 2016-05-20
Fix for mutable JVM colls in local memory for efficiency

mrrodriguez:
https://github.com/rbrush/clara-rules/pull/187 introduced a subtle issue that can happen due to Clojure records implementing java.lang.Object.equals() in a somewhat odd way.

Clojure records implement Object.equals() in a way that is independent of the type of the record.  Comparing records for Object.equals() is like comparing 2 java.util.Map's instead of classes with specific types.

Clojure records do use their type information when comparing via clojure.core/= though.  Also, clojure.core/= should actually be faster than the Object.equals() impl for records.

It would be rare to see this case come up in practice, but fixing it is important.  Also, it looks like we can squeeze out a little bit more of a performance gain from these changes too by using Clojure's equals for records instead of Java equals here.

I have a test case that demonstrates what sort of situation would fail due to the changes of https://github.com/rbrush/clara-rules/pull/187 and now it passes.

### DIFF
--- a/src/main/clojure/clara/rules/memory.cljc
+++ b/src/main/clojure/clara/rules/memory.cljc
@@ -102,6 +102,40 @@
   (to-persistent! [memory]))
 
 #?(:clj
+   (defn- coll-empty?
+     "Returns true if the collection is empty.  Does not call seq due to avoid
+      overhead that may cause for non-persistent collection types, e.g.
+      java.util.LinkedList, etc."
+     [^java.util.Collection coll]
+     (or (nil? coll) (.isEmpty coll))))
+
+#?(:clj
+   (defn- list-remove!
+     "Removes the item, to-remove, from the given list, lst.  If it is found and
+      removed, returns true.  Otherwise returns false.  Only removes the first 
+      element in the list that is equal to to-remove.  If others are equal, they
+      will not be removed.  This is similar to java.util.List.remove(Object) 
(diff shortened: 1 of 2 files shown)


WilliamParker:
+1 from me.

While doing some profiling on work for https://github.com/rbrush/clara-rules/issues/192 I noticed that the calls that are present in clojure.lang.APersistentMap.mapEquals [1] and not in record equality [2], notably the .size calls that, in our case of records withouts any additional keys that are only compared against their own type, don't do any good.  Activation.size() and Token.size() accounted for 4.4% of execution time in that particular case.

Overall, it seems to me like further optimizing our equality checks beyond what we're getting from Clojure's defrecord by default, if possible, could be valuable, but this is a good first step, and any further improvement would probably mostly build on what is here, likely changing the actual equality semantics on the relevant types or using some check other than

````
(= to-remove x)
````
in list-remove!.  In any case, as @mrrodriguez mentions above, this isn't purely a performance issue; there are correctness concerns as well.  Getting performance improvements is a very welcome bonus though. :)

1. https://github.com/clojure/clojure/blob/clojure-1.8.0/src/jvm/clojure/lang/APersistentMap.java#L52
2. https://github.com/clojure/clojure/blob/clojure-1.8.0/src/clj/clojure/core_deftype.clj#L205

PR #196, opened 2016-05-23
Fix to-dnf to support :exists inside :and and :or conditions

WilliamParker:
The basic problem here is that to-dnf currently considers everything to be either a "conjunction" or "disjunction" when processing the children of :and and :or conditions. [1]  In this code, it looks to me like anything that is not a :or condition is considered to fall in the conjunctions bucket.  This is done by filtering on the condition type for each child.  :exists is currently not considered to be either, so it just gets filtered out of the dnf form completely.

Current state:
````
clara.test-rules> (com/to-dnf [:or [:exists {:constraints []}] [:exists {:constraints []}]])
(:or)
````

State after this change:

````
clara.test-rules> (com/to-dnf [:or [:exists {:constraints []}] [:exists {:constraints []}]])
(:or [:exists {:constraints []}] [:exists {:constraints []}])
````

1. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/compiler.clj#L477

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -461,6 +461,8 @@
 
         :test expression
 
+        ;; Note that :exists does not support further nested boolean conditions.
+        ;; It is just syntax sugar over an accumulator.
         :exists expression
 
         ;; Double negation, so just return the expression under the second negation.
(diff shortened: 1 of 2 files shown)


rbrush:
Fixed a merge conflict in the commit https://github.com/rbrush/clara-rules/commit/8ec6326f417adc61f33af2500c32d13afe0ea6a6 and merged to master. Thanks!

PR #197, opened 2016-05-24
Issue 192: Return activations from the memory in the order that the c…

WilliamParker:
…orresponding rules were provided to the compiler

### DIFF
--- a/src/main/clojure/clara/rules.cljc
+++ b/src/main/clojure/clara/rules.cljc
@@ -268,7 +268,13 @@
          ;; The symbol is not qualified, so treat it as a namespace.
          (->> (ns-interns sym)
               (vals) ; Get the references in the namespace.
-              (filter #((some-fn :rule :query :production-seq) (meta %))) ; Filter down to rules, queries, and seqs of both.
+              (filter var?)
+              (filter (comp (some-fn :rule :query :production-seq) meta)) ; Filter down to rules, queries, and seqs of both.
+              ;; If definitions are created dynamically (i.e. are not reflected in an actual code file)
+              ;; it is possible that they won't have :line metadata, so we have a default of 0.
+              (sort (fn [v1 v2]
+                      (compare (or (:line (meta v1)) 0)
+                               (or (:line (meta v2)) 0))))
               (mapcat #(if (:production-seq (meta %))
                          (deref %)
                          [(deref %)])))))))
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1521,12 +1521,26 @@
   ([sources-and-options]
    (let [sources (take-while (complement keyword?) sources-and-options)
          options (apply hash-map (drop-while (complement keyword?) sources-and-options))
-         productions (into #{}
-                           (mapcat
-                            #(if (satisfies? IRuleSource %)
-                               (load-rules %)
-                               %))
-                           sources)] ; Load rules from the source, or just use the input as a seq.
+         productions (->> sources
+                          ;; Load rules from the source, or just use the input as a seq.
+                          (mapcat #(if (satisfies? IRuleSource %)
+                                     (load-rules %)
+                                     %))                                        
+                          (map (fn [n production]
+                                 (vary-meta production assoc ::rule-load-order (or n 0)))
+                               (range))
+                          ;; Ensure that we choose the earliest occurrence of a rule for the purpose of rule order.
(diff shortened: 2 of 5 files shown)


WilliamParker:
## Implementation Background

- When we originally tried to do rule order by just giving each rule its own salience group, performance became much worse (at least double the original execution time) on our sample data.  I believe this is because the engine flushes all activations as soon as the activation group changes.  For example, suppose that rules 1 and 2 insert facts of FactTypeA, and rule 3 accumulates on FactTypeA.  Ordering from rules 1 to 3 by salience will ensure that rules 1 and 2 execute before rule 3, but it also can cause work in the rules network to be performed after rule 1 that could be more efficiently delayed until after rule 2.  Since this rule ordering is for performance and consistency, rather than correctness, we need not flush after each rule.
- Accordingly, this implementation does not change the semantics of salience or use it to create rule ordering.  Rather, rule order is a separate attribute.  Instead of the activation map having sequences as values, it now has a priority queue ordered by rule order.  Activations will be removed from this queue in rule order.  An important point to note is that the new order in which activations are returned was always in fact a possible order, so we shouldn't break any guarantees here.
- The PriorityQueue will use Java equality semantics to determine which of its elements to remove, but we want to use Clojure equality semantics as discussed at https://github.com/rbrush/clara-rules/pull/195 In order to accomplish this I wrapped the Activation in a deftype that provides its own Object.equals implementation.  Also note that I checked the node-id rather than the entire node in this equals implementation; since the node-id should be unique I see no reason to spend time checking equality of the entire node.

## Ordering used
- This is fairly straightforward; the ordering chosen is simply the order in which
the rules are provided to Clara.
- If the rule comes from a defrule then the :line metadata is used.  If the rule comes from a dynamically created namespace without :line metadata then a default rule order of 0 will be assigned, with the result that rule ordering in such namespaces will be undefined.  In principle we could employ some other strategy to create ordering (say comparing names) but such a strategy would most likely have no performance benefits on average.  Arguably a better way would be to just force such use-cases to give an ordering if one was desired.  I don't think we need to address such cases now though.
- In the case of :production-seq, which was added in and is discussed more at https://github.com/rbrush/clara-rules/issues/134 we just used the ordering in the sequence.  The rules in the sequence will be "spliced in" at the relative position of their sequence to other rules or :production-seq definitions in their namespace.
- For anyone extending IRuleSource [1], the ordering used will simply be the order the implementation returns.
- If a rule occurs more than once in the rule sources provided (including both literal rule
structures and returned rule structures from rule sources) its position in the rule ordering will be that of its first occurrence.  As far as I can tell from the documentation, neither clojure.core/distinct nor clojure.core/set makes guarantees as to which instance will be chosen when there are equal candidates, so I ended up writing my own (brief) implementation that makes the guarantee to choose the first occurrence.  Even if one of them works right now it seems safer to avoid reliance on unspecified behavior.

## Implications for caching

- When session caching is used, rule order is not considered in determining whether the productions given are equal to some previous arguments.  If the same rules are provided in different orders at different times and caching is used the ordering for all sessions returned will be that of the first call to create a session.  I would expect that users varying such ordering don't care about ordering, and we have had a specific request for ordering to not be considered in session caching as mentioned at [2].

## Performance results:

- I saw significant improvement in some of our worst cases and more modest performance improvements overall with this change.  Obviously this will be heavily dependent on the rules and data in question though.  Profiling didn't show the queue removals taking much time; in my sample case pop-activation! took about .1 % of runtime overall. 


1. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/compiler.clj#L24

2. https://github.com/rbrush/clara-rules/issues/157#issuecomment-171344018

WilliamParker:
The closing of this was an accident.  I have reopened it.

mrrodriguez:
+1
This looks good to me.  There are some merge conflicts from the merge of https://github.com/rbrush/clara-rules/pull/195, but should be fairly straightforward to resolve now.

WilliamParker:
I've updated the pull request with a rebase on top of https://github.com/rbrush/clara-rules/pull/195.  I'm not sure why Travis didn't rebuild the pull request but the branch I'm pulling from did rebuild at https://travis-ci.org/WilliamParker/clara-rules/builds/132850963

rbrush:
Merged! I did so by hand to avoid a conflict but everything looks good now. The amended commit is here: https://github.com/rbrush/clara-rules/commit/8eb3cdb70e57444edea529ba22ba7184e8eeb167

PR #201, opened 2016-06-08
Issue 199: Increase determinism of rulebase compilation to increase c…

WilliamParker:
…onsistency of execution times

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -414,7 +414,6 @@
     :condition
     (first expression)))
 
-
 (defn- cartesian-join
   "Performs a cartesian join to distribute disjunctions for disjunctive normal form.,
   This distributing each disjunction across every other disjunction and also across each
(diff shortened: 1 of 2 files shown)


src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1368,11 +1376,16 @@
                                    (update-in node-map [[condition env]] conj node-id)
                                    (assoc node-map [condition env] [node-id])))
                                {}
-                               condition-to-node-ids)]
+                               condition-to-node-ids)
+
+        ;; We sort the alpha nodes by the sum of the node ids they correspond to
+        ;; in order to make the order of alpha nodes for any given type consistent.
+        condition-to-node-entries (sort-by (fn [[k v]] (apply + v))
```
mrrodriguez: I know I've already talked about this once, but again I have thought of something.

What if two node ids end up adding up to the same thing?  I think it'd be nondeterministic still then.
Like what if you had `[1 2 3]` or `[2 4]` or `[6]` etc. I think you get it.

Would a better `sort-by` be something that actually ensures they aren't the same?
```clj
(sort-by (fn [[_ v]] (sort v)) <etc>)
```

Obviously, this isn't quite as fast, but probably doesn't matter since there wouldn't be too many node ids per condition in practical cases.

WilliamParker: Agreed; we should have a better uniqueness mechanism.  I'll change this.

WilliamParker:
I have updated the pull request in response @mrrodriguez 's comment on the previous diff.

mrrodriguez:
+1
 The changes look good to me.

rbrush:
Merged, and thank you!

Issue #189, opened 2016-05-11
Facts under a binding group can cause an accumulator with an initial value to have downstream results even after retraction


WilliamParker:
I'd expect the following test to pass:

````
(deftest test-accum-all-with-bindings
  (let [mci-temps [(->Temperature 10 "MCI")
                   (->Temperature 20 "MCI")]
        sfo-temps [(->Temperature 10 "SFO")
                   (->Temperature 20 "SFO")]
        temp-rule (dsl/parse-rule [[?temps <- (acc/all) :from [Temperature (< temperature 25) (= ?location location)]]]
                                  (insert! (->TemperatureHistory (sort-by :temperature ?temps))))

        temp-history-query (dsl/parse-query [] [[TemperatureHistory (= ?temps temperatures)]])

        all-temps-session (-> (mk-session [temp-rule temp-history-query] :cache false)
                              (insert-all (concat mci-temps sfo-temps))
                              fire-rules)]

    (is (= (-> (apply retract all-temps-session mci-temps)
               fire-rules
               (query temp-history-query))
           [{:?temps []}]))))
````

However, it fails:

````
FAIL in (test-accum-all-with-bindings) (test_accumulators.clj:194)

expected: ({:?temps ()}
           {:?temps (#clara.rules.testfacts.Temperature {:location "SFO", :temperature 10}
                     #clara.rules.testfacts.Temperature {:location "SFO", :temperature 20})})
  actual: [{:?temps []}]

    diff: - [nil
             {:?temps (#clara.rules.testfacts.Temperature {:location "SFO", :temperature 10}
                       #clara.rules.testfacts.Temperature {:location "SFO", :temperature 20})}]
         
Ran 17 tests containing 42 assertions.
1 failures, 0 errors.
{:error 0, :fail 1, :pass 41, :test 17, :type :summary}
````

I believe the root cause of this is similar to that in [1], but it is
a slightly different problem so I decided to log it separately.  The difference
is that this case depends on what other bindings exist and the resolution of [2].
If, as discussed in [2], we want an accumulator with an initial value
to propagate, even with nil bindings, we probably only want one such
result to propagate, and then only in the absence of any real bindings.
Thus, when we retract all previous elements, whether we should propagate
nothing after retracting or the initial value with nil bindings depends
on if other binding groups are present.  In general, my thinking here
is that adding and retracting fact(s) or simply never inserting them
should both leave the session in the same ultimate state (excepting listeners of course).

[...]

Issue #188, opened 2016-05-11
Accumulator bindings can remain after the facts that created those bindings are retracted 


WilliamParker:
I'd expect the ?temperature binding go away in the following case:

````
(defn invalid-accum-bindings-downstream []
                    (let [r (dsl/parse-rule [[?result <- (acc/all) :from [Cold (= ?temperature temperature)]]]
                                            (insert! (->TemperatureHistory [?result ?temperature])))
                          q (dsl/parse-query [] [[TemperatureHistory (= ?temps temperatures)]])]
                      (-> (mk-session [r q] :cache false)
                          (insert (->Cold 10))
                          fire-rules
                          (retract (->Cold 10))
                          fire-rules
                          (query q))))
````

However, it does not:

````
clara.test-rules> (invalid-accum-bindings-downstream)
({:?temps [() 10]})
````

I believe the reason for this is that when we retract an element
from an AccumulateNode we repropagate the new result, if any exists,
with the same bindings as the token that was retracted. [1]
The same behavior exists for the AccumulateWithJoinFilterNode [2].
However, if the original elements that created these bindings in the
first place are retracted it is no longer valid for these bindings to exist.

I discovered this while writing tests for [3] but the behavior is the same both before
and after those changes; this is a different concern.

1. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L822
2. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/engine.cljc#L1005
3. https://github.com/rbrush/clara-rules/issues/182

rbrush:
Good find. This is similar enough to #189 that I'll comment on both here, at least for now.

I think we need to add a check when the last value is retracted and include logic similar to the logic at [1], where we only send the initial value if all bindings are satisfied. (Not clear how to do this check that the "last" value is retracted for an accumulator. I'm going to take a closer look at that.)

As for the discussion in #102, I had forgotten about that issue but think that we shouldn't propagate down the network unless all variables can be bound, which I think is what I was trying to do with the logic at [1]. That would actually simplify the fix to this problem, since we'd propagate the initial value per distinct set of bindings in the accumulator, which fits in nicely with the existing flow since we activate the accumulator with each distinct set of bindings. Furthermore, I like the guarantee that a rule can only be activated if everything is bound...eliminates errors from subsequent tests and the rule itself.

[1]
[...]

PR #202, opened 2016-06-09
AccumulateNode no longer requires :combine-fn or :retract-fn

mrrodriguez:
- Addresses the issues 188, 189, and 200 for AccumulateNode
  - When given, :retract-fn is used though
- Default :retract-fn is built into the node
  - More performant than an external "retraction wrapper impl"
- :initial-value propagation semantics have been corrected (issue 188 and 189)
- :initial-value only added to memory once per all tokens in AccumulateNode left-activate
- :combine-fn is now optional and not provided by default for clara.rules.accumulators/reduce-to-accum
- Removed :combine-fn from built-in accumulators where it is not useful
- Test for retracting un-inserted fact in accumulate node (issue 200)
- Added tracing support to mem/remove-accum-reduced!

### DIFF
--- a/src/test/clojure/clara/test_rules.clj
+++ b/src/test/clojure/clara/test_rules.clj
@@ -807,6 +807,29 @@
                     accum-join-filter-q
                     "AccumulateWithJoinFilterNode")))
 
+(deftest test-retract-fact-never-inserted-from-accum
+  (testing "AccumulateNode"
+    (let [p (dsl/parse-query [] [[?s <- (acc/sum :temperature) :from [Temperature (= ?t temperature)]]])
+          res (-> (mk-session [p])
+                  (insert (->Temperature 10 "LAX"))
+                  (retract (->Temperature 10 "MCI")) ; retracted under same bindings as inserted, but never actually inserted
+                  fire-rules
+                  (query p))]
+      (is (= (frequencies [{:?s 10 :?t 10}])
+             (frequencies res)))))
+  
+  (testing "AccumulateWithJoinFilterNode"
+    (let [p (dsl/parse-query [] [[Cold (= ?t temperature)]
+                                 [?s <- (acc/sum :temperature) :from [Temperature (< ?t temperature)]]])
+          res (-> (mk-session [p])
+                  (insert (->Cold 9)
+                          (->Temperature 10 "LAX"))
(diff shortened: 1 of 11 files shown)


src/test/clojure/clara/test_rules.clj
```diff
@@ -3876,16 +3926,18 @@
...
                  (fire-rules)
                  (retract (->Cold 10))
                  (fire-rules)
                  (query q1))
-             [{:?temps [[] 10]}])
+             (if (= "AccumulateWithJoinFilterNode" node-type)
+               [{:?temps [[] 10]}]
+               ;; FIXME: The correct assertion here is on equality to {:?temps [[] nil]}.
+               ;; Since (->Cold 10) is retracted we should use the initial value here.
+               ;; This is an existing defect that has been logged at https://github.com/rbrush/clara-rules/issues/188
```
rbrush: The comment on this PR suggest that 188 should be fixed, but this suggests there is more to do with the AccumulateWithJoinFilterNode. I'm okay with merging this either way, following up on #188 later, since it isn't a regression of current behavior.

mrrodriguez: I should have mentioned this in my synopsis of the changes.  
Where I wrote
> Addresses the issues 188, 189, and 200 for AccumulateNode

I specifically meant those issues were addressed for AccumulateNode and not AccumulateWithJoinFilterNode.  I should have made that more obvious though.  The issues #188 and #189 should remain open until we fix that one as well.

I left AccumulateWithJoinFilterNode out of this review since there were already significant changes and I figured it would be more of a cohesive set of changes if I kept it contained to the AccumulateNode for now.  

I think much of these same patterns can be used by the AccumulateWithJoinFilterNode to accomplish the same goals.  I'd hope and like to figure out a little bit more of a pattern of reuse of this logic when doing that, since the logic has became more complicated over time for both nodes.  There are some significant differences between the two nodes still, but I think we can get some reuse around the handling of the `:initial-value`.  This is why I considered that to be too much more work to add to the scope of these changes.

rbrush: Makes sense, and I agree with breaking this up rather than having a very large change. Thanks for the clarification.

rbrush:
LGTM. I'll go ahead and merge, and we can leave #188 and #189 open to track additional work.

rbrush:
Merged!

PR #206, opened 2016-07-09
Issue 205: Add generators to create random insert/retract/fire patter…

WilliamParker:
…ns for testing

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -4,8 +4,9 @@
   :license {:name "Eclipse Public License"
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :dependencies [[org.clojure/clojure "1.7.0"]
-                 [org.clojure/clojurescript "1.7.170" :scope "provided"]
                  [prismatic/schema "1.0.1"]]
+  :profiles {:dev {:dependencies [[org.clojure/math.combinatorics "0.1.3"]]}
+             :provided {:dependencies [[org.clojure/clojurescript "1.7.170"]]}}
   :plugins [[lein-codox "0.9.0" :exclusions [org.clojure/clojure]]
             [lein-javadoc "0.2.0" :exclusions [org.clojure/clojure]]
             [lein-cljsbuild "1.1.3" :exclusions [org.clojure/clojure]]
@@ -50,6 +51,14 @@
                               "phantom-advanced" ["phantomjs"
                                                   "src/test/js/runner.js"
                                                   "src/test/html/advanced.html"]}}
+  
+  ;; Factoring out the duplication of this test selector function causes an error,
+  ;; perhaps because Leiningen is using this as uneval'ed code.
+  ;; For now just duplicate the line.
+  :test-selectors {:default (complement (fn [x]
+                                          (some->> x :ns ns-name str (re-matches #"^clara\.generative.*"))))
+                   :generative (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.generative.*")))}
+  
   :scm {:name "git"
         :url "https://github.com/rbrush/clara-rules"}
   :pom-addition [:developers [:developer
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: clojure
 lein: lein2
-script: lein2 test
+script:
+  - lein2 test
+  - lein2 test :generative
 # Workaround for https://github.com/travis-ci/travis-ci/issues/4691
 sudo: required
 jdk:
--- a/src/test/clojure/clara/generative/generators.clj
+++ b/src/test/clojure/clara/generative/generators.clj
@@ -1,0 +1,101 @@
+(ns clara.generative.generators
+  (:require [clojure.math.combinatorics :as combo]
+            [clara.rules :refer :all]
+            [schema.core :as s]))
+
+(s/defschema FactSessionOperation {:type (s/enum :insert :retract)
+                                    :facts [s/Any]})
+
+(s/defschema FireSessionOperation {:type (s/enum :fire)})
+
+(s/defschema SessionOperation (s/conditional
+                                #(= (:type %) :fire) FireSessionOperation
+                                :else FactSessionOperation))
+
+(defn session-run-ops
+  "Run the provided sequence of operations on the provide session and return the final session."
+  [session ops]
+  (let [final-session (reduce (fn [session op]
+                                (let [new-session (condp = (:type op)
+                                                    :insert (insert-all session (:facts op))
+                                                    :retract (apply retract session (:facts op))
+                                                    :fire (fire-rules session))]
+                                  new-session))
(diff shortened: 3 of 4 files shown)


project.clj
```diff
@@ -5,6 +5,7 @@
             :url "http://www.eclipse.org/legal/epl-v10.html"}
   :dependencies [[org.clojure/clojure "1.7.0"]
                  [org.clojure/clojurescript "1.7.170" :scope "provided"]
+                 [org.clojure/math.combinatorics "0.1.3" :scope "test"]
```
mrrodriguez: To add "test scope" dependencies in leiningen you need to put them in a different profile.  The `:dev` profile is a common default one.  It'll be on the classpath when tests are ran automatically too.

project.clj
```diff
@@ -50,6 +51,14 @@
                               "phantom-advanced" ["phantomjs"
                                                   "src/test/js/runner.js"
                                                   "src/test/html/advanced.html"]}}
+  
+  ;; Factoring out the duplication of this test selector function causes an error,
```
mrrodriguez: You can do it.  It's a little odd.

You'd do something like
```clj
(defn gen-tests? [t] 
  (some->> t :ns ns-name str (re-matches #"^clara\.generative-tests.*")))
```

Then down here, do
```clj
:default (complement ~gen-tests?)
:generative ~gen-tests?
```

src/test/clojure/clara/generative_tests/generators.clj
```diff
@@ -0,0 +1,102 @@
+(ns clara.generative-tests.generators
+  (:require [clojure.math.combinatorics :as combo]
+            [clara.rules :refer :all]
+            [schema.core :as sc]))
```
mrrodriguez: nit: `s` is typically the alias used for this ns in this project and from schema docs themselves.

src/test/clojure/clara/generative_tests/generators.clj
```diff
@@ -0,0 +1,102 @@
...
+                               {:type :retract
+                                :facts (:facts op)}]))
+                          ops))
+        extras-with-dups (apply concat (repeat dup-level
+                                               (ops->extra ops)))
+
+        extra-subsets (combo/subsets extras-with-dups)]
+
+    (map (fn [extras]
+           (into ops (apply concat extras)))
```
mrrodriguez: nit: `(into ops cat extras)` would be the same thing (in behavior, not performance)

src/test/clojure/clara/generative_tests/generators.clj
```diff
@@ -0,0 +1,102 @@
...
+(sc/defn ops->permutations :- [[SessionOperation]]
+  "Given a sequence of operations, return all permutations of the operations for 
+   which no retraction of a fact that has not yet been inserted occurs.  By default 
+   permutations where extra insertions and retractions in equal number of inserted facts
+   are added will be present.  The default number of such pairs allowed to be added per insertion
+   is 1."
+  [ops :- [SessionOperation]
+   {:keys [dup-level] :or {:dup-level 1}}]
+  (let [dup-ops-seqs (ops->add-insert-retract ops dup-level)
+        permutations (apply concat
```
mrrodriguez: nit:
```clj
permutations (mapcat combo/permutations dup-ops-seqs)
```
just seems much clearer.

src/test/clojure/clara/generative_tests/generators.clj
```diff
@@ -0,0 +1,102 @@
+(ns clara.generative-tests.generators
```
mrrodriguez: It seems like the word "tests" is redundant since this is already under src/test. 

I'd just recommend
`clara.generative.generators`

project.clj
WilliamParker: Are you sure of this?  When I created a dummy project that has this version of clara as a dependency with no special profile or scope and uberjar'ed it, I was unable to find the combinatorics clj files inside the jar.

project.clj
WilliamParker: This seems like something that isn't really part of Leiningen's expected API.  Do you know of any docs etc. suggesting this approach is expected?

Copying 2 fairly simple lines of code isn't a big deal.  I'd prefer to just avoid any possible difficulties.

src/test/clojure/clara/generative_tests/generators.clj
WilliamParker: +1

src/test/clojure/clara/generative_tests/generators.clj
WilliamParker: +1

src/test/clojure/clara/generative_tests/generators.clj
WilliamParker: I'll change this.

src/test/clojure/clara/generative_tests/generators.clj
WilliamParker: I'll change this; I have no strong opinion either way.

WilliamParker:
I have responded to the comments here.  There is still some investigation to do on whether using :scope for dependencies has the impact desired on the build.

project.clj
mrrodriguez: If `:scope` works it is not the recommended way.  I think Leiningen started off with `:scope` and then changed to favoring profiles for this sort of thing.  If you are sure it is working, I guess you could leave it for now.  However, I'd say that'd worry me some across versions of lein.

There are two profiles automatically recognized by default for these sorts of dependencies, and they are `:dev` and `:provided`.  `:dev` is for "test" scope things.  When `lein pom` creates the Maven pom.xml it will be scoped as "test" as well.

The documentation for this is harder to find if you are searching for "scope" since lein doesn't have discussion on that terminology anymore.  The documentation is all on profiles.

project.clj
mrrodriguez: I don't care if you copy/paste since it is minimal and making it general is probably more ceremony than needed.

Leiningen has first-class for arbitrary evaluation to create dynamic project.cjl's like this though.  It is used extensively in real-world lein projects as well as lein documentation sample project that most things can be found in.
e.g.
https://github.com/technomancy/leiningen/blob/stable/sample.project.clj#L420

It is used all over the place though.  lein doesn't always do a great job of documenting every feature.  I think the expectation is to explore more and even read the codebase function documentation for some things.  There are also some StackOverflow answers that use approaches to lein problems with `~`.  

See 
https://github.com/technomancy/leiningen/blob/stable/sample.project.clj#L420
for where the main doc is for this.

WilliamParker:
@mrrodriguez , I think this should be ready for merge now with my most recent change.  Could you look over this and see if you have any further concerns?

mrrodriguez:
+1 looks good

rbrush:
Merged. Thanks!

PR #207, opened 2016-07-12
Making rule engine execution order more deterministic

mrrodriguez:
Clara currently uses `clara.rules.platform/tuned-group-by` in a number of places.  The grouping that this function does is unordered and often can vary from one JVM to the next.  In particular I've noticed this happening in the `clara.rules.engine/flush-updates` which was causing non-deterministic insertion order to happen across rules.  This makes it harder for us to track bad performance cases since they keep fluctuating as we run in different processes.

A side-effect of `flush-updates` not behaving deterministically was that insertion order across rules of the same salience group level were not deterministic and didn't necessarily respect rule load order as was implemented in https://github.com/rbrush/clara-rules/issues/192.  From my preliminary testing of this, I see a slight improvement in our execution times with this change in place.  This is likely due to the rule order now also being the insertion order used.  I've demonstrated this with a test that fails before this change and now passes.

My proposal is to (on the Java CLJ side) introduce a new `clara.rules.platform/group-by-seq` function that replaces the usages of `platform/tuned-group-by`.  Currently, I've left `platform/tuned-group-by` in case it ends up having some usages later.  It is also used in one CLJS place still, which I'm not wanting to change with this.  `platform/group-by-seq` returns a seq of tuples, which is the same as seq'ing over the map returned by `platform/tuned-group-by` before.  In all cases where we used `platform/tuned-group-by` we didn't use the map, but rather a seq over it.

I use a java.util.LinkedHashMap to maintain insertion order while building the groupings.  Then I convert this into an immutable seq structure.  If I tried to retain the usage of a map here, I'd have to return the java.util.LinkedHashMap or come up with some (expensive) way to port it into a Clojure sorted map.  This was not worthwhile to me, since we do not need a map anyways.  A java.util.LinkedHashMap would be harder to work with across the rest of the codebase since there are issues with the default Clojure seq implementations across mutable object iterators http://dev.clojure.org/jira/browse/CLJ-1738 that I see coming up when running this on JDK6 (oddly not on JDK8, which makes no sense and I want to investigate more for my own benefit).

Along with this, since I was messing with the `clara.rules.compiler/create-get-alphas-fn` already, I sorted the result of the ancestors-fn so that it would behave deterministically.  This would be more of an edge case, but since we cache this result and infrequently should hit this path, I figured it'd still be worthwhile.

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1499,9 +1499,12 @@
                                ;; Internal system types always use Clojure's type mechanism.
                                (type fact)
                                ;; All other types defer to the provided function.
-                               (fact-type-fn fact))))]
+                               (fact-type-fn fact))))
+
+        alpha-roots (get merged-rules :alpha-roots)]
+
     (fn [facts]
-      (for [[fact-type facts] (platform/tuned-group-by fact-grouping-fn facts)]
+      (for [[fact-type facts] (platform/group-by-seq fact-grouping-fn facts)]
 
         (if-let [alpha-nodes (get @alpha-map fact-type)]
 
@@ -1511,15 +1514,16 @@
           ;; The alpha nodes weren't cached for the type, so get them now.
           (let [ancestors (conj (ancestors-fn fact-type) fact-type)
 
-                ;; Get all alpha nodes for all ancestors.
-                new-nodes (distinct
-                           (reduce
-                            (fn [coll ancestor]
-                              (concat
-                               coll
-                               (get-in merged-rules [:alpha-roots ancestor])))
-                            []
-                            ancestors))]
+                ;; Get all alpha nodes for all ancestors.  Keep them sorted to maintain
+                ;; deterministic ordering of fact propagation across the network.
+                ;; Alpha nodes do not have a :node-id of their own right now, so sort
+                ;; by the :node-id of their :children.
+                new-nodes (sort-by #(mapv :node-id (:children %))
+                                   (into []
+                                         (comp (map #(get alpha-roots %))
+                                               cat
+                                               (distinct))
+                                         ancestors))]
 
             (swap! alpha-map assoc fact-type new-nodes)
             [new-nodes facts]))))))
--- a/src/main/clojure/clara/rules/platform.cljc
+++ b/src/main/clojure/clara/rules/platform.cljc
@@ -8,6 +8,37 @@
   (throw #?(:clj (IllegalArgumentException. description) :cljs (js/Error. description))))
 
 #?(:clj
+   (defn group-by-seq
+     "Groups the items of the given coll by f to each item.  Returns a seq of tuples of the form 
+      [f-val xs] where xs are items from the coll and f-val is the result of applying f to any of 
+      those xs.  Each x in xs has the same value (f x).  xs will be in the same order as they were
+      found in coll.
+      The behavior is similar to calling `(seq (group-by f coll))` However, the returned seq will
+      always have consistent ordering from process to process.  The ordering is insertion order
+      as new (f x) values are found traversing the given coll collection in its seq order.  The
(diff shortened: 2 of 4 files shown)


rbrush:
Looks good to me after an early review. I'll merge later today unless there are unexpected problems.

WilliamParker:
I'll plan to look at this today.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1511,15 +1511,16 @@
...
-                               (get-in merged-rules [:alpha-roots ancestor])))
-                            []
-                            ancestors))]
+                ;; Get all alpha nodes for all ancestors.  Keep them sorted to maintain
+                ;; deterministic ordering of fact propagation across the network.
+                ;; Alpha nodes do not have a :node-id of their own right now, so sort
+                ;; by the :node-id of their :children.
+                new-nodes (sort-by #(mapv :node-id (:children %))
+                                   (into []
+                                         (comp (map #(get (get merged-rules :alpha-roots) %))
```
WilliamParker: nit: I think this might be a bit more readable if you let-bound

````
(get merged-rules :alpha-roots)
````

File: src/main/clojure/clara/rules/platform.cljc [hunk 2]
```diff
@@ -8,6 +8,22 @@
   (throw #?(:clj (IllegalArgumentException. description) :cljs (js/Error. description))))
 
 #?(:clj
+   (defn group-by-seq [f coll]
```
WilliamParker: Could you add a (shortened) version of the commentary on the PR description here about respecting the order in the code here?  If I were just reading the code without that context the reason to set it up this way wouldn't be obvious.

src/main/clojure/clara/rules/platform.cljc
```diff
@@ -8,6 +8,22 @@
...
+     (let [^java.util.Map m (reduce (fn [^java.util.Map m x]
+                                      (let [k (f x)
+                                            xs (or (.get m k)
+                                                   (transient []))]
+                                        (.put m k (conj! xs x)))
+                                      m)
+                                    (java.util.LinkedHashMap.)
+                                    coll)]
+       (into []
+             (map (fn [^java.util.Map$Entry e] [(.getKey e) (persistent! (.getValue e))]))
```
WilliamParker: I think this will work out based on reading through http://dev.clojure.org/jira/browse/CLJ-1738 and the implementation of map at https://github.com/clojure/clojure/blob/clojure-1.8.0/src/clj/clojure/core.clj#L2618 but I wonder if just fully controlling it ourselves via a loop-recur would be safer and more resilient to any future Clojure changes.  I recall that we saw some discrepancies between Java 6 and 8 regarding this behavior - did we ever establish the reason for that?

WilliamParker:
+1 apart from my comments.

src/main/clojure/clara/rules/platform.cljc
mrrodriguez: The issue you are referring to was specifically around IdentityHashMap, which had a difference in Java 6 vs 8.
In Java 6 it's Map$Entry object was actually mutated in place every time Iterator.next() was called, which doesn't work correctly if the entry set is iterated via `reduce` or `transduce`, which is harder to avoid.

However, since these iterators have changed fairly often in Java for these classes, it would be safer to just do our own loop-recur here.  I'm changing it.

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: I'll do that.

src/main/clojure/clara/rules/platform.cljc
mrrodriguez: Yeah, adding.

mrrodriguez:
I've addressed all of the review comments made by @WilliamParker

WilliamParker:
+1

rbrush:
Merged. Thanks!

PR #208, opened 2016-07-19
Issue 102: Prevent AccumulateNode from creating binding groups when t…

WilliamParker:
…here are no corresponding elements

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -665,9 +665,6 @@
                 filtered-facts)
       (:initial-value accumulator))))
 
-(defn- remove-initial-value-binding [bindings]
-  (dissoc bindings ::initial-value))
-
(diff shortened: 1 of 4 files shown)


WilliamParker:
I've updated this on top of master after the merge of https://github.com/rbrush/clara-rules/pull/206

We can merge this now, after you've looked at it and assuming you agree with the changes of course, and I can then put in tests for this using the new generators later.  Alternatively, we can hold off entirely until generative tests are written; either works.  I'm personally slightly biased toward merging so all work can proceed from the same baseline, but I'm not aware of any work where having these changes in would be particularly beneficial.

rbrush:
I'll get this merged shortly; just poking around it a bit first.

WilliamParker:
No problem, I don't want to pressure you into merging without looking things over first. :)

rbrush:
Alright, merged! Thanks for all of your work on this.

PR #209, opened 2016-07-25
Add tests for AccumulateNode that insert the same facts and u…

WilliamParker:
…se the same rules, but vary the insert, retract, and fire patterns

I changed the default insertion duplication level from 1 to 0 here.  This is functionality that would, for example, take the pattern

- Insert A

and change it to

- Insert A
- Insert A
- Retract A

and the various order permutations of the latter, as well as keeping the original single-insertion scenario.  When I tried using the generator with the previous default I found that it made tests take far too long to run in some scenarios I wanted, presumably due to the combinatorial explosion of permutations.

Overall, none of the tests in this PR are going to get into complex scenarios in the compiler, rules network, etc.  My target here is more of a sanity check to improve confidence that the insert and retract algorithms on the AccumulateNode work correctly; this is very much a test of the AccumulateNode in isolation.   The tests are still written in terms of rules and sessions rather than direct calls to the node of course, with the attending advantages in reducing test brittleness.  However, given that we've had bugs that were purely internal to the accumulator nodes, I think such testing could be helpful.

### DIFF
--- a/src/test/clojure/clara/generative/generators.clj
+++ b/src/test/clojure/clara/generative/generators.clj
@@ -21,7 +21,8 @@
                                                     :retract (apply retract session (:facts op))
                                                     :fire (fire-rules session))]
                                   new-session))
-                              session ops)]))
+                              session ops)]
+    final-session))
 
 (defn ^:private retract-before-insertion?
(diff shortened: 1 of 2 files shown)


rbrush:
Merged. Love the improved test coverage. Thanks, Will!

PR #210, opened 2016-08-10
Issue 102: Prevent AccumulateWithJoinFilterNode from creating binding groups when t…

WilliamParker:
…here are no corresponding elements

I removed the special 1-element case of mem/remove-first-of-each since it changes the ordering of members of the collection of items that were not removed.  I believe the into call [1] is causing this.  If we look at split-with [2] it uses take-while, which creates a lazy sequence.  Into will append to the start of this sequence, so elements after the element removed will move to an earlier position.  I don't see a reason why this special case should be more performant than the current implementation for an arbitrary number of elements to remove; if anyone sees such a reason please explain it.  From looking through the history it looks like the general case was less efficient at one time.  Being able to guarantee that the ordering is unchanged allows us to make some performance optimizations.

The optimizations I made were

- In right-activate, accumulate the new elements on top of the result of the previous accumulation rather than re-accumulating from scratch.
- In right-retract, don't do anything for a token other than update the memory if the number of elements filtered for that token is the same before and after the retraction i.e. if none of the retracted elements matched that token in the first place.  If remove-first-of-each changes the ordering of elements then the ordering of elements in the memory will change, thus changing the accumulation order.  We cannot have a state where the elements are ordered like [element1 element2] in the memory but the downstream token has [element2 element1], since future operations may attempt to retract something using the former ordering, but fail to actually retract anything since the downstream tokens have another ordering.  Note that while rules need to accept any ordering the engine uses, accumulator results will not necessarily be equal (and in the case of acc/all they won't be).

I think these are valid, but would probably merit some thought from you too and anyone else looking at this PR.

Otherwise this is fairly similar to my changes for AccumulateNode, with some additional tests added to reflect AccumulateWithJoinFilterNode specific paths and previous tests for AccumulateNode parameterized to test both nodes.  I'll note that I decided to refer to the join types rather than the node names in the parameterization, since on further thought referring to node types in the tests seemed like unnecessary brittleness of test comments and descriptions.  I think my previous pattern of referring to the concrete node types was a mistake.

1. https://github.com/rbrush/clara-rules/blob/0.11.1/src/main/clojure/clara/rules/memory.cljc#L125
2. https://github.com/clojure/clojure/blob/clojure-1.8.0/src/clj/clojure/core.clj#L2888
3. https://github.com/clojure/clojure/blob/clojure-1.8.0/src/clj/clojure/core.clj#L2795

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -654,16 +654,10 @@
 
 (defn- do-accumulate
   "Runs the actual accumulation.  Returns the accumulated value."
-  [accumulator join-filter-fn token candidate-facts]
-  (let [filtered-facts (if join-filter-fn
-                         (filter #(join-filter-fn token % {}) ; TODO: add env
-                                 candidate-facts)
-                         candidate-facts)] 
-    (if (seq filtered-facts)
-      (r/reduce (:reduce-fn accumulator)
-                (:initial-value accumulator)
-                filtered-facts)
-      (:initial-value accumulator))))
+  [accumulator facts]
+  (r/reduce (:reduce-fn accumulator)
+            (:initial-value accumulator)
+            facts))
 
 (defn- retract-accumulated
   "Helper function to retract an accumulated value."
@@ -730,7 +724,7 @@
                       previous-reduced (if first-reduce?
                                          ;; Need to accumulate since this is the first time we have
                                          ;; tokens matching so we have not accumulated before.
-                                         (do-accumulate accumulator nil nil previous)
+                                         (do-accumulate accumulator previous)
                                          previous-reduced)
                       accum-reduced (when first-reduce?
                                       [previous previous-reduced])
@@ -851,19 +845,19 @@
                     ;; now matches, and nothing was reduced before.
                     (and has-matches?
                          (= ::not-reduced previous-reduced))
-                    (do-accumulate accumulator nil nil combined)
+                    (do-accumulate accumulator combined)
 
                     ;; There are matches, a previous reduced value for the previous items and a
                     ;; :combine-fn is given.  Use the :combine-fn on both the previously reduced
                     ;; and the newly reduced results.
                     (and has-matches?
                          (:combine-fn accumulator))
-                    ((:combine-fn accumulator) previous-reduced (do-accumulate accumulator nil nil facts))
+                    ((:combine-fn accumulator) previous-reduced (do-accumulate accumulator facts))
 
                     ;; There are matches and there is a previous reduced value for the previous
                     ;; items.  So just add the new items to the accumulated value.
                     has-matches?
-                    (do-accumulate (assoc accumulator :initial-value previous-reduced) nil nil facts)
+                    (do-accumulate (assoc accumulator :initial-value previous-reduced) facts)
 
                     ;; There are no matches right now.  So do not perform any accumulations.
                     ;; If there are never matches, time will be saved by never reducing.
@@ -966,7 +960,7 @@
                                       ;; remaining items after retraction.
                                       (if-let [retract-fn (:retract-fn accumulator)]
                                         (r/reduce retract-fn previous-reduced removed)
-                                        (do-accumulate accumulator nil nil retracted))
+                                        (do-accumulate accumulator retracted))
                                       ::not-reduced)
 
                   retracted-converted (when (and (some? retracted-reduced)
@@ -1032,6 +1026,11 @@
                 (send-accumulated node accum-condition accumulator result-binding token retracted-converted bindings
                                   transport memory listener)))))))))
 
+(defn- filter-accum-facts
+  "Run a filter on elements against a given token for constraints that are not simple hash joins."
+  [join-filter-fn token candidate-facts]
+  (filter #(join-filter-fn token % {}) candidate-facts))
+
 ;; A specialization of the AccumulateNode that supports additional tests
 ;; that have to occur on the beta side of the network. The key difference between this and the simple
 ;; accumulate node is the join-filter-fn, which accepts a token and a fact and filters out facts that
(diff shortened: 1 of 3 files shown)


mrrodriguez:
> I don't see a reason why this special case should be more performant than the current implementation for an arbitrary number of elements to remove; if anyone sees such a reason please explain it. From looking through the history it looks like the general case was less efficient at one time.

+1 to fixing `mem/remove-first-of-each`.  I worked on the last change to that area of code and I left the special case for size=1 just because it wasn't specifically what I was trying to fix at the time.  The behavior of the size=1 case is sneaky that it can change the order and is likely to lead to other subtle issues if left around.  Now the that the size>1 has been optimized anyways, there is really no great benefit to what the special case is doing.
Even with all that said, this function is barely used in the CLJ implementation anymore.  It's biggest use-case that I see is in AccumulateWithJoinFilterNode so we should certainly be sure to make it useful for that node.  :)

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1133,64 +1163,83 @@
...
+      
+      (doseq [token matched-tokens
+
+              :let [previous-filtered-facts (filter-accum-facts join-filter-fn token previous-candidates)
+
+                    new-filtered-facts (filter-accum-facts join-filter-fn token candidates)]
+
+              ;; If no new elements matched the token, we don't need to do anything for this token
+              ;; since the final result is guaranteed to be the same.
+              :when (seq new-filtered-facts)
```
mrrodriguez: We can skip 
```clj
previous-filtered-facts (filter-accum-facts join-filter-fn token previous-candidates)
```
as well if this is empty right?  That could be a fairly significant savings too since sometimes the join filter can be an expensive operation and we may have a lot of `previous-candidates`.  I didn't notice when looking through the logic you had previously.

```clj
(doseq [token matched-tokens

        :let [new-filtered-facts (filter-accum-facts join-filter-fn token candidates)]

        ;; If no new elements matched the token, we don't need to do anything for this token
        ;; since the final result is guaranteed to be the same.
        :when (seq new-filtered-facts)

        :let [previous-filtered-facts (filter-accum-facts join-filter-fn token previous-candidates)
              previous-accum-result-init <etc>]]
  <etc>)
```

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1133,64 +1163,83 @@
...
+                                                 (seq previous-filtered-facts)
+                                                 (do-accumulate accumulator previous-filtered-facts)
+
+                                                 (and (-> accumulator :initial-value some?)
+                                                      (empty? new-bindings))
+                                                 (:initial-value accumulator)
+
+                                                 ;; Allow direct determination later of whether there was a previous value
+                                                 ;; as determined by the preceding cond conditions.
+                                                 :else ::no-previous-value)
```
mrrodriguez: Why can't this case just be
```clj
:else
(:initial-value accumulator)
```
This special case seems hard to follow for me and it looks like it is being checked several times later too.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: Agreed, I have fixed this.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: If you accumulate to the same value as a non-nil initial value, but the initial value was not propagated because new binding groups are created, you need to propagate for the first time.  If you just used the initial-value as the previous value regardless of whether a previous value was in fact propagated you'd fail to propagate in this case because the later logic would see it as an equal previous result and not trigger downstream Rete operations due to the changes for https://github.com/rbrush/clara-rules/issues/182

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1133,64 +1163,83 @@
...
                   previous-candidates (when previously-reduced? previous-candidates)]]
 
-      ;; Combine the newly reduced values with any previous items.
-      (let [combined-candidates (into previous-candidates candidates)]
+      ;; Combine the newly reduced values with any previous items.  Ensure that new items are always added to the end so that
+      ;; we have a consistent order for retracting results from accumulators such as acc/all whose results can be in any order.  Making this
+      ;; ordering consistent allows us to skip the filter step on previous elements on right-activations.
+      (let [combined-candidates (into []
+                                      cat
+                                      [previous-candidates candidates])]
```
rbrush: Just for my own education, have you found the use of the cat transducer here to be more efficient than "(into previous-candidates candidates)" ? (I have no objections to this approach but it surprised me a bit.)

rbrush:
This is really excellent work, @WilliamParker. I'm fine with merging it if you and @mrrodriguez feel it's ready.

WilliamParker:
For my part, I think this is ready to merge.

src/main/clojure/clara/rules/engine.cljc
rbrush: Ah, makes sense. Thanks for the clarification.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: Yeah, I figured it out later.  Thanks for the explanation.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: Transducers are nice here to give us `concat` like behavior without the problem of stacking the lazyseqs unbounded depths where we get a stack overflow later.  `into` is always dependent on the `conj` behavior of the target collection.

`into` would have been faster for the case of the target being a vector though since it would only have to copy the second collection onto the first.  Either way, irrelevant here.

mrrodriguez:
+1 from me.

rbrush:
Merged. Thanks!

PR #214, opened 2016-08-12
- identical checks before equals for token/activation retraction in l…

mrrodriguez:
The changes here are described in https://github.com/rbrush/clara-rules/issues/213.

Some implementation notes:

* I did a lot of small tuning details as far as type-hinting and removing some indirection or abstraction layers typically taken by Clojure.  I was profiling these heavy data scenarios closely and these type hints and lower-level interface calls etc, definitely pay off in a significant way.  I'd not want to do this sort of work everywhere, but in this particular hot spot of code, I think it is advantageous to squeeze out all the performance as we can.  The workflow I'm looking at now needs to be performing in sub-second-like time, so these details matter.

* I did not make changes to how Element's are compared for equality here.  I did a demo run of that and didn't see as much of a payoff.  It may still be worthwhile, but I wanted to hold off on that until a separate issue if I find it to have some real value.

* clara.rules.memory namespace is indeed getting a little cluttered between the `:clj` and `:cljs` read-conditionals being used.  The functions towards the top that are only for `:clj` could certainly be pulled out to a new namespace.  I think this reorganization should be a separate issue though.

### DIFF
--- a/src/main/clojure/clara/rules/memory.cljc
+++ b/src/main/clojure/clara/rules/memory.cljc
@@ -114,29 +114,31 @@
 
 #?(:clj
    (defn- list-remove!
-     "Removes the item, to-remove, from the given list, lst.  If it is found and
-      removed, returns true.  Otherwise returns false.  Only removes the first 
-      element in the list that is equal to to-remove.  If others are equal, they
-      will not be removed.  This is similar to java.util.List.remove(Object) 
-      lst is updated in place for performance.  This implies that the list must 
-      support the mutable list interface, namely via the
+     "Removes the item, to-remove, from the given list, lst.  If it is found and removed,
+      returns true.  Otherwise returns false.  Only removes the first element in the list
+      that is equal to to-remove.  Equality is done based on the given eq-pred function.
+      If it isn't given, the default is = .  If others are equal, they will not be removed.
+      This is similar to java.util.List.remove(Object).  lst is updated in place for performance.
+      This implies that the list must support the mutable list interface, namely via the
       java.util.List.listIterator()."
-     [^java.util.List lst to-remove]
-     (if-not (coll-empty? lst)
-       (let [li (.listIterator lst)]
-         (loop [x (.next li)]
-           (cond
-             (= to-remove x)
-             (do
-               (.remove li)
-               true)
+     ([^java.util.List lst to-remove]
+      (list-remove! lst to-remove =))
+     ([^java.util.List lst to-remove eq-pred]
+      (if-not (coll-empty? lst)
+        (let [li (.listIterator lst)]
+          (loop [x (.next li)]
+            (cond
+              (eq-pred to-remove x)
+              (do
+                (.remove li)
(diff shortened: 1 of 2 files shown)


src/main/clojure/clara/rules/memory.cljc
```diff
@@ -241,7 +256,78 @@
...
+#?(:clj
+   (defprotocol IdentityComparable
+     (using-token-identity [this bool])))
+
+#?(:clj
+   (deftype RuleOrderedActivation [node-id
+                                   token
+                                   activation
+                                   rule-load-order
+                                   ^:unsynchronized-mutable use-token-identity?]
```
rbrush: So I *think* this is safe, since the mutation is only set to false and then under specific circumstances. Sessions can be shared between threads, of course, and mutations to this won't be visible to other threads since it's not at least volatile. But that should also be safe, since the calling thread would just set this to false again.

Might be worth adding a comment here as to these constraints.

rbrush:
+1 overall. (Quite the involved optimization, but worthwhile if it makes a significant impact on big loads.)

src/main/clojure/clara/rules/memory.cljc
mrrodriguez: I see what you're saying and I think this whole situation here is actually pretty sneaky.
I gave these a mutable flag to avoid more object allocation down in `remove-activations!`.

If you look carefully at [1] only local instances of RuleOrderedActivation are ever created that would mutate this field at all.  This means only a single thread would ever have one of these at a time.  The object is then discarded entirely.

`add-activations!` adds the only RuleOrderedActivation instances that would be able to be seen by multiple threads.  And those RuleOrderedActivation are never changed by having `using-token-identity` called on them.  They would always have this `use-token-identity?` field set to false (the default in `->rule-ordered-activation`).

The only reason the [2] RuleOrderedActivation.equals() implementation looks at this field on both `this` and `other` here is that I didn't want to make any assumptions about what object equals() was called via the PriorityQueue we are using.

I could have used a different object type to represent the "mutable" - strictly only local object types and the one that is shared with other threads later in memory.  However, I thought that'd also be clunky anyways and just more code.

This certainly needs some comments, I agree.  I can work on adding a few.

[1] https://github.com/rbrush/clara-rules/pull/214/files#diff-79e3c7ecc366c352f808f2a796662a30R806
[2] https://github.com/rbrush/clara-rules/pull/214/files#diff-79e3c7ecc366c352f808f2a796662a30R352

mrrodriguez: I added a comment above `using-token-identity!` describing this situation.  It isn't ideal that this even needs a comment, but I think it is fine and works for now.

I also changed the name `using-token-identity` to `using-token-identity!` since the function most certainly would have to update some sort of mutable state to be useful.  I missed this convention on the first-pass.

Let me know if you have any further issues with this right now.

rbrush:
Merged! Thanks, Mike!

PR #216, opened 2016-08-16
Fixing regression from #202

mrrodriguez:
- AccumulateNode needs to accumulate once per fact-binding group instead of sharing an accumulated result across fact-binding groups
- Added test to expose the issue

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -713,16 +713,15 @@
       (mem/add-tokens! memory node join-bindings tokens)
       
       (cond
-        ;; If there are previously accumulated results to propagate, use them.  If
-        ;; this is the first time there are matching tokens, then the reduce will
-        ;; have to happen for the first time.  However, this reduce operation is
-        ;; independent of the specific tokens since the elements join to the tokens
-        ;; via pre-computed hash join bindings for this node.  So only reduce once,
-        ;; for all tokens.
+        ;; If there are previously accumulated results to propagate, use them.  If this is the
+        ;; first time there are matching tokens, then the reduce will have to happen for the
+        ;; first time.  However, this reduce operation is independent of the specific tokens
+        ;; since the elements join to the tokens via pre-computed hash join bindings for this
+        ;; node.  So only reduce once per binding grouped facts, for all tokens. This includes
+        ;; all bindings, not just the join bindings.
         has-matches?
-        (doseq [:let [;; There is at least one previous result due to has-matches?.
-                      [_ [previous previous-reduced]] (first previous-results)
-                      first-reduce? (= ::not-reduced previous-reduced)
+        (doseq [[fact-bindings [previous previous-reduced]] previous-results
+                :let [first-reduce? (= ::not-reduced previous-reduced)
                       previous-reduced (if first-reduce?
                                          ;; Need to accumulate since this is the first time we have
                                          ;; tokens matching so we have not accumulated before.
@@ -731,8 +730,7 @@
                       accum-reduced (when first-reduce?
                                       [previous previous-reduced])
                       converted (when (some? previous-reduced)
-                                  (convert-return-fn previous-reduced))]
-                [fact-bindings [previous previous-reduced]] previous-results]
+                                  (convert-return-fn previous-reduced))]]
           
           ;; Newly accumulated results need to be added to memory.
           (when first-reduce?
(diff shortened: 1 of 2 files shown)


src/test/clojure/clara/test_rules.clj
```diff
@@ -907,6 +907,41 @@
            (set batch-inserts)
            (set single-inserts)))))
 
+(def test-accum-needing-token-partitions-correctly-on-fact-binding
+  (let [q (dsl/parse-query []
+                           [[WindSpeed]
+                            [?ts <- (acc/all) :from [Temperature (= ?loc location)]]])
```
WilliamParker: I suggest making a rule that uses a filter join (and thus exercises AccumulateWithJoinFilterNode) and then doseq'ing over the two rules in a manner similar to the accumulator tests I've added recently.  It is a little extra code, but since the assertions will be the same and the rule structure should be the same the extra mental overhead for someone reading these tests should be fairly minimal. You might need different facts; basically you'd want a field that is ignored in the AccumulateNode rule but used to join in the AccumulateWithJoinFilterNode rule and have that field be equal in both cases.  Having an unused field be equal in the AccumulateNode case shouldn't do any harm there.

mrrodriguez: I can do that.  The only hesitation I have to adding stuff like that is it can obscure what the test is trying to test.  However, I can see the argument that it is good to get these sort of cases tested for all variations of accumulate nodes.

mrrodriguez: I've updated this test so that the AccumulateWithJoinFilterNode is also tested for this scenario.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -800,15 +792,20 @@
...
             ;; that come from their alpha nodes. Keep in mind that these element groups need elements to be created
             ;; and cannot come from initial values if they have bindings that are not shared with tokens.
-            [fact-bindings [previous previous-reduced]] previous-results]
+            [fact-bindings [previous previous-reduced]] previous-results
+            :let [;; If there were tokens before that are now removed, the value would have been accumulated already.
+                  ;; This means there is no need to check for ::not-reduced here.
+                  previous-converted (when (some? previous-reduced)
+                                       ((:convert-return-fn accumulator) previous-reduced))]
+            ;; A nil previous result should not have been propagated before.
+            :when (some? previous-converted)
```
WilliamParker: Is this part tested somewhere else?  I don't see anything you added in the tests covering this scenario.

src/test/clojure/clara/test_rules.clj
```diff
@@ -907,6 +907,49 @@
...
+                                                  fire-rules
+                                                  (query q)
+                                                  frequencies)
+                  left-activates-before-right (-> s
+                                                  (insert (->WindSpeed 10 "MCI"))
+                                                  (insert t1mci
+                                                          t2lax)
+                                                  fire-rules
+                                                  (query q)
+                                                  frequencies)
```
WilliamParker: You're testing the left-activate of AccumulateNode fix here, but what about the mirror fix in left-retract of AccumulateNode?

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -770,18 +767,13 @@
...
-                  ;; pairs retracted.
-                  [_ [_ previous-reduced]] (first previous-results)
-                  previous-converted (when (some? previous-reduced)
-                                       ((:convert-return-fn accumulator) previous-reduced))
-                  removed-tokens (mem/remove-tokens! memory node join-bindings tokens)
+    (doseq [:let [removed-tokens (mem/remove-tokens! memory node join-bindings tokens)
                   remaining-tokens (mem/get-tokens memory node join-bindings)
 
+                  ;; Note:  Memory *must* be read here before the memory is potentially cleared in the
+                  ;; following lines.
```
WilliamParker: +1 to commenting on things like this. :)

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -713,16 +713,14 @@
...
-        ;; this is the first time there are matching tokens, then the reduce will
-        ;; have to happen for the first time.  However, this reduce operation is
-        ;; independent of the specific tokens since the elements join to the tokens
-        ;; via pre-computed hash join bindings for this node.  So only reduce once,
-        ;; for all tokens.
+        ;; If there are previously accumulated results to propagate, use them.  If this is the
+        ;; first time there are matching tokens, then the reduce will have to happen for the
+        ;; first time.  However, this reduce operation is independent of the specific tokens
+        ;; since the elements join to the tokens via pre-computed hash join bindings for this
+        ;; node.  So only reduce once per group of facts, for all tokens.
```
WilliamParker: You might specifically note that we group on all bindings, not just the join bindings.

WilliamParker:
I'd like some more clarity on https://github.com/rbrush/clara-rules/pull/216/files#r75279029 and I think the testing could use some improvement.  Otherwise +1.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: I'm not sure but if the tokens were empty it would be broken.  It is hard to get a scenario that does that.  I know it is definitely possible to get `right-retract` with elements that do not remove anything.  I'm not 100% sure if that is true on the `left-retract` side.  I don't really think it is worth dumping a lot of time in right now since the `seq` check is cheap enough anyways.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: How about making it
"So only reduce once per binding grouped facts, for all tokens.  This includes all bindings, not just the join bindings."

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: I actually broke things at one point by flipping the two lines.  So I put the comment to avoid that happening again.  :P

src/test/clojure/clara/test_rules.clj
mrrodriguez: I am adding cases for that to this test.

mrrodriguez:
I have addressed all the comments I've seen at this point.

rbrush:
Merged!

PR #220, opened 2016-09-02
Added listening capability to alpha-activate and alpha-retract in alpha network

kulkarnipushkar:
Added a listening capability to alpha-activate and alpha-retract, so that we can listen to the facts added or retracted the alpha network. We need it internally to know which facts that were inserted or retracted, matched some of the alpha nodes.
Listener on insert-facts and retract-facts will not give me facts that are guaranteed to match something in the alpha network.
I am not appending the trace for alpha-activate and alpha-retract in tracing listener, as it will add redundant information from insert-facts and retract-facts.

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -329,25 +329,30 @@
 (defrecord AlphaNode [env children activation]
   IAlphaActivate
   (alpha-activate [node facts memory transport listener]
-    (send-elements
-     transport
-     memory
-     listener
-     children
-     (for [fact facts
-           :let [bindings (activation fact env)] :when bindings] ; FIXME: add env.
-       (->Element fact bindings))))
+    (let [fact-binding-pairs (for [fact facts
+                                   :let [bindings (activation fact env)] :when bindings] ; FIXME: add env.
+                               [fact bindings])]
+      (l/alpha-activate! listener node (map first fact-binding-pairs))
+      (send-elements
+        transport
+        memory
+        listener
+        children
+        (for [[fact bindings] fact-binding-pairs]
+          (->Element fact bindings)))))
 
   (alpha-retract [node facts memory transport listener]
-
-    (retract-elements
-     transport
-     memory
-     listener
-     children
-     (for [fact facts
-           :let [bindings (activation fact env)] :when bindings] ; FIXME: add env.
-       (->Element fact bindings)))))
+    (let [fact-binding-pairs (for [fact facts
+                                   :let [bindings (activation fact env)] :when bindings] ; FIXME: add env.
+                               [fact bindings])]
+      (l/alpha-retract! listener node (map first fact-binding-pairs))
+      (retract-elements
+        transport
+        memory
+        listener
+        children
+        (for [[fact bindings] fact-binding-pairs]
+          (->Element fact bindings))))))
 
 (defrecord RootJoinNode [id condition children binding-keys]
   ILeftActivate
--- a/src/main/clojure/clara/rules/listener.cljc
+++ b/src/main/clojure/clara/rules/listener.cljc
@@ -12,8 +12,10 @@
   (right-activate! [listener node elements])
   (right-retract! [listener node elements])
   (insert-facts! [listener facts])
+  (alpha-activate! [listener node facts])
   (insert-facts-logical! [listener node token facts])
   (retract-facts! [listener facts])
+  (alpha-retract! [listener node facts])
   (retract-facts-logical! [listener node token facts])
   (add-accum-reduced! [listener node join-bindings result fact-bindings])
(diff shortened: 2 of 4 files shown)


src/main/clojure/clara/rules/engine.cljc
```diff
@@ -329,6 +329,7 @@
 (defrecord AlphaNode [env children activation]
   IAlphaActivate
   (alpha-activate [node facts memory transport listener]
+    (l/alpha-activate! listener node facts)
```
WilliamParker: I think you actually want to use the facts that make it through the activation function.  If you look at https://github.com/rbrush/clara-rules/blob/0.12.0/src/main/clojure/clara/rules/compiler.clj#L204 which is where I believe the activation ultimately comes from, you'll see that it can return nil.  The existing logic filters out facts where the activation returns nil from being passed downstream:

````
(for [fact facts
           :let [bindings (activation fact env)] :when bindings] ; FIXME: add env.
       (->Element fact bindings))
````

So, for example, if you have something like

````
(defrule hot-temperatures
      [Temperature (> temperature 100)]
      ;; other-conditions
    => ... )
````

Temperatures that aren't over 100 can just be filtered out in the alpha node.  Right now you're calling the listener on all insertions whose type matched some alpha node, not all insertions that were passed downstream from an alpha node.

**Comment by WilliamParker** [comment]
To elaborate a bit (and please correct me if I'm out of date; we discussed this a while ago) the use-case here is ultimately to be able to retract facts when all we have is a unique ID of the fact to retract, not the fact itself.  This support in the listener protocol would allow us to write a listener that will keep track of which relevant facts matched something in the network; if they didn't match anything there is no need to retract them later, since they never had any impact on the session in the first place.  If we can know that there is no need to retract a fact later, then that fact doesn't need to be serialized when persisting the session.  The simplest approach would be to simply keep track of all insertions, but this would greatly increase the size of a serialized session and thus the time required to deserialize it.

This isn't inherently specific to our use-case though.  I think it could be helpful for anyone else who has specialized truth maintenance needs that are best handled in user code (while keeping the core Clara API uncluttered with their details).

**Comment by WilliamParker** [comment]
Regarding tracing, personally I'd be inclined to add it.  It isn't too difficult to filter the trace elements down to what you're looking for in a REPL; as a practical matter it is large enough in the real use-cases I've seen that reading through the entire output manually isn't very practical.  I don't think it is a big deal either way though.

File: src/main/clojure/clara/rules/engine.cljc [hunk 2]
```diff
@@ -329,6 +329,7 @@
 (defrecord AlphaNode [env children activation]
```
WilliamParker: I'd be inclined to add a test that the listener is called correctly since this would be a functional part of Clara's API.

mrrodriguez: There are a lot of missing listener calls (a good thing to fix in a future commit). So I don't necessarily see this one as being any different than that.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: You're right here. I missed this when looking at it the first time.

mrrodriguez:
>To elaborate a bit (and please correct me if I'm out of date; we discussed this a while ago) the use-case here is ultimately to be able to retract facts when all we have is a unique ID of the fact to retract, not the fact itself. This support in the listener protocol would allow us to write a listener that will keep track of which relevant facts matched something in the network; if they didn't match anything there is no need to retract them later, since they never had any impact on the session in the first place. If we can know that there is no need to retract a fact later, then that fact doesn't need to be serialized when persisting the session. The simplest approach would be to simply keep track of all insertions, but this would greatly increase the size of a serialized session and thus the time required to deserialize it.

>This isn't inherently specific to our use-case though. I think it could be helpful for anyone else who has specialized truth maintenance needs that are best handled in user code (while keeping the core Clara API uncluttered with their details).

This is right.  Clara could provide more functionality to meet this need, however, we are shooting for a minimal "hook" into this right now.  I think it'd be worth considering taking it steps further perhaps in the future.

mrrodriguez:
>Regarding tracing, personally I'd be inclined to add it. It isn't too difficult to filter the trace elements down to what you're looking for in a REPL; as a practical matter it is large enough in the real use-cases I've seen that reading through the entire output manually isn't very practical. I don't think it is a big deal either way though.

The `insert-facts!` and `retract-facts!` would be duplicated, if not even more than 1 time (depending on how many alpha nodes a single fact matched).  So the tracing output just seemed like it could get into quite a bit of spam with this.  Perhaps that is ok.  I've been torn on that one.

kulkarnipushkar:
I have updated the commit/PR with provided suggestions.

rbrush:
LGTM. Merged!

PR #222, opened 2016-09-15
Add listener calls to all nodes

WilliamParker:
I've found tracing to be quite useful, even with a limited subset of nodes making the listener calls.  I've ended up adding tracing calls in my local branch before to solve this.  Note that since the node is available to the listener calls, either the listener or processing on data collected by the listener can filter on the things that are relevant to a particular use-case.  I think there is some future work to be done here in tracing; for example, I'd like to have a helper function that gets only the tracing for descendants of alpha nodes for a particular type.  The first step is to have all the information available though.  This could break any listeners that depend on only a subset of nodes having tracing, but I think the correct way to do that sort of logic is for the listener to dispatch on the node given.  The information in the session object (e.g. alpha-roots and id-to-node) should supply any necessary information to a consumer for such dispatching logic.

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -408,6 +408,7 @@
   (left-activate [node join-bindings tokens memory transport listener]
     ;; Add token to the node's working memory for future right activations.
     (mem/add-tokens! memory node join-bindings tokens)
+    (l/left-activate! listener node tokens)
     (send-tokens
      transport
      memory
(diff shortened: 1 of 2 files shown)


mrrodriguez:
+1

This is helpful.  I've added these changes manually before when I was troubleshooting via the tracer.

WilliamParker:
I've rebased this on top of the changes from https://github.com/rbrush/clara-rules/pull/223

rbrush:
Merged. Thanks, Will!

PR #227, opened 2016-10-12
Fix clara.rules.engine/flush-rhs-insertions! to remove duplicate call…

WilliamParker:
…s to clara.rules.listener/retract-facts!

Currently a fact retracted in a rule RHS will show up in a tracing listener once per matching alpha node, while a fact externally retracted will show up once.  [1]  I introduced this inconsistency in https://github.com/rbrush/clara-rules/pull/173/files This pull request makes the two retraction types consistent by making a single RHS retraction listener call.  If a user wants information on activation of particular alpha nodes that information is directly available now after https://github.com/rbrush/clara-rules/pull/220

1. https://github.com/rbrush/clara-rules/blob/0.13.0-RC2/src/main/clojure/clara/rules/engine.cljc#L1513

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -233,11 +233,12 @@
     ;; Update the count so the rule engine will know when we have normalized.
     (swap! insertions + (count facts))
 
+    (when listener
+      (l/retract-facts! listener facts))
+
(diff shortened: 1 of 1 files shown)


mrrodriguez:
+1

rbrush:
Merged!

PR #228, opened 2016-10-13
Sorted coll Fressian handler fixes

mrrodriguez:
The sorted collections Fressian SerDe handlers didn't preserve metadata upon deserialization.  This means a deserialized sorted collection could potentially fail if it were re-serialized again afterwards.  This only applies to sorted collections that have a custom sort comparator fn, which must have its name specified in the metadata under the key `:clara.rules.durability/comparator-name`.

I added the start of some explicit testing of the behavior of the Fressian handlers in `clara.test-fressian` to try to be sure to catch more issues like this in the future.

I also added a test in `clara.test-durability` demonstrating the use-case that triggered this issue from the Serialize+Deserialize+Re-serialize a rulebase perspective.

`clara.rules.durability/assemble-restored-session` had a 2-arity added as well.  This is only used in tests here, but we have experienced places where it'd be helpful for our use-cases as well.  If you deserialize a rulebase and want a brand new working memory, rather than having an existing one, it is convenient to have this assembly function that behaves sort of like `clara.rules/mk-session`, but with a rulebase given as an argument - along with the same sort of options.

### DIFF
--- a/src/main/clojure/clara/rules/durability.clj
+++ b/src/main/clojure/clara/rules/durability.clj
@@ -401,60 +401,84 @@
                 :activation-group-fn
                 :alphas-fn))))
 
-(defn create-default-get-alphas-fn [rulebase]
-  (@#'com/create-get-alphas-fn type ancestors rulebase))
+(def ^:private create-get-alphas-fn @#'com/create-get-alphas-fn)
+
+(defn- opts->get-alphas-fn [rulebase opts]
+  (let [fact-type-fn (:fact-type-fn opts type)
+        ancestors-fn (:ancestors-fn opts ancestors)]
+    (create-get-alphas-fn fact-type-fn
+                          ancestors-fn
+                          rulebase)))
 
 (defn assemble-restored-session
-  "Builds a Clara session from the given rulebase and memory components.
+  "Builds a Clara session from the given rulebase and memory components.  When no memory is given a new 
+   one is created with all of the defaults of eng/local-memory.
    Note!  This function should not typically be used.  It is left public to assist in ISessionSerializer 
           durability implementations.  Use clara.rules/mk-session typically to make rule sessions.
    
    Options can be provided via opts.
    These include:
 
+   * :fact-type-fn
+   * :ancestors-fn
    * :activation-group-sort-fn 
    * :activation-group-fn
-   * :get-alphas-fn 
 
    If the options are not provided, they will default to the Clara session defaults.
    These are all described in detail in clara.rules/mk-session docs.
 
    Note!  Currently this only supports the clara.rules.memory.PersistentLocalMemory implementation
           of memory."
-  [rulebase memory opts]
-  (let [opts (-> opts
-                 (assoc :rulebase rulebase)
-                 ;; Right now activation fns do not serialize.
-                 (update :activation-group-sort-fn
-                         #(eng/options->activation-group-sort-fn {:activation-group-sort-fn %}))
-                 (update :activation-group-fn
-                         #(eng/options->activation-group-fn {:activation-group-fn %}))
-                 ;; TODO: Memory doesn't seem to ever need this or use it.  Can we just remove it from memory?
-                 (update :get-alphas-fn
-                         #(or % (create-default-get-alphas-fn rulebase))))
+  ([rulebase opts]
+   (let [opts (assoc opts
+                     :rulebase
+                     rulebase
+                     :get-alphas-fn
+                     (opts->get-alphas-fn rulebase opts))
+         {:keys [listeners transport get-alphas-fn]} opts]
+     
+     (eng/assemble {:rulebase rulebase
+                    :memory (eng/local-memory rulebase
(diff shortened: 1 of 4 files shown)


src/test/clojure/clara/test_durability.clj
```diff
@@ -254,4 +254,24 @@
...
+          s (mk-session 'clara.durability-rules)
+          rb (-> s eng/components :rulebase)
+          deserialized1 (rb-serde s)
+          ;; Need a session to do the 2nd round of SerDe.
+          restored (d/assemble-restored-session deserialized1 {})
+          deserialized2 (rb-serde restored)]
+      
+      (is (= (-> rb :id-to-node keys set)
+             (-> deserialized1 :id-to-node keys set)
+             (-> deserialized2 :id-to-node keys set))))))
```
WilliamParker: Perhaps you could factor out the parts of durability-test that are just normal rule operations and assertions from the serialization parts and assert on the former here too?  This doesn't seem to do much to validate functional correctness after the second pass through serialization rather than just validating that the second deserialization doesn't cause an exception.  Functional correctness depends on the node structure itself; I don't even think the id-to-node map is used in the engine.

WilliamParker:
I noticed https://github.com/rbrush/clara-rules/issues/230 while reviewing this, but it looks like we have the same issue before and after these changes.

WilliamParker:
I'm not very knowledgeable about Fressian, but the write and read handler changes look appropriate.

src/test/clojure/clara/test_durability.clj
mrrodriguez: Yeah I just did this to get a quick test.  It wasn't meant to be much more than asserting that something structurally coherent can serialize and deserialize.  I know the `:id-to-node` doesn't really prove much of anything in terms of the correctness of a rulebase.  

Prior to these changes there were exceptions before even getting this far.  I can try to work in a test about inserting facts to these rulebases and showing they all have the same queried results.

mrrodriguez:
I'll work in #230 to this I believe.

mrrodriguez:
I pushed changes to this.  More testing was added and clean up of the support of the `mk-session` style options given to `clara.rules.durability/assemble-restored-session`.

This also should be addressing #230 which just came as a side-effect of fixing up this function in its various use-cases.

src/main/clojure/clara/rules/durability.clj
```diff
@@ -401,60 +401,87 @@
                 :activation-group-fn
                 :alphas-fn))))
 
-(defn create-default-get-alphas-fn [rulebase]
-  (@#'com/create-get-alphas-fn type ancestors rulebase))
+(def ^:private create-get-alphas-fn @#'com/create-get-alphas-fn)
+
+(defn- opts->get-alphas-fn [rulebase opts]
```
WilliamParker: I don't think this is quite right either, and is unnecessarily complex.  How about

````
(defn opts->get-alphas-fn [rulebase opts]
       (let [fact-type-fn (or (:fact-type-fn opts) type)
              ancestors-fn (or (:ancestors-fn opts) ancestors)]
       (create-get-alphas-fn fact-type-fn ancestors-fn rulebase)))
````

Currently this doesn't allow the user to specify the ancestors-fn without overriding the fact-type-fn as well.  More generally, I don't think we need to intermingle the "default or override" selection logic with the call to actually create the get-alphas-fn; doing so just creates unnecessary complexity.

Comment (mrrodriguez) [inline src/main/clojure/clara/rules/durability.clj]: Yeah, you're right.  I think it is unlikely to override one and not the other, but obviously possible.
Either way, the implementation like what you are saying here makes more sense.  I just didn't put enough thought into this part since it wasn't a usage I was concerned with.

File: src/test/clojure/clara/test_durability.clj [hunk 3]
```diff
@@ -254,4 +306,72 @@
...
+          deserialized1 (rb-serde s)
+          ;; Need a session to do the 2nd round of SerDe.
+          restored1 (d/assemble-restored-session deserialized1 {})
+          deserialized2 (rb-serde restored1)
+          restored2 (d/assemble-restored-session deserialized2 {})
+
+          init-qresults (:query-results (session-test s))
+          restored-qresults1 (:query-results (session-test restored1))
+          restored-qresults2 (:query-results (session-test restored2))]
+
```
mrrodriguez: Yes.  Obviously it doesn't test total correctness of the session or anything, but gives reasonable confidence since the results end up coming out the same after inserting the test facts, firing, and querying.

The working memory style test uses this same test, but takes it a step further and checks the object-identity found within the query results etc.

src/main/clojure/clara/rules/durability.clj
mrrodriguez: They are pushed now.

WilliamParker: +1

WilliamParker:
+1

nrayapati:
+1

rbrush:
Merged!

PR #232, opened 2016-10-18
Issue 231: NegationNode should only retract downstream upon a right-a…

WilliamParker:
…ctivation if there were no elements in the binding group previously


WilliamParker:
Closing this pull request from "issue231-NegationNode" in favor of https://github.com/rbrush/clara-rules/pull/233 which fixes both NegationNode and NegationWithJoinFilterNode.  This is discussed more fully on https://github.com/rbrush/clara-rules/issues/231

PR #233, opened 2016-10-21
Issue 231: Negation nodes should only retract a token downstream upon…

WilliamParker:
… right-activation with matching elements if there were no matching elements previously

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -594,10 +594,16 @@
 
   IRightActivate
   (right-activate [node join-bindings elements memory transport listener]
-    (l/right-activate! listener node elements)
-    (mem/add-elements! memory node join-bindings elements)
-    ;; Retract tokens that matched the activation, since they are no longer negatd.
-    (retract-tokens transport memory listener children (mem/get-tokens memory node join-bindings)))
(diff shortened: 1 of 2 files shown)


src/main/clojure/clara/rules/engine.cljc
```diff
@@ -659,19 +665,32 @@
...
-
-                          :when (matches-some-facts? token
-                                                     elements
-                                                     join-filter-fn
-                                                     condition)]
-                      token)))
+    ;; Put the previous elements into a persistent collection since LocalMemory will return
+    ;; a mutable list here on the JVM, and we need to examine the previous elements on
+    ;; a per-token basis here.
+    (let [previous-elements (into [] (mem/get-elements memory node join-bindings))]
```
mrrodriguez: Why not avoid this copying of data and instead just hold off on the `mem/add-elements!` until after you are done looking at the `previous-elements`?
You could pull the `for` out of the `retract-tokens` and instead bind to the results of the `for` in the same `let`-binding as the `previous-elements`.  So then the `mem/add-elements!` wouldn't happen until after.

The issue to me just appears to be the delay in "looking" at the `previous-elements`.

WilliamParker: +1  I've updated the pull request.

mrrodriguez:
+1

rbrush:
Merged!

PR #235, opened 2016-11-14
Issue 234: Share the get-alphas-fn across deserialized sessions

WilliamParker:


src/main/clojure/clara/rules/compiler.clj
```diff
@@ -64,7 +64,13 @@
                         ;; Map of queries to the nodes hosting them.
                         query-nodes :- {sc/Any QueryNode}
                         ;; Map of id to one of the beta nodes (join, accumulate, etc).
-                        id-to-node :- {sc/Num BetaNode}])
+                        id-to-node :- {sc/Num BetaNode}
+                        ;; Function for sorting activation groups of rules for firing.
+                        activation-group-sort-fn
+                        ;; Function that takes a rule and returns its activation group.
+                        activation-group-fn
+                        ;; Function that takes facts and determines what alpha nodes they match.
```
mrrodriguez: Small detail:  Doesn't it just take "a fact" instead of "facts"

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1576,6 +1534,60 @@
...
+        get-alphas-fn (create-get-alphas-fn fact-type-fn ancestors-fn alpha-map)]
+
+    (strict-map->Rulebase
+     {:alpha-roots alpha-map
+      :beta-roots beta-roots
+      :productions productions
+      :production-nodes production-nodes
+      :query-nodes query-map
+      :id-to-node id-to-node
+      :get-alphas-fn get-alphas-fn
```
mrrodriguez: It doesn't really matter, but it'd make more sense to me to keep these keys in the same order as the basis fields of the Rulebase record.

src/main/clojure/clara/rules/durability.clj
```diff
@@ -392,40 +392,35 @@
...
-                          rulebase)))
+                          (:alpha-roots rulebase))))
 
 (defn assemble-restored-session
   "Builds a Clara session from the given rulebase and memory components.  When no memory is given a new 
    one is created with all of the defaults of eng/local-memory.
    Note!  This function should not typically be used.  It is left public to assist in ISessionSerializer 
           durability implementations.  Use clara.rules/mk-session typically to make rule sessions.
-   
-   Options can be provided via opts.
```
mrrodriguez: This doc doesn't seem accurate.  What `opts` are still accepted?  It looks like `:rulebase`?

src/main/clojure/clara/rules/durability.clj
```diff
@@ -392,40 +392,35 @@
...
-
-   * :fact-type-fn
-   * :ancestors-fn
-   * :activation-group-sort-fn 
-   * :activation-group-fn
 
    If the options are not provided, they will default to the Clara session defaults.
-   These are all described in detail in clara.rules/mk-session docs.
+   These are all described in detail in clara.rules/mk-session docs.  Note that the :activation-group-fn,
+   :activation-group-sort-fn, :fact-type-fn, and :ancestors-fn are attached to the rulebase, not the session,
```
mrrodriguez: Why would they provided here?  It doesn't make sense to even mention them anymore.

This API is experimental.  I do not think we should try to accept arguments that don't make sense anymore.

src/main/clojure/clara/rules/durability.clj
```diff
@@ -440,45 +435,27 @@
...
                     :memory (eng/local-memory rulebase
                                               (clara.rules.engine.LocalTransport.)
-                                              (eng/options->activation-group-sort-fn opts)
-                                              (eng/options->activation-group-fn opts)
+                                              (:activation-group-sort-fn rulebase)
+                                              (:activation-group-fn rulebase)
                                               ;; TODO: Memory doesn't seem to ever need this or use
                                               ;; it.  Can we just remove it from memory?
-                                              get-alphas-fn)
+                                              (:get-alphas-fn rulebase))
```
mrrodriguez: We shouldn't call `opts->get-alphas-fn` on the given `opts` anymore above either.  Relates to
https://github.com/cerner/clara-rules/pull/235/files#r87932559

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -535,17 +539,34 @@
           indexed-facts))))
 
   (deserialize [_ mem-facts opts]
+    ;; Options can be provided via opts.
```
mrrodriguez: This seems like it should be documented as doc strings at the protocol function `deserialize` level in `clara.rules.durability`.

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -535,17 +539,34 @@
...
+            maybe-base-rulebase (when (and (not rulebase-only?) base-rulebase)
+                                  base-rulebase)
+
+            rulebase (if maybe-base-rulebase
+                       maybe-base-rulebase
+                       (let [without-ops-rulebase (binding [d/*node-id->node-cache* (volatile! {})
+                                                            d/*clj-record-holder* record-holder
+                                                            d/*compile-expr-fn* (memoize (fn [id expr] (com/try-eval expr)))]
+                                                    (fres/read-object rdr))]
+                         (assoc without-ops-rulebase
```
mrrodriguez: It may be more convenient to have a `(clara.rules.durability/assemble-restored-rulebase [opts])` to do this part and that is analogous to `clara.rules.durability/assemble-restored-session`.

Thoughts?

src/test/clojure/clara/test_durability.clj
```diff
@@ -336,12 +339,13 @@
 (deftest test-assemble-restored-session-opts
   (let [orig (mk-session 'clara.durability-rules)
 
-        test-assemble (fn [rulebase memory]
-                        (let [activation-group-fn-called? (volatile! false)
+        test-assemble (fn [orig assemble-with-memory?]
+                        (let [memory (.memory orig)
```
mrrodriguez: I'd prefer to avoid the interop/reflection and instead go with 
`(-> orig eng/components :memory)`

It is more the public API anyways.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: We're talking about the get-alphas-fn correct?  That takes multiple facts.

https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules/compiler.clj#L1554

src/main/clojure/clara/rules/compiler.clj
WilliamParker: +1 I'll change this accordingly.

src/main/clojure/clara/rules/durability.clj
WilliamParker: Right, I'm not trying to accept them, more calling out that the API has changed.  However, since is a *highly* experimental API perhaps you're right that calling it out specifically is just noise.

src/main/clojure/clara/rules/durability.clj
WilliamParker: It is still possible to supply listeners and an ITransport implementation.  I doubt the latter is being used, although it is certainly possible, but the former seems quite normal.  Perhaps it would be good to enumerate the accepted options and remove the reference to things not accepted below as discussed there.

src/main/clojure/clara/rules/durability.clj
WilliamParker: +1, that can be removed.  With these changes it isn't even used anymore; it is just a dead code path.

src/main/clojure/clara/rules/durability/fressian.clj
WilliamParker: Do you just mean the step of associating the new options onto the rulebase?  I suppose I could move it, I don't think it matters much either way though to be honest.  I think this overall API could probably use some more extensive cleanup and I don't see much to be gained by moving this prior to such cleanup.

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: You're right, sorry for the confusion.

src/main/clojure/clara/rules/durability/fressian.clj
mrrodriguez: Yes, I meant pulling that part out.  The idea is to have things that are SerDe implementation independent in `clara.rules.durability`.  So it'd be a start.

I'd like to have the implementation of Fressian SerDe to do only things that are specific to Fressian as much as possible.

However, it's mostly just organizational.  It wouldn't be expected that there'd be a lot of different implementations of `d/ISessionSerializer` since it is highly tied to the impl details.  However, if we ended up wanting to add more than just Fressian support, it'd be nice to have the implementations as thin as possible.

WilliamParker:
@mrrodriguez I've updated the PR in response to your comments.  I don't plan to merge this today in any case though since I'll leaving for the day shortly and would like to look over it again after fixing things but before merging.

mrrodriguez:
+1 to changes.

rbrush:
+1; LGTM.

I'll let one of you guys merge this one. ;)

PR #242, opened 2016-11-28
Issue 236: Optimize the get-alphas-fn for cases where there are multi…

WilliamParker:
…ple primary fact types with a common ancestor that has corresponding productions in the rulebase

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1487,52 +1487,99 @@
         env (assoc :env env))
       {:alpha-expr alpha-expr})))
 
+;; Wrap the fact-type so that Clojure equality and hashcode semantics are used
+;; even though this is placed in a Java map.
+(deftype AlphaRootsWrapper [roots roots-hash]
+  Object
+  (equals [this other]
+    ;; Since we memoize the function that retrieves these on the fact type, and since the alpha roots for
+    ;; any given type primary type T (excluding descendants and ancestors) are a disjoint
+    ;; set with respect to the alpha roots for any other primary type T2, we know that we will
+    ;; only have one AlphaRootsWrapper object for any given set of alpha roots per get-alphas-fn instance.
+    ;; Therefore we can just use reference equality in the interest of maximizing performance.
+    (identical? this other))
+  ;; Since know we will need to find the hashcode of this object in all cases just eagerly calculate it upfront
+  ;; and avoid extra calls to hash later.  The hashcode of the Clojure vector of roots would be cached but there
+  ;; is no need to even make the call to that cache.
+  (hashCode [this] roots-hash))
+
 (defn- create-get-alphas-fn
   "Returns a function that given a sequence of facts,
   returns a map associating alpha nodes with the facts they accept."
   [fact-type-fn ancestors-fn alpha-roots]
 
-  ;; We preserve a map of fact types to alpha nodes for efficiency,
-  ;; effectively memoizing this operation.
-  (let [alpha-map (atom {})
-
-        ;; If a customized fact-type-fn is provided,
+  (let [ ;; If a customized fact-type-fn is provided,
         ;; we must use a specialized grouping function
         ;; that handles internal control types that may not
         ;; follow the provided type function.
-        fact-grouping-fn (if (= fact-type-fn type)
-                           type
-                           (fn [fact]
-                             (if (isa? (type fact) :clara.rules.engine/system-type)
-                               ;; Internal system types always use Clojure's type mechanism.
-                               (type fact)
-                               ;; All other types defer to the provided function.
-                               (fact-type-fn fact))))]
+        wrapped-fact-type-fn (if (= fact-type-fn type)
+                               type
+                               (fn [fact]
+                                 (if (isa? (type fact) :clara.rules.engine/system-type)
+                                   ;; Internal system types always use Clojure's type mechanism.
+                                   (type fact)
+                                   ;; All other types defer to the provided function.
+                                   (fact-type-fn fact))))
+
+        ;; Wrap the ancestors-fn so that we don't send internal facts such as NegationResult
+        ;; to user-provided productions.  Note that this work is memoized inside fact-type->roots.
+        wrapped-ancestors-fn (fn [fact-type]
+                               (if (isa? fact-type :clara.rules.engine/system-type)
+                                 ;; Exclude system types from having ancestors for now
+                                 ;; since none of our use-cases require them.  If this changes
+                                 ;; we may need to define a custom hierarchy for them.
+                                 #{}
+                                 (ancestors-fn fact-type)))
+
+        fact-type->roots (memoize
+                          (fn [fact-type]
+                            ;; There is no inherent ordering here but we put the AlphaRootsWrapper instances
+                            ;; in a vector rather than a set to avoid nondeterministic ordering (and thus nondeterministic
+                            ;; performance).
+                            (into []
+                                  ;; If a given type in the ancestors has no matching alpha roots,
+                                  ;; don't return it as an ancestor.  Fact-type->roots is memoized on the fact type,
+                                  ;; but work is performed per group returned on each call the to get-alphas-fn.  Therefore
+                                  ;; removing groups with no alpha nodes here will improve performance on subsequent calls
+                                  ;; to the get-alphas-fn with the same fact type.
+                                  (keep #(when-let [roots (not-empty (get alpha-roots %))]
+                                           (AlphaRootsWrapper. roots (hash roots))))
+                                  ;; If a user-provided ancestors-fn returns a sorted collection, for example for
+                                  ;; ensuring determinism, we respect that ordering here by conj'ing on to the existing
+                                  ;; collection.
+                                  (conj (wrapped-ancestors-fn fact-type) fact-type))))
+        
+        update-roots->facts! (fn [^java.util.Map roots->facts roots-group fact]
+                               (if-let [v (.get roots->facts roots-group)]
+                                 (.add ^java.util.List v fact)
+                                 (.put roots->facts roots-group (doto (java.util.LinkedList.)
+                                                                  (.add fact)))))]
 
     (fn [facts]
-      (for [[fact-type facts] (platform/group-by-seq fact-grouping-fn facts)]
+      (let [roots->facts (java.util.LinkedHashMap.)]
 
-        (if-let [alpha-nodes (get @alpha-map fact-type)]
+        (doseq [fact facts
+                roots-group (fact-type->roots (wrapped-fact-type-fn fact))]
+          (update-roots->facts! roots->facts roots-group fact))
+        
+        (let [return-list (java.util.LinkedList.)
+              entries (.entrySet roots->facts)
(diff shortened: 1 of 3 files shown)


src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1507,32 +1519,62 @@
                                ;; Internal system types always use Clojure's type mechanism.
                                (type fact)
                                ;; All other types defer to the provided function.
-                               (fact-type-fn fact))))]
+                               (fact-type-fn fact))))
+
+        ;; Wrap the ancestors-fn so that we don't send internal facts such as NegationResult
+        ;; to user-provided productions.  Note that this work is memoized inside fact-type->roots.
+        wrapped-ancestors-fn (fn [fact-type]
```
mrrodriguez: If you name this "wrapped" should the name `fact-grouping-fn` be changed to `wrapped-fact-type-fn` for consistency?

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1507,32 +1519,62 @@
...
+          (update-roots->facts! roots->facts roots-group fact))
+        
+        (let [return-list (java.util.LinkedList.)]
+          ;; We iterate over the LinkedHashMap manually to avoid potential issues described at http://dev.clojure.org/jira/browse/CLJ-1738
+          ;; where a Java iterator can return the same entry object repeatedly and mutate it after each next() call.  We use mutable lists
+          ;; for performance but wrap them in unmodifiableList to make it clear that the caller is not expected to mutate these lists.
+          ;; Since after this function returns the only reference to the fact lists will be through the unmodifiedList we can depend elsewhere
+          ;; on these lists not changing.  Since the only expected workflow with these lists is to loop through them, not add or remove elements,
+          ;; we don't gain much from using a transient (which can be efficiently converted to a persistent data structure) rather than a mutable type. 
+          (doseq [^java.util.Map$Entry e (.entrySet roots->facts)]
```
mrrodriguez: `doseq` over the entry set will still be subject to the dangers described in the paragraph before since `doseq` is based around "seq"ing over the collection.

To avoid this, you need to explicitly manage a `loop` using manually managed `hasNext()` and `next()` calls.
The up-side of this is that you know you are getting maximum performance too when not going through the seq-abstraction.
```clj
(let [<etc>
      es (.entrySet root->facts)]

  (loop [e (when (.hasNext es)
             (.next es))]

    (.add return-list [(-> e ^AlphaRootsWrapper (.getKey) .roots)
                       (java.util.Collections/unmodifiableList (.getValue e))])

    (when (.hasNext es)
      (recur (.next es)))))
```

src/test/clojure/clara/test_rules.clj
```diff
@@ -1183,75 +1183,85 @@
...
-           (-> s
-               (fire-rules)
-               (query cold-not-match-temp))))
+        s (mk-session [cold-not-match-temp object-query] :cache false)
+        s-with-prior (mk-session [negation-with-prior-bindings object-query] :cache false)
+        s-with-nested (mk-session [nested-negation-with-prior-bindings object-query] :cache false)
+
+        no-system-types? (fn [session]
+                           (not-any? (fn [fact] (isa? (type fact) :clara.rules.engine/system-type))
+                                     (as-> session r
```
mrrodriguez: Why the name `r` to mean a `session`?

src/test/clojure/clara/test_rules.clj
```diff
@@ -1183,75 +1183,85 @@
...
-    (is (= [{}]
-           (-> s
-               (fire-rules)
-               (query cold-not-match-temp))))
+        s (mk-session [cold-not-match-temp object-query] :cache false)
+        s-with-prior (mk-session [negation-with-prior-bindings object-query] :cache false)
+        s-with-nested (mk-session [nested-negation-with-prior-bindings object-query] :cache false)
+
+        no-system-types? (fn [session]
+                           (not-any? (fn [fact] (isa? (type fact) :clara.rules.engine/system-type))
```
mrrodriguez: nit: could use `::eng/system-type` to take advantage of the ns alias.

src/test/clojure/clara/test_rules.clj
```diff
@@ -1183,75 +1183,85 @@
...
-                    (query negation-with-prior-bindings))))
+    (let [end-session (-> s-with-prior
+                          (insert (->WindSpeed 10 "MCI")
+                                  (->Temperature 10 "MCI")
+                                  (->Cold 10))
+                          (fire-rules))]
+      (is (empty? (query end-session negation-with-prior-bindings)))
+      (is (no-system-types? end-session)))
 
     ;; There should be only one root to the beta tree because the top condition is reused.
```
mrrodriguez: Why comment this out?

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Ugh, not sure how I missed that.. will fix

src/test/clojure/clara/test_rules.clj
WilliamParker: I was using "r" for "result".  I agree it is weird though, I just haven't thought of a better name and I like using the as-> here.  Any ideas?

src/test/clojure/clara/test_rules.clj
WilliamParker: I don't know, I was just setting it aside for now since it was commented out already.  It seems undesirable to write tests directly against to-beta-tree though, not sure why it was done in the first place.  It can probably just be deleted.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Works for me, will change.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1507,32 +1519,62 @@
...
+        fact-type->roots (memoize
+                          (fn [fact-type]
+                            ;; There is no inherent ordering here but we put the AlphaRootsWrapper instances
+                            ;; in a vector rather than a set to avoid nondeterministic ordering (and thus nondeterministic
+                            ;; performance).
+                            (into []
+                                  ;; If a given type in the ancestors has no matching alpha roots,
+                                  ;; don't return it as an ancestor.  Fact-type->roots is memoized on the fact type,
+                                  ;; but work is performed per group returned on each call the to get-alphas-fn.  Therefore
+                                  ;; removing groups with no alpha nodes here will improve performance on subsequent calls.=
```
WilliamParker: The "=" is just a typo.  Will remove.

src/test/clojure/clara/test_rules.clj
```diff
@@ -1183,75 +1183,85 @@
...
 
-    (is (= [{}]
-           (-> s
-               (fire-rules)
-               (query cold-not-match-temp))))
+        s (mk-session [cold-not-match-temp object-query] :cache false)
+        s-with-prior (mk-session [negation-with-prior-bindings object-query] :cache false)
+        s-with-nested (mk-session [nested-negation-with-prior-bindings object-query] :cache false)
+
+        no-system-types? (fn [session]
```
mrrodriguez: It isn't clear to me why this check has been added to all of these test cases.  It looks like revealing an impl detail unnecessarily in the test, but I'm guessing I'm missing something?

If there is a meaningful reason to do this, a comment above this line may be helpful to make that clear for future reading.

WilliamParker: This is testing the wrapping of the ancestors-fn, namely to make sure that no internal implementation details leak out to the user in the form of NegationResult matching on their productions.  The other way to do this would be to explicitly assert on an enumerated list of all Object instances in the session after the test - maybe we should do that but it doesn't seem ideal either.

mrrodriguez: That makes sense to me that you want to test that our internal objects do not match user-provided rules/queries.  It'd probably be nice to have a comment above this that mentions that?

src/test/clojure/clara/test_rules.clj
mrrodriguez: Ah yes.  I see how `as->` can get awkward since the binding really represents multiple phases in a thread of results.  That's fair.  I was thinking `r` for `session` was weird, but I see how you use it later as the query result.

I don't know what the idiom is for this one.  Maybe just `x` or something that doesn't try to name anything specifically.  It doesn't matter though.  I asked more out of curiosity here.

src/test/clojure/clara/test_rules.clj
mrrodriguez: That's fine.  I was just wondering if it should be deleted instead.

WilliamParker: Looking at this again I can't think of a good reason to test this implementation detail.  I'll remove the commented-out assertions.

src/test/clojure/clara/test_rules.clj
WilliamParker: +1 to a comment, will add

src/test/clojure/clara/test_rules.clj
WilliamParker: "x" works for me, will change.

WilliamParker:
Updated in response to @mrrodriguez 's comments

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1487,52 +1487,99 @@
...
+          ;; where a Java iterator can return the same entry object repeatedly and mutate it after each next() call.  We use mutable lists
+          ;; for performance but wrap them in unmodifiableList to make it clear that the caller is not expected to mutate these lists.
+          ;; Since after this function returns the only reference to the fact lists will be through the unmodifiedList we can depend elsewhere
+          ;; on these lists not changing.  Since the only expected workflow with these lists is to loop through them, not add or remove elements,
+          ;; we don't gain much from using a transient (which can be efficiently converted to a persistent data structure) rather than a mutable type. 
+          (loop []
+            (when (.hasNext entries-it)
+              (let [^java.util.Map$Entry e (.next entries-it)]
+                (.add return-list [(-> e ^AlphaRootsWrapper (.getKey) .roots)
+                                   (java.util.Collections/unmodifiableList (.getValue e))])
```
mrrodriguez: I don't believe this is a reflection warning right?  I think the compiler can resolve the method since it only has one signature of this length.  Just double checking.

mrrodriguez:
+1

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Apparently not:

````
user=> (set! *warn-on-reflection* true)
true
user=> (require 'clara.test-rules)
Reflection warning, clara/test_rules.clj:55:39 - reference to field getCause can't be resolved.
Reflection warning, clara/test_rules.clj:2094:19 - reference to field rulebase can't be resolved.
Reflection warning, clara/test_rules.clj:2101:19 - reference to field rulebase can't be resolved.
Reflection warning, clara/test_rules.clj:3305:31 - reference to field rulebase can't be resolved.
Reflection warning, clara/test_rules.clj:3365:13 - reference to field rulebase can't be resolved.
nil
user=> 
````

It makes sense, as you say the compiler has all the information it needs to resolve this.

PR #243, opened 2016-12-01
Remove metadata before trying to compile alpha nodes

EthanEChristian:
Preserving metadata from the condition on the alpha expression doesn't seem to have much benefit and can lead to some rather odd exceptions if it contains unresolvable symbols, eg. 

the tested add, prior to the removal of the metadata, throws
```
 ERROR in (test-unresolvable-symbols-in-metadata-on-conditions) (core.clj:4593)
Uncaught exception, not in assertion.
expected: nil
  actual: clojure.lang.ExceptionInfo: Failed compiling alpha node
{:expr (clojure.core/fn [?__fact__ {:keys [rule-output]}] (clojure.core/let [this ?__fact__ temperature (.-temperature ?__fact__) ?__bindings__ (clojure.core/atom {})] (if (clojure.core/< temperature 20) (clojure.core/deref ?__bindings__) nil))), :condition {:type clara.rules.testfacts.Temperature, :constraints [(clojure.core/< temperature 20)]}, :env {:rule-output #object[clojure.lang.Atom 0x6cfcd230 {:status :ready, :val nil}]}}

 at clojure.core$ex_info.invoke (core.clj:4593)
    clara.rules.compiler$try_eval.invoke (compiler.clj:225)
    clara.rules.compiler$eval6096$compile_alpha_nodes__6097$fn__6098$iter__6099__6103$fn__6104$fn__6105$fn__6109.invoke (compiler.clj:1485)
    clara.rules.compiler$eval6096$compile_alpha_nodes__6097$fn__6098$iter__6099__6103$fn__6104$fn__6105.invoke (compiler.clj:1481)
    clara.rules.compiler$eval6096$compile_alpha...
...
...
Caused by: clojure.lang.Compiler$CompilerException: java.lang.RuntimeException: Unable to resolve symbol: unresolvable-symbol in this context, compiling:(/private/var/folders/bn/d534tk9x6qj41tvgpfkr1mysfnmmcl/T/form-init2742502976728890894.clj:1:6419)
 at clojure.lang.Compiler.analyze (Compiler.java:6543)
    clojure.lang.Compiler.analyze (Compiler.java:6485)
    clojure.lang.Compiler$MapExpr.parse (Compiler.java:3050)
    clojure.lang.Compiler$FnExpr.parse (Compiler.java:3998)
...
```

@WilliamParker @mrrodriguez

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -3,7 +3,10 @@ Cerner Corporation
 - Ryan Brush [@rbrush]
 - Mike Rodriguez [@mrrodriguez]
 - William Parker [@WilliamParker]
+- Ethan Christian [@EthanEChristian]
 
 [@rbrush]: https://github.com/rbrush
 [@mrrodriguez]: https://github.com/mrrodriguez
 [@WilliamParker]: https://github.com/WilliamParker
+[@EthanEChristian]: https://github.com/EthanEChristian
+
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1472,9 +1472,12 @@
         :let [{:keys [type constraints fact-binding args]} condition
               cmeta (meta condition)
               alpha-expr (with-meta (compile-condition
-                                     type (first args)  constraints
-                                     fact-binding env)
-                           (meta condition))]]
+                                      type (first args)  constraints
+                                      fact-binding env)
+                           ;; Remove all metadata but file and line number
+                           ;; to protect from evaluating usafe metadata
+                           ;; See PR 243 for more detailed discussion
+                           (select-keys cmeta [:line :file]))]]
 
     (with-meta
       (cond-> {:type (effective-type type)
--- a/src/test/clojure/clara/test_rules.clj
+++ b/src/test/clojure/clara/test_rules.clj
@@ -4974,3 +4974,20 @@
           (str "As long as one negation condition in the :or matches, regardless of how many facts match the other "
                "negation the rule should fire for join type " join-type)))))
 
+(deftest test-unresolvable-symbols-in-metadata-on-conditions
+  ;; Test removal of unwanted metadata; see PR 243 for details
+  (let [rule-output (atom nil)
+        ns-nom (ns-name *ns*)
+        rule {:ns-name ns-nom
+              :name (str ns-nom "/unresolvable-metadata-rule")
+              :lhs [(with-meta {:type Temperature
+                                :constraints [(list `< 'temperature 20)]}
+                      {:custom-meta 'unresolvable-symbol})]
+              :rhs `(do (reset! ~'rule-output ~'?__token__))
+              :env {:rule-output rule-output}}
+
+        session (-> (mk-session [rule])
+                    (insert (->Temperature 10 "MCI"))
+                    (fire-rules))]
+
+    (is (has-fact? @rule-output (->Temperature 10 "MCI")))))


WilliamParker:
I think we could remove the metadata from the top-level condition since that just ends up being used by Clara and nothing else; removing metadata from things nested inside the condition would be a functional problem since user code could depend on it, most likely for type hints.

This seems like a problem with the user code that is passing in such conditions though.  This is happening for productions targeting the Clara data model [1] directly, not productions written using the DSL syntax [2, 3], correct?  Making such a manually built condition safe to eval seems like a user responsibility rather than something that Clara needs to defend against.  In any case, as discussed above, I don't think Clara can defend against this class of problems in general.  Could you elaborate on what exactly is causing this?  We can discuss off-thread if the problem inherently involves Cerner IP. 

I'd also be curious if @rbrush had functional reasons for keeping that metadata (I don't see any but could be missing something) or whether it was just for debugging and REPL-based exploration.


1. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules/schema.clj#L82
2. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules.cljc#L334
3. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules.cljc#L365

mrrodriguez:
@WilliamParker Now that I look at it, I think the metadata is being added in case there is information that the compiler can use, like line # and file metadata.  This would be to help report errors.

You can see the difference if you look at the `*e` when compiling
```clj
(eval '(fn [x] y))

;; vs

(eval (with-meta '(fn [x] y) {:line 1}))
```

Other than that, no metadata is preserved after compilation of this form, so it isn't useful.
However, this is a pretty big inconvenience if the caller happens to have metadata attached to this part of the rule construct (programmatically generating the rule).  I wonder if this should just preserve specific metadata prior to compilation instead.

WilliamParker:
Something that strikes me about this is that I don't see this metadata problem as limited to this particular case.  What about rule RHS code for example?  It seems like if we want some sort of normalization like this that it should occur as soon as possible, not be scattered through the compiler at the point where eval is called.  Thoughts?

rbrush:
My original intention was to keep file and line metadata for debugging purposes, as @mrrodriguez suggests. We should try to preserve that if we can. 

I don't have a great understanding of how unresolvable symbols would show up in our metadata in a real scenario, but if we need to selectively remove them that seems reasonable.

EthanEChristian:
Without going into too much detail, this is a product of generating rules where the conditions used still have metadata attached.

If we are only concerned with maintaining `:line` and `:file`, that should be a pretty simple change.

WilliamParker:
To be honest I still think the best place to deal with this is in user code, but it isn't a big deal; Clara's external API isn't impacted at all and it is just a tiny bit of extra complexity in the code.  If others want to take care of this in Clara itself I'll stop objecting to that. :)

Could we add some comments, both here and in the test, about why this is being done, in particular that this is removing metadata that originally came from the user though?  I think why this is being done only for alpha node expressions, rather than in other places in the compiler (namely beta expression compilation), and in the case of the test why we are using unresolvable symbols, would be very confusing to someone reading through that code without the context of this PR.

mrrodriguez:
>To be honest I still think the best place to deal with this is in user code, but it isn't a big deal; Clara's external API isn't impacted at all and it is just a tiny bit of extra complexity in the code

The issue here is that Clara is trying to call `eval` on metadata from a structure that is larger than the actual form being compiled.  So I think this is an, in general, dangerous thing to do.  It is calling `eval` on the metadata of the entire Condition structure,  not the metadata involved in the forms actually being evaluated.  

It only broke with an external generation of the rules right now, that is true.  However, in general I see this as just `eval`ing more details than need to be and can be something subtle we could break later if using other sorts of metadata on the Condition constructs.

Also, an "API" that doesn't break with "extra/unknown" keys in maps, including metadata, is going to be a more robust one in general.  It'd be odd that the caller has to drop their metadata off Condition maps they generate to be defensive of Clara breaking on it.

WilliamParker:
"It is calling eval on the metadata of the entire Condition structure, not the metadata involved in the forms actually being evaluated." => Fair point.  As a user I'd probably expect Clara to use the metadata on the actual forms used to generate the alpha filter function, not the Condition, but the needed metadata is on the Condition so we use that metadata to improve our error reporting.  Looking briefly through the beta network compilation steps it looks like metadata is just lost there currently, so I don't think this change will introduce a new consistency problem between the alpha and beta networks.

In short, +1 to the functional change now, although I do think we need some improved documentation in the code (i.e. comments).

EthanEChristian:
Added a small comment to the location where i removed the metadata

mrrodriguez:
+1

WilliamParker:
Could we reference the pull request number in that comment in compiler.clj and in the test?Something like "See PR 243 for more detailed discussion" and "Test removal of unwanted metadata; see PR 243 for details"?  I realize that the Git history will keep the information but later changes can make that difficult to track down.  Otherwise +1.

EthanEChristian:
made the recommended changes from @WilliamParker's comment above

Issue #239, opened 2016-11-27
Custom fact-type-fn causes degradation of performance in rulesets with many facts


WilliamParker:
When we have a custom fact-type-fn [1] we have to make an isa? call that uses Clojure's custom hierarchies for every fact that is inserted [2].  This turns out to have some pretty significant impacts on performance.  Distinguishing between system fact types and user-provided fact types is necessary for functional correctness since a user-provided function might otherwise disrupt the functioning of internally created rules.  Currently our only use case for internally created rules is the fix for https://github.com/cerner/clara-rules/issues/149.  We may also use them to solve https://github.com/cerner/clara-rules/issues/218.  However, we'll need to find something more performant.  Using the JVMs type system directly rather than Clojure's hierarchies is a likely candidate.

1. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules.cljc#L294
2. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules/compiler.clj#L1543

WilliamParker:
This was resolved by https://github.com/cerner/clara-rules/pull/244 ; thanks, @EthanCChristian

Closing.

Issue #241, opened 2016-11-28
The get-alphas-fn in ClojureScript should wrap the fact-type-fn and ancestors-fn


WilliamParker:
The fact-type-fn is currently wrapped to detect system types in the Clojure implementation [1] but no such wrapping exists in the ClojureScript implementation [2, 3].  This could cause the fix for https://github.com/cerner/clara-rules/issues/149 to break in ClojureScript if a custom fact-type-fn is used.  The ClojureScript implementation should also use the wrapped version on the ancestors-fn that I anticipate will be added to resolve https://github.com/cerner/clara-rules/issues/236 for reasons discussed there that are the same in Clojure and ClojureScript.

1. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules/compiler.clj#L1530
2. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules.cljc#L129
3. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules.cljc#L229

Zylox:
I think link one is supposed to be https://github.com/cerner/clara-rules/blob/master/src/main/clojure/clara/rules/compiler.clj#L1515

WilliamParker:
@Zylox agreed; I've updated the original post with a link to the same tag as previously.

WilliamParker:
I've started working on this.

WilliamParker:
I have submitted a PR to fix this at https://github.com/cerner/clara-rules/pull/295

WilliamParker:
https://github.com/cerner/clara-rules/pull/295 fixed this issue and has been merged.  Closing.

Issue #186, opened 2016-05-02
ClojureScript tests can fail without the main build failing


WilliamParker:
As discussed at https://github.com/rbrush/clara-rules/pull/185 in at least one scenario it is possible for the ClojureScript tests to not pass without causing the main Leiningen build to fail.  The objective of this issue is to find and implement a strategy that eliminates such "silent failure".

WilliamParker:
Pull request at https://github.com/cerner/clara-rules/pull/336

WilliamParker:
The PR has been merged and [a build of a branch with a ClojureScript-only test failure failed.](https://travis-ci.org/WilliamParker/clara-rules/builds/265704495?utm_source=github_status&utm_medium=notification).

Closing.

PR #244, opened 2016-12-01
Issue 239: Remove clojure hierarchy checks on internal facts, replace…

EthanEChristian:
… with a marker interface

Issue #239: 

It could be done faster, if instead of the interface we went with a straight
`(indentical? (class fact) NegationResult)`
but this would be less extensible for future system facts.

Running a couple quick bench marks:

Setup
```
(require '[criterium.core :as crit])

(defprotocol ISystemFact)

(defmacro gen-record-classes
  [num-record-classes]
  (let [gen-record-symbol #(symbol (str "TestRecord" %))
        gen-record-form #(if (= 0 (mod % 50)) 
                           (list `defrecord (gen-record-symbol %) [] `ISystemFact)
                           (list `defrecord (gen-record-symbol %) []))
        gen-derived-form #(list `derive (gen-record-symbol %) ::system-type-fact)
        derived-forms (for [x (range (Math/floor (/ num-record-classes 50)))]
                        (gen-derived-form (* 50 x)))]
    `(do ~@(mapv gen-record-form (range num-record-classes))
         ~@derived-forms)))

;; TestRecords 0-99
;; TestRecord 0 and 50 ::system-type-fact
;; TestRecord 0 and 50 ISystemFact
(gen-record-classes 100)


(let [rec-symbols (mapv #(resolve (symbol (str "->TestRecord" %))) (range 100))]
          (def test-records (doall
                              (shuffle 
                                (into []
                                  (for [_ (range (/ 1E5 100))
                                        y rec-symbols]
                                    (y)))))))
```

Benchmarks
```
(crit/with-progress-reporting 
  (crit/bench (doseq [x test-records]
                (isa? (type x) ::system-type-fact))))


Evaluation count : 60 in 60 samples of 1 calls.
             Execution time mean : 1.350904 sec
    Execution time std-deviation : 37.959478 ms
   Execution time lower quantile : 1.322498 sec ( 2.5%)
   Execution time upper quantile : 1.458715 sec (97.5%)
                   Overhead used : 2.016640 ns

Found 11 outliers in 60 samples (18.3333 %)
	low-severe	 5 (8.3333 %)
	low-mild	 6 (10.0000 %)


(crit/with-progress-reporting 
  (crit/bench (doseq [x test-records]
                (instance? ISystemFact x))))

Evaluation count : 286860 in 60 samples of 4781 calls.
             Execution time mean : 207.726407 µs
    Execution time std-deviation : 4.254857 µs
   Execution time lower quantile : 203.674234 µs ( 2.5%)
   Execution time upper quantile : 213.449557 µs (97.5%)
                   Overhead used : 2.016640 ns

Found 3 outliers in 60 samples (5.0000 %)
	low-severe	 2 (3.3333 %)
	low-mild	 1 (1.6667 %)
 Variance from outliers : 9.3793 % Variance is slightly inflated by outliers
```

The instance check does quite a bit better.

@WilliamParker @mrrodriguez

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -28,7 +28,8 @@
             AccumulateWithJoinFilterNode
             LocalTransport
             LocalSession
-            Accumulator]
+            Accumulator
+            ISystemFact]
            [java.beans
             PropertyDescriptor]))
 
@@ -1508,14 +1509,14 @@
   returns a map associating alpha nodes with the facts they accept."
   [fact-type-fn ancestors-fn alpha-roots]
 
-  (let [ ;; If a customized fact-type-fn is provided,
+  (let [;; If a customized fact-type-fn is provided,
         ;; we must use a specialized grouping function
         ;; that handles internal control types that may not
         ;; follow the provided type function.
         wrapped-fact-type-fn (if (= fact-type-fn type)
                                type
                                (fn [fact]
-                                 (if (isa? (type fact) :clara.rules.engine/system-type)
+                                 (if (instance? ISystemFact fact)
                                    ;; Internal system types always use Clojure's type mechanism.
                                    (type fact)
                                    ;; All other types defer to the provided function.
@@ -1524,7 +1525,7 @@
         ;; Wrap the ancestors-fn so that we don't send internal facts such as NegationResult
         ;; to user-provided productions.  Note that this work is memoized inside fact-type->roots.
         wrapped-ancestors-fn (fn [fact-type]
-                               (if (isa? fact-type :clara.rules.engine/system-type)
+                               (if (isa? fact-type ISystemFact)
                                  ;; Exclude system types from having ancestors for now
                                  ;; since none of our use-cases require them.  If this changes
                                  ;; we may need to define a custom hierarchy for them.
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -30,11 +30,30 @@
 (def empty-token (->Token [] {}))
 
 ;; Record indicating the negation existing in the working memory.
-(defrecord NegationResult [gen-rule-name])
+;;
+;; Determining if an object is an instance of a class is a primitive
+;; JVM operation and is much more efficient than determining
+;; if that object descends from a particular object through
+;; Clojure's hierarchy as determined by the isa? function.
+;; See Issue 239 for more details.
+#?(:clj
+    (do
+      ;; A marker interface to identify internal facts.
+      (definterface ISystemFact)
+      (defrecord NegationResult [gen-rule-name]
+        ISystemFact))
 
-;; Make the negation result a "system type", so its type is not overridden
-;; with a customized fact type function.
-(derive NegationResult :clara.rules.engine/system-type)
+    :cljs
+    (do
+      (defrecord NegationResult [gen-rule-name])
+      ;; Make NegationResult a "system type" so that NegationResult
+      ;; facts are special-cased when matching productions. This serves
+      ;; the same purpose as implementing the ISystemFact Java interface
+      ;; on the Clojure version of NegationResult.
+      ;; ClojureScript does not have definterface; if we experience performance
+      ;; problems in ClojureScript similar to those on the JVM that are
+      ;; described in issue 239 we can investigate a similar strategy in JavaScript.
+      (derive NegationResult ::system-type)))
 
 ;; Schema for the structure returned by the components
 ;; function on the session protocol.
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -3,7 +3,10 @@ Cerner Corporation
 - Ryan Brush [@rbrush]
 - Mike Rodriguez [@mrrodriguez]
 - William Parker [@WilliamParker]
+- Ethan Christian [@EthanEChristian]
 
 [@rbrush]: https://github.com/rbrush
 [@mrrodriguez]: https://github.com/mrrodriguez
 [@WilliamParker]: https://github.com/WilliamParker
+[@EthanEChristian]: https://github.com/EthanEChristian
+
--- a/src/test/clojure/clara/test_rules.clj
+++ b/src/test/clojure/clara/test_rules.clj
@@ -17,6 +17,8 @@
             [schema.test])
   (import [clara.rules.testfacts Temperature WindSpeed Cold Hot TemperatureHistory
            ColdAndWindy LousyWeather First Second Third Fourth FlexibleFields]
+          [clara.rules.engine
+           ISystemFact]
           [java.util TimeZone]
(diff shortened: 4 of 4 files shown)


src/main/clojure/clara/rules/engine.cljc
```diff
@@ -29,12 +29,12 @@
 ;; Token with no bindings, used as the root of beta nodes.
 (def empty-token (->Token [] {}))
 
-;; Record indicating the negation existing in the working memory.
-(defrecord NegationResult [gen-rule-name])
+;; A marker interface to identify internal facts.
+(definterface ISystemFact)
```
WilliamParker: Definterface doesn't exist in ClojureScript at this time [1, 2].  This makes sense since it is fundamentally about a JVM construct that doesn't apply to a JS VM.  This causes the ClojureScript compilation to fail as can be seen in the Travis log for this job.  [3]  Unfortunately this doesn't cause the main build to fail as discussed at https://github.com/cerner/clara-rules/issues/186.

My suggestion would be to make this a protocol instead, which appears to exist in ClojureScript based on usage in cljs.core.  On the JVM that will create a concrete interface that we can use with instance?; in ClojureScript when we address https://github.com/cerner/clara-rules/issues/241 we can use satisfies?.  We could continue to define an interface, derive the type using a custom hierarchy for ClojureScript, and work it out with reader conditionals between Clojure and ClojureScript but I'd prefer to avoid that if we can.

1. http://dev.clojure.org/jira/browse/CLJS-1190
2. https://github.com/clojure/clojurescript/search?utf8=%E2%9C%93&q=definterface
3. https://travis-ci.org/cerner/clara-rules/builds/180306623

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1524,7 +1525,7 @@
         ;; Wrap the ancestors-fn so that we don't send internal facts such as NegationResult
         ;; to user-provided productions.  Note that this work is memoized inside fact-type->roots.
         wrapped-ancestors-fn (fn [fact-type]
-                               (if (isa? fact-type :clara.rules.engine/system-type)
+                               (if (isa? fact-type ISystemFact)
```
WilliamParker: Why did you use isa? here instead of instance?  This work ends up being memoized so it isn't as big of a deal as the fact-type-fn but I don't see a reason not to just use instance? since you're moving the JVM class dispatch here anyway.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: Using `defprotocol` for its interface is technically using an implementation detail on the JVM.  It isn't their intended purpose.  They just happen to make an interface for efficiency in some cases.

Reader conditionals seems reasonable.  Just use `isa?` in cljs perhaps.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Right, nevermind.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: True, but realistically I don't expect that to change and tests would fail immediately if it did when we failed to match on NegationResult facts in extracted rules.  To be honest I see that as one of those areas of Clojure where it isn't an "official" API but in practice it is; maybe there are contrary conversations on that that I haven't seen though.  Reader conditionals would work though albeit a bit awkwardly; I'm fine with that approach.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -30,11 +30,19 @@
 (def empty-token (->Token [] {}))
 
 ;; Record indicating the negation existing in the working memory.
-(defrecord NegationResult [gen-rule-name])
```
WilliamParker: Can you add some comment(s) explaining why we have this divergence between Clojure and ClojureScript here?  This could be pretty puzzling for someone looking at this later without some context.  In particular I suggest mentioning this Clara issue number i.e. https://github.com/cerner/clara-rules/issues/239

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1508,14 +1509,18 @@
   returns a map associating alpha nodes with the facts they accept."
   [fact-type-fn ancestors-fn alpha-roots]
 
-  (let [ ;; If a customized fact-type-fn is provided,
+  (let [internal-fact? (if (compiling-cljs?)
+                         #(isa? (type %) ::eng/system-type)
+                         #(instance? ISystemFact %))
```
WilliamParker: You don't actually need this.  clara.rules.compiler/create-get-alphas-fn returns a (JVM) Clojure function object; this can't be called inside a JavaScript VM.  Clara's ClojureScript version uses another implementation of the get-alphas-fn at [1, 2].  The general principle underlying this is that ClojureScript code can use Clojure macros that return valid ClojureScript code since the ClojureScript compiler runs on the JVM.  In Clara's case, we have the defsession macro [3] that is referred to by ClojureScript code (in this case tests, but it would be done the same way in production code). [4]   However, this is only possible for raw ClojureScript code; other JVM objects can't be used.  This is why we have a significant amount of compilation code that is different between Clojure and ClojureScript.

The ClojureScript version of the get-alphas-fn doesn't have the wrapping of the get-fact-type-fn and ancestors-fn that it should; I logged https://github.com/cerner/clara-rules/issues/241 for that.

1. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules.cljc#L129
2. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules.cljc#L229
3. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules.cljc#L313
4. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/test/common/clara/test_common.cljc#L8

WilliamParker:
I think we should update the tests at https://github.com/cerner/clara-rules/blob/abc11037ad6439e96c8e77d81675835054cc7234/src/test/clojure/clara/test_rules.clj#L11with something like

````
no-system-types? (fn [session]
                           (not-any? (fn [fact] (or (isa? (type fact) ::eng/system-type))
                                                                  (instance? ISystemFact fact))
                                     (as-> session x
                                       (query x object-query)
                                       (map :?o x))))]
````

File: src/main/clojure/clara/rules/engine.cljc (hunk 3 above)
Comment (EthanEChristian) [inline src/main/clojure/clara/rules/engine.cljc]: updated

**Comment by EthanEChristian** [comment]
I have updated that [test] (https://github.com/cerner/clara-rules/blob/abc11037ad6439e96c8e77d81675835054cc7234/src/test/clojure/clara/test_rules.clj#L1165 ) to 

```
no-system-types? (fn [session]
                           (not-any? (fn [fact] (instance? ISystemFact fact))
                                     (as-> session x
                                       (query x object-query)
                                       (map :?o x))))] ```

as the :eng/system-type is specific to the cljs implementation now

PR #251, opened 2016-12-15
Clara durability should prefer ThreadLocal to dynamic vars

kulkarnipushkar:
The use of the dyanamic vars in clara.rules.durability and clara.rules.durability.fressian namsespace are wasting unnecessary cpu cycles. A ThreadLocal does a subset of the work done by a Clojure dynamic var and we do not really need any of the more sophisticated properties of dynamic vars for our purposes here. The proposal is to change these to ThreadLocal.

With this change we could see the performance gain, please look at the below test result.
```
> (defrecord MyRecord1 [n]) 
> (defrecord MyRecord2 [n]) 
> (defrecord MyRecord3 [n]) 
> (defrecord MyRecord4 [n]) 
> (defrecord MyRecord5 [n]) 

> (def records (into [] 
cat 
(for [n (range 1e6)] 
[(MyRecord1. n) (MyRecord2. n) (MyRecord3. n) (MyRecord4. n) (MyRecord5. n)]))) 
> (time (do (serde records) nil)) 

BEFORE: "Elapsed time: 27723.323316 msecs"
AFTER: "Elapsed time: 24548.738085 msecs"
```
We could also see noticible performance gain in our production workflows.

### DIFF
--- a/src/main/clojure/clara/rules/platform.cljc
+++ b/src/main/clojure/clara/rules/platform.cljc
@@ -56,3 +56,23 @@
           (persistent!)))
     :cljs
     (def tuned-group-by clojure.core/group-by))
+
+#?(:clj
+    (defmacro thread-local-binding
+      "Wraps given body in a try block, where it sets each given ThreadLocal binding 
+       and removes it in finally block."
+      [bindings & body]
+      (when-not (vector? bindings)
+        (throw (ex-info "Binding needs to be a vector."
+                        {:bindings bindings})))
+      (when-not (even? (count bindings))
+        (throw (ex-info "Needs an even number of forms in binding vector"
+                        {:bindings bindings})))
+      (let [binding-pairs (partition 2 bindings)]
+        `(try
+           ~@(for [[tl v] binding-pairs]
+               `(.set ~tl ~v))
+           ~@body
+           (finally
+             ~@(for [[tl] binding-pairs]
+                 `(.remove ~tl)))))))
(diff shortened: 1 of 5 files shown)


src/main/clojure/clara/rules/platform.cljc
```diff
@@ -56,3 +56,22 @@
           (persistent!)))
     :cljs
     (def tuned-group-by clojure.core/group-by))
+
```
kulkarnipushkar: moved thread-local-binding macro inside clj-only reader conditional

WilliamParker:
@kulkarnipushkar please squash this PR into one commit.

WilliamParker:
Elaborating a bit on the motivation for these changes, looking up a Clojure dynamic var involves more work than looking up a single JVM ThreadLocal.  The implementation can be found at [1].  Among other things, there is a map of the bindings for the thread that the particular dynamic var involved has to be looked up in.  This more dynamic approach allows Clojure to duplicate bindings between threads when a Clojure construct retrieves a new thread from Clojure's thread pool; the heart of this is [2] and its usages.  In this case we don't need these properties since our durability implementation doesn't create new threads.  Note that user code can be multithreaded; this doesn't impact that.

1. https://github.com/clojure/clojure/blob/clojure-1.8.0/src/jvm/clojure/lang/Var.java#L194
2. https://github.com/clojure/clojure/blob/clojure-1.8.0/src/clj/clojure/core.clj#L1930

PR #252, opened 2016-12-16
Issue 250: Do not remove both a pending activation and the logical in…

WilliamParker:
…sertions from a previous activation based on the left-retract of a single token from a ProductionNode

### DIFF
--- a/src/main/clojure/clara/rules/memory.cljc
+++ b/src/main/clojure/clara/rules/memory.cljc
@@ -95,7 +95,11 @@
   ;; Returns the group of the next activation, or nil if none are pending.
   (next-activation-group [memory])
 
-  ;; Remove the given activations from the working memory.
+  ;; Remove the given activations from the working memory.  This is expected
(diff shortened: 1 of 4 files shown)


src/main/clojure/clara/rules/memory.cljc
```diff
@@ -803,35 +807,55 @@
...
              (let [not-removed (loop [to-remove-item (first to-remove)
                                       to-remove (next to-remove)
                                       not-removed (transient [])]
                                  (if to-remove-item
-                                   (let [act (->rule-ordered-activation to-remove-item true)]
+                                   (let [^RuleOrderedActivation act (->rule-ordered-activation to-remove-item true)]
                                      (if (.remove activations act)
-                                       (recur (first to-remove) (next to-remove) not-removed)
+                                       (do
+                                         (.add removed-activations (.activation ^RuleOrderedActivation act))
```
mrrodriguez: nit: you shouldn't need this type hint since `act` is already hinted above.

mrrodriguez:
+1 except for my minor comment.

PR #254, opened 2016-12-20
Issue 253: Synchronize the cache of classes to factory function symbols

WilliamParker:


mrrodriguez:
+1 works for now.  Not the most optimal solution for a multithreaded workflow, but we can revisit it later.

WilliamParker:
Yeah, we can look at this later.  I'd also like to consider making separate handlers for different record classes, which would avoid the workflows of looking up stuff for a record entirely.  Right now I'm just looking for something simple and possibly temporary to remove the defect.

Issue #255, opened 2017-01-07
Provide better feedback when constraints throw exceptions


carlosphillips:
It is currently difficult to identify which constraint threw an exception during alpha node activation and retraction and during the activation and retraction of expression join nodes. This is important because even the most innocuous constraint can throw an exception. For example, the constraint (> ?rate 1) throws in the following code because rate is nil in the provided fact.

```Clojure
(ns test
  (:require [clara.rules :as rules]
            [clara.rules.dsl :as dsl]))

(defrecord RateToUSD [currency rate])

(let [rule-template (dsl/parse-rule [[RateToUSD (= ?rate rate)
                                      (= ?currency currency)
                                      (> ?rate 1)]]
                                    (println "Ok"))
      rule (assoc rule-template :name "find-all-rates")]
  (-> (rules/mk-session [rule] :cache false)
      (rules/insert (->RateToUSD "CAD" nil))
      (rules/fire-rules)))

```

A NullPointerException is thrown by clojure.lang.Numbers.ops with the following stack trace.

```
NullPointerException
	clojure.lang.Numbers.ops (Numbers.java:1013)
	clojure.lang.Numbers.gt (Numbers.java:229)
	clojure.lang.Numbers.gt (Numbers.java:3864)
	test/eval24511/fn--24512/fn--24514 (form-init2109496090701568388.clj:2)
	test/eval24511/fn--24512 (form-init2109496090701568388.clj:2)
	clara.rules.engine/alpha-node-matches/iter--21928--21932/fn--21933 (engine.cljc:440)
	clojure.lang.LazySeq.sval (LazySeq.java:40)
	clojure.lang.LazySeq.seq (LazySeq.java:49)
	clojure.lang.RT.seq (RT.java:507)
	clojure.core/seq--4128 (core.clj:137)
	clara.rules.engine.AlphaNode/iter--21960--21964/fn--21965 (engine.cljc:459)
	clojure.lang.LazySeq.sval (LazySeq.java:40)
```
Notice how no information is provided as to which constraint threw the exception, which fact was being processed, or what the bindings were when the constraint threw the exception. This makes it difficult to debug constraints that throw exceptions. Better exception messages should help the development process. For example, printing the following message could help in addition to the stack trace.

```
Exception raised during execution of constraint
  (> ?rate 1)
of condition
  [test.RateToUSD (= ?rate rate) (= ?currency currency) (> ?rate 1)]
of rule
  find-all-rates
when processing fact
[...]

Issue #23, opened 2013-10-31
Add no-loop option for rules


sitch:
I get some strange behavior when I have circular types of rules:

``` clojure
(defrule optimist
  [?g <- Glass (= ?full full) (= ?capacity capacity)]
  [Faucet (= ?flow flow)]
  [:test (not= ?full ?capacity)]
  =>
  (retract! (->Glass ?full ?capacity))
  (insert! (->Glass (+ ?full ?flow) capacity))
  (println ?g))

```

I was hoping this would work something like:

```
{:capacity 100 :full 0}
{:capacity 100 :full 25}
{:capacity 100 :full 50}
{:capacity 100 :full 100}
```

I was wonder what your thoughts are on handling a rule like this

Issue #19, opened 2013-10-06
Another way to improve effeciency.


ylando2:
Do not fire the rules on all the production nodes. Keep a list of node that has unfired tokens and fire them.

PR #256, opened 2017-01-07
Issue 255: Provide better feedback when constraints throw exceptions

carlosphillips:
Please see issue #255.

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -235,7 +235,11 @@
   "Compiles a sequence of constraints into a structure that can be evaluated.
 
    Callers may also pass a collection of equality-only-variables, which instructs
-   this function to only do an equality check on them rather than create a unification binding."
+   this function to only do an equality check on them rather than create a unification binding.
+
+   The returned expression evaluates to a two item vector, [bindings error], which contains
+   the evaluated bindings in the first argument and a map with information on any exception
+   that might have been thrown by evaluating a constraint as the second vector entry."
   ([exp-seq]
    (compile-constraints exp-seq #{}))
   ([exp-seq equality-only-variables]
@@ -289,7 +293,18 @@
 
          ;; No variables to unify, so simply check the expression and
          ;; move on to the rest.
-         `(if ~exp ~compiled-rest nil))))))
+         (let [exception-type (if (compiling-cljs?) :default Exception)]
+           `(let [test# (try ~exp
+                                 (catch ~exception-type ~'e
+                                   (throw (ex-info "Constraint exception"
+                                                   {:exception ~'e
+                                                    :constraint-exp '~exp
+                                                    :bindings @~'?__bindings__
+                                                    :env ~'?__env__
+                                                    ::constraint-exception true}))))]
+              (if test#
+                ~compiled-rest))))))))
+
 
 
 (defn flatten-expression
@@ -355,7 +370,7 @@
 
         ;; The destructured environment, if any
         destructured-env (if (> (count env) 0)
-                           {:keys (mapv #(symbol (name %)) (keys env))}
+                           {:keys (mapv #(symbol (name %)) (keys env)) :as '?__env__}
                            '?__env__)
 
         ;; Initial bindings used in the return of the compiled condition expresion.
@@ -410,7 +425,7 @@
         binding-keys (variables-as-keywords constraints)
 
         destructured-env (if (> (count env) 0)
-                           {:keys (mapv #(symbol (name %)) (keys env))}
+                           {:keys (mapv #(symbol (name %)) (keys env)) :as '?__env__}
                            '?__env__)
 
         destructured-fact (first args)
@@ -437,7 +452,7 @@
          ~(add-meta '?__fact__ type)
           ~destructured-env]
        (let [~@assignments
-             ~'?__bindings__ (atom {})]
+             ~'?__bindings__ (atom (:bindings ~'?__token__))]
          ~(compile-constraints constraints equality-only-variables)))))
 
 (defn- expr-type [expression]
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -163,6 +163,36 @@
   (alpha-activate [node facts memory transport listener])
   (alpha-retract [node facts memory transport listener]))
 
+
+;; Protocol for getting the type (e.g. :production and :query) and name of a
+;; terminal node.
+(defprotocol ITerminalNode
+  (terminal-node-type [this]))
+
+;; Protocol for getting a node's condition expression.
+(defprotocol IConditionNode
+  (get-condition-description [this]))
+
+(defn get-terminal-node-types
+  [node]
+  (->> node
+       (tree-seq (comp seq :children) :children)
+       (keep #(when (satisfies? ITerminalNode %)
+                (terminal-node-type %)))
+       (into (sorted-set))))
+
+(defn get-conditions-and-rule-names
+  "Returns a map from conditions to sets of rules."
+  ([node]
+   (if-let [condition (when (satisfies? IConditionNode node)
+                           (get-condition-description node))]
+     {condition (get-terminal-node-types node)}
+     (->> node
+         :children
+         (map get-conditions-and-rule-names)
+         (reduce (partial merge-with conj) {})))))
+
+
 ;; Record indicating pending insertion or removal of a sequence of facts.
 (defrecord PendingUpdate [type facts])
 
@@ -373,7 +403,10 @@
 
   (get-join-keys [node] [])
 
-  (description [node] "ProductionNode"))
+  (description [node] "ProductionNode")
+
+  ITerminalNode
+  (terminal-node-type [this] [:production (:name production)]))
 
 ;; The QueryNode is a terminal node that stores the
 ;; state that can be queried by a rule user.
@@ -389,7 +422,96 @@
 
   (get-join-keys [node] param-keys)
 
-  (description [node] (str "QueryNode -- " query)))
+  (description [node] (str "QueryNode -- " query))
+
+  ITerminalNode
+  (terminal-node-type [this] [:query (:name query)]))
+
+
+(defn node-rule-names
+  [child-type node]
+  (->> node
+       (tree-seq (comp seq :children) :children)
+       (keep child-type)
+       (map :name)
+       (distinct)
+       (sort)))
+
+(defn- list-of-names
+  "Returns formatted string with correctly pluralized header and
+  list of names. Returns nil if no such node is found."
+  [singular plural prefix names]
+  (let [msg-for-unnamed (str "  An unnamed " singular ", provide names to your "
+                             plural " if you want them to be identified here.")
+        names-string (->> names
+                          (sort)
+                          (map #(if (nil? %) msg-for-unnamed %))
+                          (map #(str prefix "  " %))
+                          (string/join "\n"))]
+    (if (pos? (count names))
+      (str prefix plural ":\n" names-string "\n"))))
+
+
+(defn- single-condition-message
+  [condition-number [condition-definition terminals]]
+  (let [productions (->> terminals
+                         (filter (comp #{:production} first))
+                         (map second))
+        queries (->> terminals
+                     (filter (comp #{:query} first))
+                     (map second))
+        production-section (list-of-names "rule" "rules" "   " productions)
+        query-section (list-of-names "query" "queries" "   " queries)]
+    (string/join
+     [(str (inc condition-number) ". " condition-definition "\n")
+      production-section
+      query-section])))
+
+(defn- throw-constraint-exception
+  "Adds a useful error message when executing a constraint node raises an exception."
+  [error node fact env]
+  (let [exp (:constraint-exp error)
+        bindings (:bindings error)
+        bindings-description (if (empty? bindings)
+                               "with no bindings\n"
+                               (str "with bindings\n  " bindings))
+        message-header (string/join ["Constraint exception.\n"
+                                     "Exception raised during execution of constraint\n"
+                                     (str "  " exp "\n")
+                                     "when processing fact\n"
+                                     (str "  " (pr-str fact) "\n")
+                                     (str bindings-description "\n")
+                                     "Conditions:\n"])
+        conditions-and-rules (get-conditions-and-rule-names node)
+        condition-messages (->> conditions-and-rules
+                                (map-indexed single-condition-message)
+                                (string/join "\n"))
+        cause (:exception error)
+        message (str message-header "\n" condition-messages)
+        info (-> error
+                 (dissoc :exception)
+                 (merge {:conditions-and-rules conditions-and-rules
+                         :fact fact
+                         :node node}))]
+    (throw (ex-info message info cause))))
+
+(defn constraint-exception-data
+  "Returns exception data for constraint exceptions and nil otherwise."
+  [exception]
+  (let [data (ex-data exception)]
+    (if (:clara.rules.compiler/constraint-exception data)
+      (dissoc data :clara.rules.compiler/constraint-exception))))
+
+(defn- alpha-node-matches
+  [facts env activation node]
+  (for [fact facts
+        :let [bindings (try (activation fact env)
+                            (catch #?(:clj Exception :cljs :default) e
+                              (if-let [error (constraint-exception-data e)]
+                                (throw-constraint-exception error node fact env)
+                                (throw e))))]
+        :when bindings] ; FIXME: add env.
+    [fact bindings]))
 
 ;; Record representing alpha nodes in the Rete network,
 ;; each of which evaluates a single condition and
@@ -397,9 +519,8 @@
 (defrecord AlphaNode [env children activation]
   IAlphaActivate
   (alpha-activate [node facts memory transport listener]
-    (let [fact-binding-pairs (for [fact facts
-                                   :let [bindings (activation fact env)] :when bindings] ; FIXME: add env.
-                               [fact bindings])]
+    node
+    (let [fact-binding-pairs (alpha-node-matches facts env activation node)]
       (l/alpha-activate! listener node (map first fact-binding-pairs))
       (send-elements
         transport
@@ -410,9 +531,7 @@
           (->Element fact bindings)))))
 
   (alpha-retract [node facts memory transport listener]
-    (let [fact-binding-pairs (for [fact facts
-                                   :let [bindings (activation fact env)] :when bindings] ; FIXME: add env.
-                               [fact bindings])]
+    (let [fact-binding-pairs (alpha-node-matches facts env activation node)]
       (l/alpha-retract! listener node (map first fact-binding-pairs))
       (retract-elements
         transport
@@ -466,7 +585,12 @@
      listener
      children
      (for [{:keys [fact bindings] :as element} (mem/remove-elements! memory node join-bindings elements)]
-       (->Token [[fact (:id node)]] bindings)))))
+       (->Token [[fact (:id node)]] bindings))))
+
+  IConditionNode
+  (get-condition-description [this]
+    (let [{:keys [type constraints]} condition]
+      (into [type] constraints))))
 
 ;; Record for the join node, a type of beta node in the rete network. This node performs joins
 ;; between left and right activations, creating new tokens when joins match and sending them to
@@ -527,8 +651,21 @@
      children
      (for [{:keys [fact bindings] :as element} (mem/remove-elements! memory node join-bindings elements)
            token (mem/get-tokens memory node join-bindings)]
-       (->Token (conj (:matches token) [fact id]) (conj (:bindings token) bindings))))))
+       (->Token (conj (:matches token) [fact id]) (conj (:bindings token) bindings)))))
 
+  IConditionNode
+  (get-condition-description [this]
+    (let [{:keys [type constraints]} condition]
+      (into [type] constraints))))
+
+(defn- join-node-matches
+  [node join-filter-fn token fact env]
+  (let [beta-bindings (try (join-filter-fn token fact {})
+                           (catch #?(:clj Exception :cljs :default) e
+                               (if-let [error (constraint-exception-data e)]
+                                 (throw-constraint-exception error node fact env)
+                                 (throw e))))]
+    beta-bindings))
 
 (defrecord ExpressionJoinNode [id condition join-filter-fn children binding-keys]
   ILeftActivate
@@ -545,7 +682,7 @@
            token tokens
            :let [fact (:fact element)
                  fact-binding (:bindings element)
-                 beta-bindings (join-filter-fn token fact {})]
+                 beta-bindings (join-node-matches node join-filter-fn token fact {})]
            :when beta-bindings]
        (->Token (conj (:matches token) [fact id])
                 (conj fact-binding (:bindings token) beta-bindings)))))
@@ -561,7 +698,7 @@
            element (mem/get-elements memory node join-bindings)
            :let [fact (:fact element)
                  fact-bindings (:bindings element)
-                 beta-bindings (join-filter-fn token fact {})]
+                 beta-bindings (join-node-matches node join-filter-fn token fact {})]
            :when beta-bindings]
        (->Token (conj (:matches token) [fact id])
                 (conj fact-bindings (:bindings token) beta-bindings)))))
@@ -581,7 +718,7 @@
      children
      (for [token (mem/get-tokens memory node join-bindings)
            {:keys [fact bindings] :as element} elements
-           :let [beta-bindings (join-filter-fn token fact {})]
+           :let [beta-bindings (join-node-matches node join-filter-fn token fact {})]
            :when beta-bindings]
        (->Token (conj (:matches token) [fact id])
                 (conj (:bindings token) bindings beta-bindings)))))
@@ -595,10 +732,18 @@
      children
      (for [{:keys [fact bindings] :as element} (mem/remove-elements! memory node join-bindings elements)
            token (mem/get-tokens memory node join-bindings)
-           :let [beta-bindings (join-filter-fn token fact {})]
+           :let [beta-bindings (join-node-matches node join-filter-fn token fact {})]
            :when beta-bindings]
        (->Token (conj (:matches token) [fact id])
-                (conj (:bindings token) bindings beta-bindings))))))
+                (conj (:bindings token) bindings beta-bindings)))))
+
+  IConditionNode
+  (get-condition-description [this]
+    (let [{:keys [type constraints original-constraints]} condition
+          full-constraints (if (seq original-constraints)
+                             original-constraints
+                             constraints)]
+      (into [type] full-constraints))))
 
 ;; The NegationNode is a beta node in the Rete network that simply
 ;; negates the incoming tokens from its ancestors. It sends tokens
@@ -639,13 +784,18 @@
     (l/right-retract! listener node elements)
     (mem/remove-elements! memory node join-bindings elements)
     (when (empty? (mem/get-elements memory node join-bindings))
-      (send-tokens transport memory listener children (mem/get-tokens memory node join-bindings)))))
+      (send-tokens transport memory listener children (mem/get-tokens memory node join-bindings))))
+
+  IConditionNode
+  (get-condition-description [this]
+    (let [{:keys [type constraints]} condition]
+      [:not (into [type] constraints)])))
 
 (defn- matches-some-facts?
   "Returns true if the given token matches one or more of the given elements."
-  [token elements join-filter-fn condition]
+  [node token elements join-filter-fn condition]
   (some (fn [{:keys [fact]}]
-          (join-filter-fn token fact (:env condition)))
+          (join-node-matches node join-filter-fn token fact (:env condition)))
         elements))
 
 ;; A specialization of the NegationNode that supports additional tests
@@ -665,7 +815,8 @@
                  listener
                  children
                  (for [token tokens
-                       :when (not (matches-some-facts? token
+                       :when (not (matches-some-facts? node
+                                                       token
                                                        (mem/get-elements memory node join-bindings)
                                                        join-filter-fn
                                                        condition))]
@@ -682,7 +833,8 @@
                     ;; Retract only if it previously had no matches in the negation node,
                     ;; and therefore had an activation.
                     (for [token tokens
-                          :when (not (matches-some-facts? token
+                          :when (not (matches-some-facts? node
+                                                          token
                                                           (mem/get-elements memory node join-bindings)
                                                           join-filter-fn
                                                           condition))]
@@ -708,11 +860,13 @@
                             ;; smaller than the previous elements most of the time
                             ;; and that the time to check the elements will be proportional
                             ;; to the number of elements.
-                            :when (and (matches-some-facts? token
+                            :when (and (matches-some-facts? node
+                                                            token
                                                             elements
                                                             join-filter-fn
                                                             condition)
-                                       (not (matches-some-facts? token
+                                       (not (matches-some-facts? node
+                                                                 token
                                                                  previous-elements
                                                                  join-filter-fn
                                                                  condition)))]
@@ -736,15 +890,26 @@
 
                        ;; Propagate tokens when some of the retracted facts joined
                        ;; but none of the remaining facts do.
-                       :when (and (matches-some-facts? token
-                                                     elements
-                                                     join-filter-fn
-                                                     condition)
-                                  (not (matches-some-facts? token
+                       :when (and (matches-some-facts? node
+                                                       token
+                                                       elements
+                                                       join-filter-fn
+                                                       condition)
+                                  (not (matches-some-facts? node
+                                                            token
                                                             (mem/get-elements memory node join-bindings)
                                                             join-filter-fn
                                                             condition)))]
-                   token))))
+                   token)))
+
+  IConditionNode
+  (get-condition-description [this]
+    (let [{:keys [type constraints original-constraints]} condition
+          full-constraints (if (seq original-constraints)
+                             original-constraints
+                             constraints)]
+      [:not (into [type] full-constraints)])))
+
 
 ;; The test node represents a Rete extension in which
 (defrecord TestNode [id test children]
@@ -1149,12 +1314,20 @@
                                      transport memory listener))
               (doseq [token matched-tokens]
                 (send-accumulated node accum-condition accumulator result-binding token retracted-converted bindings
-                                  transport memory listener)))))))))
+                                  transport memory listener))))))))
+
+  IConditionNode
+  (get-condition-description [this]
+    (let [{:keys [accumulator from]} accum-condition
+          {:keys [type constraints]} from
+          condition (into [type] constraints)
+          result-symbol (symbol (name result-binding))]
+      [result-symbol '<- accumulator :from condition])))
 
 (defn- filter-accum-facts
   "Run a filter on elements against a given token for constraints that are not simple hash joins."
-  [join-filter-fn token candidate-facts]
-  (filter #(join-filter-fn token % {}) candidate-facts))
+  [node join-filter-fn token candidate-facts]
+  (filter #(join-node-matches node join-filter-fn token % {}) candidate-facts))
 
 ;; A specialization of the AccumulateNode that supports additional tests
 ;; that have to occur on the beta side of the network. The key difference between this and the simple
@@ -1180,7 +1353,7 @@
                 [fact-bindings candidate-facts] grouped-candidate-facts
 
                 ;; Filter to items that match the incoming token, then apply the accumulator.
-                :let [filtered-facts (filter-accum-facts join-filter-fn token candidate-facts)]
+                :let [filtered-facts (filter-accum-facts node join-filter-fn token candidate-facts)]
 
                 :when (or (seq filtered-facts)
                           ;; Even if there no filtered facts, if there are no new bindings we may
@@ -1237,7 +1410,7 @@
         (doseq [token tokens
                 [fact-bindings candidate-facts] grouped-candidate-facts
 
-                :let [filtered-facts (filter-accum-facts join-filter-fn token candidate-facts)]
+                :let [filtered-facts (filter-accum-facts node join-filter-fn token candidate-facts)]
 
                 :when (or (seq filtered-facts)
                           ;; Even if there no filtered facts, if there are no new bindings an initial value
@@ -1305,13 +1478,13 @@
       
       (doseq [token matched-tokens
 
-              :let [new-filtered-facts (filter-accum-facts join-filter-fn token candidates)]
+              :let [new-filtered-facts (filter-accum-facts node join-filter-fn token candidates)]
 
               ;; If no new elements matched the token, we don't need to do anything for this token
               ;; since the final result is guaranteed to be the same.
               :when (seq new-filtered-facts)
 
-              :let [previous-filtered-facts (filter-accum-facts join-filter-fn token previous-candidates)
+              :let [previous-filtered-facts (filter-accum-facts node join-filter-fn token previous-candidates)
 
                     previous-accum-result-init (cond
                                                  (seq previous-filtered-facts)
@@ -1407,9 +1580,9 @@
       (doseq [;; Get all of the previously matched tokens so we can retract and re-send them.
               token matched-tokens
               
-              :let [previous-facts (filter-accum-facts join-filter-fn token previous-candidates)
+              :let [previous-facts (filter-accum-facts node join-filter-fn token previous-candidates)
 
-                    new-facts (filter-accum-facts join-filter-fn token new-candidates)]
+                    new-facts (filter-accum-facts node join-filter-fn token new-candidates)]
 
               ;; The previous matching elements are a superset of the matching elements after retraction.
               ;; Therefore, if the counts before and after are equal nothing retracted actually matched
@@ -1462,7 +1635,18 @@
           (not= previous-converted new-converted)
           (do
             (retract-accumulated node accum-condition accumulator result-binding token previous-converted bindings transport memory listener)
-            (send-accumulated node accum-condition accumulator result-binding token new-converted bindings transport memory listener))))))) 
+            (send-accumulated node accum-condition accumulator result-binding token new-converted bindings transport memory listener)))))) 
+
+  IConditionNode
+  (get-condition-description [this]
+    (let [{:keys [accumulator from]} accum-condition
+          {:keys [type constraints original-constraints]} from
+          result-symbol (symbol (name result-binding))
+          full-constraints (if (seq original-constraints)
+                             original-constraints
+                             constraints)
+          condition (into [type] full-constraints)]
+      [result-symbol '<- accumulator :from condition])))
 
 (defn variables-as-keywords
   "Returns symbols in the given s-expression that start with '?' as keywords"
--- a/src/test/clojure/clara/test_rules.clj
+++ b/src/test/clojure/clara/test_rules.clj
@@ -5059,3 +5059,136 @@
     
     (is (not-any? #(= 1 (count %)) @accum-state)
         "Facts with common ancestors should be batched together, expected either the initial accumulator value or a vector containing both lists but never a vector containing one list.")))
+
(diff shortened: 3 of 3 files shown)


src/main/clojure/clara/rules/engine.cljc
```diff
@@ -391,15 +391,74 @@
...
+                  (str "  " exp)
+                  "of condition"
+                  (str "  " condition-definition)
+                  rule-header
+                  rule-list
+                  "when processing fact"
+                  (str "  " (pr-str fact))
+                  bindings-description])]
+    (binding [*out* *err*] (println message))
+    (throw (Exception. message (:exception error)))))
```
WilliamParker: I don't think building the exception like this will work in ClojureScript; I think this relies on JVM interop.  If you look at the Travis build you'll see the message "WARNING: Use of undeclared Var clara.rules.engine/Exception at line 444 /home/travis/build/cerner/clara-rules/src/main/clojure/clara/rules/engine.cljc" and a similar one for \*err\*.

My suggestion is to use ex-info.  Basically, Clojure and ClojureScript define their own exception type that carries a payload of user-defined data.  This user-defined data can then be inspected in a REPL when the exception is thrown during development.   I created a quick gist demonstrating this functionality at https://gist.github.com/WilliamParker/d033cb89ea6ac996115e1712adf5a56a.

The docs on ex-info are at https://clojuredocs.org/clojure.core/ex-info

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -235,15 +235,21 @@
   "Compiles a sequence of constraints into a structure that can be evaluated.
 
    Callers may also pass a collection of equality-only-variables, which instructs
-   this function to only do an equality check on them rather than create a unification binding."
+   this function to only do an equality check on them rather than create a unification binding.
+
+   The returned expression evaluates to a two item vector, [bindings error], which contains
+   the evaluated bindings in the first argument and a map with information on any exception
+   that might have been thrown by evaluating a constraint as the second vector entry."
```
WilliamParker: Thanks for updating the docstring.

WilliamParker:
I think we'll want some tests on this, at least enough to validate that we aren't suppressing exceptions.  My suggestion would be to use the functionality in test_rules.clj for testing that an ExceptionInfo with the expected data is thrown.  Testing such data tends to be less brittle than testing exception messages (i.e. asserting the content and formatting of a string) directly in my experience.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -391,15 +391,74 @@
...
+      (recur (into rest (:children next)) (conj flattened next))
+      (->> flattened
+           (keep :production)
+           (map :name)
+           (distinct)
+           (sort)))))
+
+(defn throw-constraint-exception
+  [error node fact env]
+  "Adds a useful error message when executing a constraint node raises an exception."
```
mrrodriguez: Doc strings go before the arg vector.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -235,15 +235,21 @@
...
   ([exp-seq]
-   (compile-constraints exp-seq #{}))
+   (compile-constraints exp-seq #{} 0))
   ([exp-seq equality-only-variables]
+   (compile-constraints exp-seq equality-only-variables 0))
+  ([exp-seq equality-only-variables constraint-idx]
```
carlosphillips: Agreed, the original implementation published constraint-idx and the node reconciled it. There is no need for constraint-idx now.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -391,15 +391,74 @@
 
   (description [node] (str "QueryNode -- " query)))
 
+
+(defn node-rule-names
+  [children]
+  (loop [[next & rest] children flattened []]
```
carlosphillips: Excellent, I didn't know that tree-seq existed!

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -391,15 +391,74 @@
...
+        rule-header (if (> (count rule-names) 1)
+                      "of rules"
+                      "of rule")
+        unnamed-rule-msg (str "  An unnamed rule, provide names to your rules if"
+                              "you want them to be identified here.")
+        rule-list (->> rule-names
+                       (sort)
+                       (map #(if (nil? %) unnamed-rule-msg %))
+                       (map #(str "  " %))
+                       (clojure.string/join "\n"))
```
carlosphillips: ok

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -391,15 +391,74 @@
...
   (description [node] (str "QueryNode -- " query)))
 
+
+(defn node-rule-names
+  [children]
+  (loop [[next & rest] children flattened []]
+    (if next
+      (recur (into rest (:children next)) (conj flattened next))
+      (->> flattened
+           (keep :production)
```
carlosphillips: Ok, let me look at that.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -391,15 +391,74 @@
...
+                 ["Exception raised during execution of constraint"
+                  (str "  " exp)
+                  "of condition"
+                  (str "  " condition-definition)
+                  rule-header
+                  rule-list
+                  "when processing fact"
+                  (str "  " (pr-str fact))
+                  bindings-description])]
+    (binding [*out* *err*] (println message))
```
carlosphillips: No, mostly just to demo the feature.

src/main/clojure/clara/rules/engine.cljc
carlosphillips: Ok, that makes sense. Thank you for the demo. I'll change to ex-info.

src/main/clojure/clara/rules/engine.cljc
carlosphillips: Ok

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -391,15 +391,74 @@
...
+           (sort)))))
+
+(defn throw-constraint-exception
+  [error node fact env]
+  "Adds a useful error message when executing a constraint node raises an exception."
+  (let [children (:children node)
+        exp (:constraint-exp error)
+        rule-names (node-rule-names children)
+        condition (or (:condition node)
+                      (-> children first :condition))
```
carlosphillips: That's correct. Let me know. This or expression is a bit opaque here anyway.

carlosphillips:
I pushed a changeset trying to address all of the recommendations. Let me know whether this was the right way to proceed?

WilliamParker:
@carlosphillips We prefer to have a single commit in the history for each enhancement, bug fix, etc.  Could you squash your commits for this together into one commit and update the pull request?  I'll plan to look at the changes sometime over the next couple days.

carlosphillips:
Ok, I have squashed the commits and pushed it through.

WilliamParker:
I've looked over the changes @carlosphillips .  There were some outstanding issues with how this is retrieving conditions from nodes and a couple of other minor things that should be quick to fix.  On retrieving conditions, I'd like to resolve the issue with getting the condition off of AccumulateWithJoinFilterNode but don't personally mind if we leave the edge case of dealing with alpha nodes with children with different conditions for another time; I don't know others views on that though.  Note that the conditions would only be different in being ordinary conditions, negations, or accumulation; the constraints would the same.  From previous comments it sounds like @rbrush may want to look over this as well before merging.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -391,15 +391,94 @@
...
+  [error node fact env]
+  (let [exp (:constraint-exp error)
+        productions (node-rule-names :production node)
+        queries (node-rule-names :query node)
+        production-section (list-of-names "rule" "rules" productions)
+        query-section (list-of-names "query" "query" queries)
+        condition (or (:condition node) ; beta node
+                      (-> node :children first :condition) ; alpha node
+                      (-> node :children first :accum-condition) ; accum node
+                      )
```
carlosphillips: I ended up implementing a couple of small protocols and a couple of new functions based on your comments. We can now handle multiple different conditions containing the same constraint. Here's an example output.
```
Constraint exception.
Exception raised during execution of constraint
  (> 1 rate)
when processing fact
  #msg_test.RateToUSD{:currency "CAD", :rate nil}
with no bindings

Conditions:

1. [msg_test.RateToUSD (> 1 rate)]
   rules:
     rule1-name
     rule3-name
   queries:
     query1-name

2. [:?wind <- (clara.rules.accumulators/all) :from [msg_test.RateToUSD (> 1 rate)]]
   rules:
     rule2-name
```

src/test/clojure/clara/test_rules.clj
```diff
@@ -5025,3 +5025,129 @@
...
+  (let [query-template (dsl/parse-query [] [[Temperature
+                                             (= ?location location)
+                                             (> temperature 0)]])
+        temperature-query1 (assoc query-template :name "temperature-query1")
+        temperature-query2 (assoc query-template :name "temperature-query2")
+        temperature-fact (->Temperature nil "MCI")]
+    (assert-ex-data {:constraint-exp '(> temperature 0)
+                     :bindings {:?location "MCI"}
+                     :fact temperature-fact
+                     :queries ["temperature-query1" "temperature-query2"] ; ordering may change with implementation details
```
carlosphillips: Makes sense

src/test/clojure/clara/test_rules.clj
```diff
@@ -5025,3 +5025,129 @@
...
+        rule-template (dsl/parse-rule [[Temperature
+                                        (= ?location location)
+                                        (> temperature 0)]]
+                                      (println "Ok"))
+        temperature-rule (assoc rule-template :name "temperature-rule")
+        temperature-fact (->Temperature nil "MCI")]
+    (assert-ex-data {:constraint-exp '(> temperature 0)
+                     :bindings {:?location "MCI"}
+                     :fact temperature-fact
+                     :queries ["temperature-query"] ; ordering may change with implementation details
```
carlosphillips: Yes

src/test/clojure/clara/test_rules.clj
```diff
@@ -5025,3 +5025,129 @@
...
+
+(deftest test-single-condition-accum-constraint-exception
+  (let [query-template (dsl/parse-query [] [[?wind <- (acc/all) :from [Temperature (> temperature 0)]]])
+        temperature-query (assoc query-template :name "temperature-query")
+        temperature-fact (->Temperature nil "MCI")]
+    (assert-ex-data {:constraint-exp '(> temperature 0)
+                     :bindings {}
+                     :fact temperature-fact
+                     :queries ["temperature-query"] ; ordering may change with implementation details
+                     :productions []}
```
carlosphillips: The condition structure was too brittle and I worried that it would easily in the future. Not a problem with the new protocols.

src/main/clojure/clara/rules/engine.cljc
carlosphillips: I still am not handling TestNode properly, but I'll probably need someone else to take that over. It's getting harder to find free time.

carlosphillips:
Hi guys, thank you for having a look.

WilliamParker:
A couple notes on the latest changeset:

1. The comment here isn't accurate:
````
;; Protocol for getting a node's condition expression.
(defprotocol ITerminalNode
  (terminal-node-type [this]))
````
This comment is identical to the comment on IConditionNode:
````
;; Protocol for getting a node's condition expression.
(defprotocol IConditionNode
  (get-condition-description [this]))
````

but these nodes aren't actually doing the same thing.  terminal-node-type returns a tuple of a type keyword and and a name, not a condition.  I don't mind just fixing this myself after a merge if you don't get to it before then.

2. I tried this out in my Emacs REPL and I also got weird formatting on the message string, although when I directly printed out the exception string with

````
(-> *e .getMessage println)
````
it looked the way I'd expect.  I'm not sure whether something is off in the string building here or if something in my development tooling is misbehaving, but in any case this sort of thing shouldn't block merging IMO since useful information with strange formatting is far better than a lack of useful information, which is the current state.

3. You mentioned that TestNode isn't being handled, but I wouldn't expect TestNode to be a direct child of an AlphaNode so an AlphaNode wouldn't get its conditions from TestNode.  TestNode corresponds to a :test condition in a rule like

````
(defrule cold-rule
    [Temperature (= ?t temperature)]
    [:test (< ?t 0)]
    =>
    (insert! (->Cold ?t)))
````

I'd consider adding useful error information to TestNode to be a separate, although related, issue that shouldn't block merging this.

In short, these changes look good to me and I don't see any further blockers to merging.  I'll hold off on merging for now though to give @rbrush a chance to look at this if he wants to do so.

**Comment by carlosphillips** [comment]
I'm getting the same thing as you in the Emacs REPL. I'm not sure how to fix that. I did fix the comment.

**Comment by carlosphillips** [comment]
Btw, you provide excellent feedback and it has been a pleasure working with you.

File: src/main/clojure/clara/rules/compiler.clj (hunk 2 above)
Comment (rbrush) [inline src/main/clojure/clara/rules/compiler.clj]: Apologies for being slow here...I had been traveling and then got swamped with other items.

Overall I think this looks good. One thing that's not clear to me: rather than returning a [bindings error] tuple, could the compiled function simple return the bindings (as it did before) or throw a specific type of an exception that the caller could handle when an error occurs? The thrown exception could include the same information as seen in the error in the returned tuple, so we wouldn't lose anything.

It's probably not a huge deal, but throw/catching a specialized exception here would save us from creating and de-structuring an extra for every time we run our constraint expression.

Comment (rbrush) [inline src/main/clojure/clara/rules/compiler.clj]: I know Carlos is busy with other commitments so I'm happy to help make these adjustments, but first I just wanted to understand if my reasoning is valid.

File: src/main/clojure/clara/rules/engine.cljc [hunk 14]
```diff
@@ -389,17 +422,96 @@
...
+                 (dissoc :exception)
+                 (merge {:conditions-and-rules conditions-and-rules
+                         :fact fact
+                         :node node}))]
+    (throw (ex-info message info cause))))
+
+(defn- alpha-node-matches
+  [facts env activation node]
+  (for [fact facts
+        :let [[bindings error] (activation fact env)]
```
rbrush: Similar to the above comments...rather than returning a [bindings error] tuple, I wonder if we could just throw a specific exception we'd only catch for this case that contains the error, saving us the creation/destructuring of a tuple that only helps in an exception case.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Regarding the cost of using the error/bindings tuples, that is a good point @rbrush and I agree.  Obviously the cost of instantiating and destructuring the vector is small but it could add up.  We've seen some very tiny optimizations in our rule constraints that end up being part of the AlphaNode filters in particular be noticeable before.

 I think the reason this was implemented this way is that it derives a lot of information about the context in which the exception was thrown - i.e. the entire condition, the productions(rules/queries) that the constraint is a part of, etc. by examining the rule network at the time the exception is thrown.  In principle this information could be determined at compilation time rather than re-derived but it looked to me like this would involve significant development work in the compiler and since this analysis is only done when an exception is thrown, and thus should not be a performance problem, I suggested that we stick with deriving the information at runtime.

However, on further examination this does not preclude the approach you suggest.  Take part of the code from the current PR:

````
(defn- alpha-node-matches
  [facts env activation node]
  (for [fact facts
        :let [[bindings error] (activation fact env)]
        :when (or bindings error)] ; FIXME: add env.
    (if error
      (throw-constraint-exception error node fact env)
      [fact bindings])))
````

The throw-constraint-exception function is called with error, node, fact, and env.  The contents of error are derived from fact and env by the activation call.  Therefore, if we called the activation with the node in addition to its existing arguments it could derive all this information itself and then throw the exception itself.  I don't think this would be too difficult to do; I would anticipate that we could just keep the existing infrastructure Carlos has built to gather the information, format the messages, etc. and just call it elsewhere i.e. in the compiled filter or join function.

Regarding throwing a specialized exception, are you suggesting that we implement our own exception class?  If so I suggest that it implement IExceptionInfo, which would not be too difficult to do.  I've found the ability to inspect data relevant to an exception programmatically at the REPL to be quite valuable.  However, I'm not sure this is something that adds much value.  It isn't obvious to me why a consumer would only want to catch such a specialized exception type as opposed to a more general exception type.  For example, obviously I'd hope this doesn't happen, but if Clara itself threw an exception due to a bug I'd think a consumer would want to handle it in much the same way as an exception from a function call in their rule constraints.  I'm not opposed to making our own type if we land on use-cases for it though, it just seems somewhat premature to me if we don't know of a need for it yet, especially since this would be platform-specific work i.e. different between Clojure and ClojureScript.

I agree that especially since Carlos has other commitments we should be willing to get this over the finish line.  This sort of error handling is likely to be very beneficial.  Regarding the procedural aspects of that I'm not quite sure what form that would take; I do want to be sure that the Git history ends up appropriately crediting Carlos for this significant piece of work.

Comment (WilliamParker) [inline src/main/clojure/clara/rules/compiler.clj]: Looking at this again I think I may have misinterpreted parts of @rbrush 's comment.  We could do something like

````
;;; compiler
(compile-constraints
 (try constraint-stuff
      (catch Exception e
        (throw (ex-info "message" {:exception e
                                   :constraint-exp val-here
                                   :bindings val-here
                                   :env env
                                   ::constraint-exception true})))))
;; engine
(defn- alpha-node-matches
  [facts env activation node]
  (for [fact facts
        :let [bindings (try (activation fact env)
                            (catch #?(:clj Exception :cljs :default) e
                              (if (-> e ex-data :clara.rules.compiler/constraint-exception)
                                (throw-constraint-exception
                                 (-> e ex-data
                                     (dissoc :clara.rules.compiler/constraint-exception)
                                     node
                                     fact
                                     env))
                                (throw e)
                                ;; Or add some info in the else if we want but that isn't necessary for this enhancement.
                                )))]
        :when bindings] ; FIXME: add env.
    [fact bindings]))
````

I think this is what Ryan was actually getting at, but he can speak to whether my first interpretation or this one is correct, or if I was misinterpreting both times. :)  This sort of approach is probably even easier than what I suggested before.  The one thing I did change here is to use a key in an ex-data map to dispatch on whether the exception was thrown from a constraint rather than using a different exception type which I think is probably easier than needing to define a new exception type for both Clojure and ClojureScript.

Comment (rbrush) [inline src/main/clojure/clara/rules/compiler.clj]: Yep, this is what I had in mind.  I hadn't worked out the specific like you did, but this seems like a more efficient and idiomatic approach. If it turns out there is common exception handling code we could shift it into a separate function called from multiple places.

Comment (carlosphillips) [inline src/main/clojure/clara/rules/compiler.clj]: This makes a lot of sense. I'm happy to adjust and push.

Comment (carlosphillips) [inline src/main/clojure/clara/rules/compiler.clj]: I'm going to push soon. Unfortunately, I'm generating a bunch of ClojureScript dead code warnings when compiling tests because the compiler has picked up that ClojureScript expressions such as  `(< degrees 20)` cannot throw. However, `(< nil 20)` does throw in Clojure. There is no correctness issue, but it does pollute the output of `lein compile`.

Comment (rbrush) [inline src/main/clojure/clara/rules/compiler.clj]: Hmm, bummer about the noise in the Clojurescript code. 

I do wonder if we could eliminate the try/catch per expression without losing much, and just have the try/catch in the alpha-node-matches in Will's example. So we'd let the exception thrown by an expression propagate to alpha-node-matches, and do the work there. We'd have the node that caused the exception there, and can get the full constraint from that node.

That should remove the ClojureScript dead code complaints and eliminate a bunch of generated try/catches, at the cost of some reduced granularity...but users will still see the exact node, high-level expression, and rules that caused it. 

Does this seem like it's worth considering?



 How much context do we lose if we just catch the in the outer expression

Comment (carlosphillips) [inline src/main/clojure/clara/rules/compiler.clj]: That makes sense, but can we get the exact throwing constraint this way? Sequences of constraints are often compiled together by `compile-constraints`. 

We can get `?__bindings__` by capturing exceptions in `compile-condition` and `compile-join-filter` functions. We can do similarly for constraint expressions if we store the current expression in an atom just like bindings, but I'm not sure whether that would outperform the original vector-based approach. I imagine that atoms themselves have some overhead for thread synchronization. Also, we would no longer be able to distinguish between a constraint expression throwing and Clara throwing. This distinction can have an impact on users' confidence in the system.

Beyond setting up some benchmarks, I'm not sure how to make the performance call.

Comment (carlosphillips) [inline src/main/clojure/clara/rules/compiler.clj]: Also, I forgot to update the comment for `compile-constraints`. It continues to mention a return vector.

Comment (carlosphillips) [inline src/main/clojure/clara/rules/compiler.clj]: Probably not the way we want to go, but for the record the following cljs compiler setting turns off the warnings.
 
:closure-warnings {:check-useless-code :off}

Comment (WilliamParker) [inline src/main/clojure/clara/rules/compiler.clj]: It looks like Closure, which is used by ClojureScript and I believe is ultimately the source of these warnings, supports suppressing them in particular parts of JavaScript code.  Perhaps ClojureScript has some support for passing such annotations through?  I couldn't immediately find anything on warning suppression in ClojureScript but it looks like it can pass them through for type annotation: https://github.com/clojure/clojurescript/wiki/Compile-Time-Type-Checking

The problem I see with these warnings is that they'll come up in user code consuming the cljs version of Clara, correct?.  If it were just our build that was the issue it would be simple enough to just replace functions in constraints in the cljs tests that can't throw with ones that can.

Comment (carlosphillips) [inline src/main/clojure/clara/rules/compiler.clj]: I think that you're right about warnings in users' builds, not a good idea.

Comment (WilliamParker) [inline src/main/clojure/clara/rules/compiler.clj]: I gave this some thought and I think I see a way out here.

Steps:
1. We can just eliminate the try-catch in build-constraints if we're compiling ClojureScript.
In essence, this would bring us back to the previous behavior in ClojureScript while keeping the improvement in Clojure.
2. We could then add an option to add such try-catch blocks for ClojureScript that would primarily be used at development time (and would create unreachable code warnings in some cases).
3. We can then add information to the "else" case where there the exception isn't from a constraint
in the engine.  The error message could specify that the exception might be from either Clara itself or user code if that is desired. If the session is a ClojureScript one the message could reference the option created in (2) above.  

Normally I'd be pretty reluctant to add the complication of another user-facing option, but this one would have the nice property that anyone who needs it can have it very directly pointed out by the exception message, and the user can then recompile with that option on if they want more information.  For immediate purposes, I'd be fine with just implementing (1) for now and logging work to do (2) and (3).

Thoughts?  @rbrush @carlosphillips

Comment (rbrush) [inline src/main/clojure/clara/rules/compiler.clj]: A variation of Option 1 would be to keep the outer, node-level try/catch in ClojureScript, and the fine-grained, per-expression try/catch in Clojure. This way ClojureScript users could at least see the entire expression that threw an exception (although not the specific constraint.)

I'd be in favor of that variation, or Will's Option 1 as he described it as well. I think it's okay that we offer slightly different error reporting on these platforms based on their capabilities, especially since the overall flow is the same.

**Comment by rbrush** [comment]
One more general comment on this: we need to make sure the presence of a try/catch for each constraint doesn't hurt performance. try/catch is cheap on the JVM, but this flow here could add tens of millions of flows through try/catch blocks for some of our workloads.

If there is a significant performance penalty to this, we might revert to having the outer, node-level try/catch. It won't tell you the exact sub-expression, but it will at least get you to the right top-level constraint.

**Comment by rbrush** [comment]
Another thing we could do when eval'ing the constraint is to attach metadata to each expression in the constraint that includes a line number. So a constraint like:

```clj
 [Temperature (>temperature 40) (= "LAX" location)] 
```

could generate code where ```(>temperature 40)``` is line 1, and ```(= "LAX" location)``` is line 2, which would show up in the exception. We could then catch an exception on each node, and look at the line number in the stack trace to determine exactly which expression threw it.

Of course, this flow would only work on the JVM, but is a way to get back to the exact expression that threw the exception without needing try/catch blocks everywhere. Something we could consider if we're worried about code generation size or performance here.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: @rbrush to clarify, that wasn't exactly an options list.  1, 2, and 3 above are proposed sequential pieces of work.

rbrush: @WilliamParker Gotcha. What I meant was I'd be fine with step 1 for now, even if we deferred 2 and 3 for ClojureScript indefinitely.

WilliamParker:
@rbrush Regarding the performance of try-catch, I did some reading on the subject and it doesn't sound like we should expect much if any performance impact.  It seems that when the JVM throws an exception in any bytecode instruction, it looks up the handler for the exception in a table that is indexed on the range of bytecode instruction positions for which the handler should be used.  The program counter, the position of the bytecode instruction the VM is currently on, is used to look up the needed handler in this table.  Thus, adding try-catch catch shouldn't add additional bytecode instructions to the main path.  It would add to the total memory taken by compiled classes, but my expectation would be that in the environments Java typically runs in (i.e. server-size machines) this probably isn't a major issue.  I suppose it could be an issue if anyone were trying to run huge rulesets in a highly memory-limited JVM environment but I'm not aware of any such use cases.

It does sound like the introduction of more finely-grained try-catch blocks increases the complexity of JIT optimizations that reorder bytecode instructions at runtime, and thus could reduce them.  I'd expect any such impact to be heavily dependent on the exact constraints, so it will be hard/impossible to prove that such impacts don't exist and in any event would likely be dependent on the VM implementation, possibly OS and processor architecture, etc.  I'll go ahead and test Clara with these changes against one our workflows though and validate that performance doesn't significantly worsen.

Sources:
- [JVM Specification, Section 2.10](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-2.html#jvms-2.10)
- http://stackoverflow.com/questions/16451777/is-it-expensive-to-use-try-catch-blocks-even-if-an-exception-is-never-thrown
- http://stackoverflow.com/questions/25007427/how-are-exceptions-caught-and-dealt-with-at-the-low-assembly-level

WilliamParker:
@rbrush I'm very surprised at this, but I tried this change against one of our workflows that I know taxes the alpha network extremely heavily and it seems this results in double-digit percentage increases in runtime for this workflow.  Relating to the conversation above, I'm guessing this has something to do with hitting less optimal paths in the JIT compiler, which unfortunately is likely to be a pretty difficult issue to diagnose.  I tried inlining the check in alpha-node-matches into the AlphaNode (i.e. putting the code directly in the body instead of in a function) to avoid the Var indirection but this did not noticeably reduce runtime.

Perhaps we can think of a way to bring execution time down or just check on the condition as a whole rather than the constraints.  I wonder if the right solution here is to introduce a :debug-mode option to mk-session that enables this and any other checks that could be too expensive to perform at runtime in production environments.  Aggressive schema checking would be another example of this.  I don't think the additional complexity in the code would be too bad.  Needing to complicate the user-facing API would be unfortunate but is perhaps our best option.  Such an option would have the added bonus of resolving our issues with dead code in ClojureScript.

rbrush:
Gah, bummer on the performance hit.

Even so, I'd still be in favor of checking on the condition as a whole rather than each constraint to avoid the performance penalty. If we can get a user to the right condition, he or she has a much better chance of figuring out the right cause.

mrrodriguez:
@WilliamParker 
>  I'm very surprised at this, but I tried this change against one of our workflows that I know taxes the alpha network extremely heavily and it seems this results in double-digit percentage increases in runtime for this workflow. Relating to the conversation above, I'm guessing this has something to do with hitting less optimal paths in the JIT compiler, which unfortunately is likely to be a pretty difficult issue to diagnose. I tried inlining the check in alpha-node-matches into the AlphaNode (i.e. putting the code directly in the body instead of in a function) to avoid the Var indirection but this did not noticeably reduce runtime.

Did you get any profiling data to be sure it had anything to do with the try-catch situation?
Was it only the alpha network affected?

Another point to consider is if the issue is not the try-catch, but rather causing too large bytecode size for inlining due to the code movement.  You can enable flags to watch for this for the JVM.  I'd do so with Java7+ though due to some odd results I found for some of the inlining output for Java6 before (the Mac version of JDK6 at least).

It is hard to believe that the try-catch blocks are costing time given the actual byte code that is generated by it.  It essentially doesn't add any runtime overhead and only minimal bytecodes.  

I'll demonstrate what I'm seeing from Clojure's compiled code regarding try-catch.

```clj

;;;; WITHOUT try-catch
  (eval '(conj [:a] :b)))

;;;; WITH try-catch
  (eval '(try
           (conj [:a] :b)
              (catch Exception e
                (throw e)))))
```

Now look at the `*compile-path*`, which is something like 
`"/Users/<user>/<path-to-repls-project>/target/classes"`

Using `javap`, I'll truncate to the most relevant part of the discussion.

WITHOUT try-catch
```
  public static java.lang.Object invokeStatic();
    Code:
       0: getstatic     #15                 // Field const__0:Lclojure/lang/Var;
       3: invokevirtual #20                 // Method clojure/lang/Var.getRawRoot:()Ljava/lang/Object;
       6: checkcast     #22                 // class clojure/lang/IFn
       9: getstatic     #26                 // Field const__2:Lclojure/lang/AFn;
      12: getstatic     #30                 // Field const__3:Lclojure/lang/Keyword;
      15: invokeinterface #34,  3           // InterfaceMethod clojure/lang/IFn.invoke:(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;
      20: areturn

```

WITH try-catch
```
  public static java.lang.Object invokeStatic();
    Code:
       0: getstatic     #15                 // Field const__0:Lclojure/lang/Var;
       3: invokevirtual #20                 // Method clojure/lang/Var.getRawRoot:()Ljava/lang/Object;
       6: checkcast     #22                 // class clojure/lang/IFn
       9: getstatic     #26                 // Field const__2:Lclojure/lang/AFn;
      12: getstatic     #30                 // Field const__3:Lclojure/lang/Keyword;
      15: invokeinterface #34,  3           // InterfaceMethod clojure/lang/IFn.invoke:(Ljava/lang/Object;Ljava/lang/Object;)Ljava/lang/Object;
      20: astore_0
      21: goto          36
      24: astore_1
      25: aload_1
      26: aconst_null
      27: astore_1
      28: checkcast     #36                 // class java/lang/Throwable
      31: athrow
      32: astore_0
      33: goto          36
      36: aload_0
      37: areturn
    Exception table:
       from    to  target type
           0    21    24   Class java/lang/Exception

```

It looks like Clojure's compiler is doing 2 potentially bad things:
(1) Adding quite a bit more bytecodes.  If the JIT inlining threshold was bordering on the edge of being too high before, this could be enough to send it over the top.
(2) The return value is stored into local variable 0 with and there is a `goto` for the regular flow of execution.  Then the value is loaded back from local variable 0 to return.

I do not think (1) is likely just because I think many of Clara's functions are already compiling into too large of methods for inlining.  This is unfortunate, but nothing that has changed here.

(2) could be suspect though.

WilliamParker:
Frankly, I was pretty stunned as well to see the performance degradation and I'm still not ruling out that I'm missing something or that there is some other piece of this other than the try-catch that is causing problems.  It merits further thought/investigation.

- I went with something that tested the alpha network heavily because that is the only dataset I know of that would test one of these conditions that heavily.  In this case, it taxes alpha conditions that call functions that we've heavily optimized where we found that things like using ThreadLocals instead of dynamic vars and doing identity rather than equality checks on keywords yielded large performance benefits.  As a result of this optimization the cost of the rules engine is (I think) pretty isolated.  I did do some profiling but it isn't quite as conclusive as I'd like since it consists of randomly selected profiles from a MapReduce job.
- The difference in the bytecode in the main path is interesting since the literature I read on the JVM suggests that the only difference should be the addition of the exception handler.  See my comment at https://github.com/cerner/clara-rules/pull/256#issuecomment-274877383  Comparing this with the bytecode generated from similar logic in Java might be enlightening. That said, I do think we need to keep in mind that the bytecode is ultimately an abstraction, not what actually gets executed.  As I discussed in my comment there is some fairly authoritative information that adding try-catch blocks prevents some reordering operations of the JVM that would otherwise be allowed to optimize performance.  Unfortunately I have yet to find any sources that elaborate on what sorts of optimizations would be prevented.  Note that my tests were against Java 6.
- If we can achieve our performance goals while wrapping all conditions, rather than constraints, in try-catch blocks that would be a reasonable solution IMO.  As @carlosphillips pointed out it does have the downside of making internal exceptions not necessarily readily distinguishable from exceptions in user code, but the benefits of providing the condition rather than nothing outweigh this downside by a considerable margin IMO.  That said, we'd need to check the performance of that as well.  The case I used has predominantly two constraints in the poorly-performing conditions; if performance degradation were a straight linear function of the number of try-catch blocks I don't think a third of the degradation I saw would be acceptable.  That said I doubt that the relationship between try-catch blocks and performance is quite that simple.

mrrodriguez:
> The difference in the bytecode in the main path is interesting since the literature I read on the JVM suggests that the only difference should be the addition of the exception handler. See my comment at #256 (comment) Comparing this with the bytecode generated from similar logic in Java might be enlightening. That said, I do think we need to keep in mind that the bytecode is ultimately an abstraction, not what actually gets executed. As I discussed in my comment there is some fairly authoritative information that adding try-catch blocks prevents some reordering operations of the JVM that would otherwise be allowed to optimize performance. Unfortunately I have yet to find any sources that elaborate on what sorts of optimizations would be prevented. Note that my tests were against Java 6.

Your references were with respect to the Java compiler (i.e. `javac`).  Clojure's compiler is completely different and can do things much differently.  I'm actually not too happy now that I've seen how Clojure compiles the try-catch in a way that adds obvious runtime overhead.
The bytecode isn't purely "an abstraction".  Adding these local variable stores and loads are additional work.  They could maybe be optimized, but there isn't much guarantees on that.
I'd like to see some additional feedback on this, which I think would have to be asking on the Clojurians #clojure slack or the google groups.

The other issue is with inlining, that's what you're mostly discussing.  Certain sorts of "happens before" relationships have to hold in optimizations and the body of code in the try-catch has to be done in a controlled scope so that when an exception is thrown it is caught at the right time and able to jump to the exception table location.  That has many different ways it could manifest itself as a place that can break up optimizations.  

The highest level way to put it is just that the code between try-catch must all stay one unit and can't shift around before the try starts or after the catch.  That sort of thing.  Places where that matter would be non-obvious.

However, there is also just a straightforward bytecode # limit for JIT inlining.  The numbers vary, but say it is 35 (there are actually several different thresholds).  That just means any method longer than 35 bytecodes cannot be inlined into the caller.  

The try-catch adds about 10 bytecodes to the compiled bytecodes I demonstrate above.  That is 10 bytecodes closer to exceeding an inlining limit threshold value.

WilliamParker:
I agree that there seems to be some variance in your example between what the Clojure compiler is doing and what the Java compiler (according to my references) does for similar logic that could explain performance differences.  Extra bytecode instructions are definitely noticeable/concerning.  I haven't yet compared with the actual output of a Java compiler yet.

The general idea of the sorts of optimizations this could disrupt makes sense to me.  In general, from Clara's POV, the exact impacts would likely be dependent on the actual alpha condition so investigating exactly what is going on with my example wouldn't necessarily say anything about the general case.

mrrodriguez:
Microbenchmarks always have their quirks, but in cases with optimized operations like this:

```clj
(time (dotimes [_ 1E9]
        (+ 1 2)))
;; "Elapsed time: 503.603477 msecs"

(time (dotimes [_ 1E9]
        (try (+ 1 2) (catch Exception e e))))

;; "Elapsed time: 4250.413995 msecs"
``

The performance effect is very large.
 The compiled bytecode:

NO try-catch:
```
  public static java.lang.Object invokeStatic();
    Code:
       0: lconst_1
       1: ldc2_w        #12                 // long 2l
       4: invokestatic  #19                 // Method clojure/lang/Numbers.add:(JJ)J
       7: invokestatic  #23                 // Method clojure/lang/Numbers.num:(J)Ljava/lang/Number;
      10: areturn

```

WITH try-catch
```
  public static java.lang.Object invokeStatic();
    Code:
       0: lconst_1
       1: ldc2_w        #12                 // long 2l
       4: invokestatic  #19                 // Method clojure/lang/Numbers.add:(JJ)J
       7: invokestatic  #23                 // Method clojure/lang/Numbers.num:(J)Ljava/lang/Number;
      10: astore_0
      11: goto          22
      14: astore_1
      15: aload_1
      16: aconst_null
      17: astore_1
      18: astore_0
      19: goto          22
      22: aload_0
      23: areturn
    Exception table:
       from    to  target type
           0    11    14   Class java/lang/Exception

``

It looks like the Clojure number optimizations are all in place in both cases.  So that wasn't messed up due to something unforeseen in the microbenchmark.  However, the movement of the number in memory to and from the local variable seems significant.  Or perhaps it is not able to take some optimizing path due to this.

Just an example.

WilliamParker:
Hmm.. I get very different results from that numerical benchmark.

````
;;; Clojure 1.8.0
user=> (time (dotimes [_ 1E9]                                             
        (try (+ 1 2) (catch Exception e e))))
"Elapsed time: 315.164 msecs"
nil
user=> (System/getProperty "java.version")
"1.6.0_65"
user=> (time (dotimes [_ 1E9]
        (try (+ 1 2) (catch Exception e e))))
"Elapsed time: 303.749 msecs"
nil
user=> 
````

What sort of environment (Clojure version, JVM, OS, etc.) are you running it in?

**Comment by mrrodriguez** [comment]
Comparing with the Java compiler for similar code as the `Number.add()` example above:

```java
package test;

import clojure.lang.Numbers;

public class Test {
    public static Object invokeStatic() {
        return Numbers.num(Numbers.add(1L, 2L));
    }

    public static Object invokeStaticWithTry() {
        try {
            return Numbers.num(Numbers.add(1L, 2L));
            } catch (Exception e) {
            throw e;
        }
    }
}

```

Generates

NO try-catch:
```
  public static java.lang.Object invokeStatic();
    Code:
       0: lconst_1
       1: ldc2_w        #2                  // long 2l
       4: invokestatic  #4                  // Method clojure/lang/Numbers.add:(JJ)J
       7: invokestatic  #5                  // Method clojure/lang/Numbers.num:(J)Ljava/lang/Number;
      10: areturn
```

WITH try-catch:
```
  public static java.lang.Object invokeStaticWithTry();
    Code:
       0: lconst_1
       1: ldc2_w        #2                  // long 2l
       4: invokestatic  #4                  // Method clojure/lang/Numbers.add:(JJ)J
       7: invokestatic  #5                  // Method clojure/lang/Numbers.num:(J)Ljava/lang/Number;
      10: areturn
      11: astore_0
      12: aload_0
      13: athrow
    Exception table:
       from    to  target type
           0    10    11   Class java/lang/Exception
```

Notice that the "WITH try-catch" case doesn't do any more work on the normal path.  It doesn't need to.  The "Exception table" is all the info the JVM needs to know how to jump in the case of an exception.

mrrodriguez:
> Hmm.. I get very different results from that numerical benchmark. 

I wonder if it has to do with JVM.  Although, I'd typically expect 8 to outperform 6.

Clojure:
1.8

JVM:
"1.8.0_112"

WilliamParker:
So getting back to this after test runs using a particularly heavy workflow we have in production against various iterations of this:

- Running with the full constraints try-check increased the runtime of this case by about 32% relative to master.
- Interestingly, when I tried the full constraints try-check but without the exception data map, the runtime was about 17% longer than the baseline.
- When I tried running without the exception data map by removing the let binding around the constraint expression result, and instead directly inlining it into a when conditional, results were similar to those in the case with only the exception data map removed.
- I tried running with the JIT disabled to compare the baseline with the changes without JIT (and thus determine whether to what extent the impacts were due to JIT or the actual bytecode) , but the performance of that was so atrocious that it was obvious the job wouldn't complete.
- When I changed this to just have a try-catch on the condition, rather than the individual constraints, runtime increased by about 1.5%.  It should be noted that fluctuations of several percentage points in runtimes on our production compute cluster are fairly normal.

I didn't find anything in the bytecode of the generated functions I looked at in this case that jumped out at me as likely to cause these kinds of performance problems.  I find the fact that the runtime decreased drastically with the removal of the exception data map quite notable.  My suspicion is that the need to maintain variables across the different constraint try-catch loops is causing problems for the HotSpot JIT compiler.  Unfortunately, I'm skeptical that we'll be able to precisely diagnose what is going on there.  I did some research on the HotSpot JIT to try and determine what here could potentially cause problems with it, but while there is some general information out there there seems to be a lack of really detailed descriptions of the behavior (short of the openJDK source of course).

As far as how to move forward on this, it should be kept in mind that the workflow I tested is a remarkably demanding workflow (which I used for this reason) and that the constraints involved have been heavily optimized well beyond what idiomatic Clojure allows.  I wouldn't expect a more normal workflow to be so heavily impact, but nevertheless I don't think a 32% increase is going to be acceptable.  My inclination is to move forward with the option of just wrapping a try-catch on the condition rather than the individual constraints.  It would be nice to have the exact constraint that failed, but I suspect that practically speaking narrowing down to a particular condition, especially with the fact and bindings from prior nodes in the exception data map, will deliver most of the value here.  Regarding the downside @carlosphillips pointed out about also catching exceptions from within Clara, on reflection I think this is probably a good thing on the whole.  Clara shouldn't be throwing such exceptions and if they are localized to a particular condition it should hopefully be easier to diagnose any such exceptions that occur.  In the case of JVM Clojure, one intriguing potential future enhancement I see is that since eval is available, we could potentially catch an exception from the condition and only then compile the condition with try-catch on individual constraints.  However, this would be pretty complex and IMO is best left for another time rather than blocking progress here.

Thoughts?  @rbrush @mrrodriguez

rbrush:
> My inclination is to move forward with the option of just wrapping a try-catch on the condition rather than the individual constraints. It would be nice to have the exact constraint that failed, but I suspect that practically speaking narrowing down to a particular condition, especially with the fact and bindings from prior nodes in the exception data map, will deliver most of the value here.

These are my thoughts as well. 

+1 to implementing that here, and we can look at doing something more involved that matches typically code style on the JVM (like attaching line-number metadata as I briefly discussed earlier)  t a later point.

mrrodriguez:
Yeah, I'd just say any significant performance degradation is worth it.  It sounds like the more minimal approach is useful enough for now.  Anything excessively heavy-weight probably needs to be something that could be turned-on in a "debug" mode if it came down to features like that.  Doesn't sound necessary yet here.

WilliamParker:
I have created a branch that changes this to only wrap the entire condition in a try-catch block, rather than each individual constraint, at https://github.com/WilliamParker/clara-rules/tree/fix-constraint-messages.  The commit is at https://github.com/WilliamParker/clara-rules/commit/055b930d910b25cf3176786a62469c5a831165e2.  Note that this is commit is on top of the commit that adds a try-catch block to each individual constraint.

I'll test this against the problematic workflow discussed earlier when I get a chance.  It is basically but not exactly the same as the version I tested against that ran with no major perf degradation, so I don't anticipate different performance characteristics, but I want to make sure.

WilliamParker:
It looks like the draft I posted at https://github.com/WilliamParker/clara-rules/tree/fix-constraint-messages basically isn't impacting performance. :)

@rbrush I've posted more details on the runs I did internally at Cerner since unfortunately I can't post them here.

carlosphillips:
I'm excited to see that you have found a way to make some of this work.

rbrush:
+1 to merge. (Sorry for the slow responses...been kind of a crazy week.)

WilliamParker:
OK, regarding how to merge this, I'll plan to push the state of https://github.com/WilliamParker/clara-rules/commits/issue255-condition-messages-only onto master after validating that the commit history (credited appropriately to Carlos and with appropriate ordering of the commits) looks the same way in master on my fork as in that branch unless anyone has other ideas/suggestions/objections.  I don't particularly care about the mechanics as long as we get those things right and I'm not experienced with these sorts of manual merges of PRs from others in Git so I could certainly be missing better options.

A colleague of mine is looking through the manual performance testing I did but I anticipate putting this in master today or tomorrow unless something comes up.

WilliamParker:
I pushed these changes to master manually since I needed to add some changes on top of the PR was modified.  You can see the commits here:

https://github.com/cerner/clara-rules/commit/0f85ae70135b512de3ca165a0d97609c0bdfe90c
https://github.com/cerner/clara-rules/commit/da0845a95dc00bf98b24306db2ec08f2923244e4

The Travis build for the second commit passed at https://travis-ci.org/cerner/clara-rules/builds/200391914

I don't see anything left to do on this PR.  Closing.

rbrush:
Awesome. Thanks everyone for pushing this through!

Issue #257, opened 2017-01-09
Equals implementation on AlphaRootsWrapper gives false negatives, causing unnecessary cache misses


WilliamParker:
In https://github.com/cerner/clara-rules/pull/242 we did work to improve performance of rulesets where there are productions that match on ancestors of fact types in the session.  The original problem is discussed at https://github.com/cerner/clara-rules/issues/236.  There is some design discussion later in that thread, but the parent post fully describes the original problem.  

When we tried to use this performance improvement we found that we still had the performance problems characteristic of a lack of batching.  The problem appears to be that I incorrectly assumed that we would only have one AlphaRootsWrapper object per distinct fact type, and thus used reference equality in its equals implementation.  The source of restrictions on the distinct instances of the AlphaRootsWrapper is the fact that fact-type->roots [1] is memoized on the fact type, and the AlphaRootsWrapper objects are created inside this function.  However, the ancestors-fn is called inside this function, not outside it.  As a result, there can an AlphaRootsWrapper not just per primary type, but per primary type per descendant fact type.  We thus can have AlphaRootsWrapper objects that have the same roots, but are not equal, and thus the facts for each are not combined in [2].  Since fact-type->roots is memoized on the type of the fact involved, for any given concrete fact type, where the concrete primary fact type is the return value of the :fact-type-fn on a fact in the session.  We will thus have the caching that we did before https://github.com/cerner/clara-rules/pull/242, but facts that have different primary fact types but common ancestors will not be batched for activating the alpha roots corresponding to the common ancestors.

We originally wanted to limit the number of times that the fact type was used, since this can be a data structure of arbitrary complexity.  However, object identity can't be used here as originally implemented for the reasons discussed above.  In practice, we expect most fact types to be of minimal complexity and testing equality on them should be a very fast operation.  My plan is to add the fact type to the AlphaRootsWrapper and use this in the equals implementation.  If we run into cases where this sort of equality testing is expensive we can look into other solutions.  One possibility would be to add node IDs to the AlphaNode, in which case we could test the equality of a set of numbers.  However, we can make the equals check on the fact type have shortcuts that return true when the fact types are identical objects and false when their hashes are not equal, which allows users to have most of the previous performance benefits of checking by identity if user code is structured such that the fact type returned is identical.  Of course, the equals implementation will need to fall back to equality checking if neither of these conditions is met.

@EthanEChristian is working on implementing this.  Once we have a fix for this merged I plan to create a new RC release.

[...]

Issue #236, opened 2016-11-17
Get-alphas-fn batching is suboptimal for condition types with ancestors that also correspond to conditions


WilliamParker:
The get-alphas-fn groups facts by type before returning them to the engine for insertion into the rules network. [1]  In combination with the iteration in flush-updates [2] the effective result is that each alpha node is activated once (with an arbitrary number of facts) per matching type in the facts to be inserted.  When fact types do not have parents that works out so that we minimize the number of alpha activations.  However, when fact types have ancestors this is not the case.  Say something like

````
(definterface Temperature)

(defrecord Cold [temperature] Temperature)

(defrecord Hot [temperature] Temperature)

(defrule cold-weather
      (?c <- [Cold])
      => ;; doesn't matter
      )

(defrule hot-weather
    (?h <- [Hot])
      => ;; doesn't matter
      )

(defrule all-weather
    (?t <- [Temperature])
            => ;; doesn't matter
            )
      

(-> (mk-session [cold-weather hot-weather all-weather])
    (insert (->Cold 0) (->Hot 100))
    fire-rules)
````

We will have one AlphaNode each for Hot, Cold, and Temperature in this rules network; these will be designated H, C, and T.  The current iteration behavior will be

- We gather the nodes matching Cold, namely C and T, and activate them.
- We gather the nodes matching Hot, namely H and T, and activate them.

Note that T is activated twice with one fact instead of once with two facts.  Since Clara is optimized for batching of operations this is inefficient.  For this simple case it is trivial, but we have a case where we will potentially have hundreds of types inheriting from a single parent.  In this case the performance is more significantly degraded.

The purpose of this issue is to track discussion to land on an approach for such batching and to implement it.  This may involve changing the behavior in all cases or making the get-alpha-fn behavior in this regard customizable in some way.  Also note that although I'm logging this issue others may be involved.  @mrrodriguez

1. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules/compiler.clj#L1555
2. https://github.com/cerner/clara-rules/blob/0.13.0-RC4/src/main/clojure/clara/rules/engine.cljc#L198

WilliamParker:
I've been thinking about possible resolutions to this.  I implemented a draft of a replacement get-alphas-fn that would group facts by node: https://github.com/WilliamParker/clara-rules/commit/85e7aa9590a77af43edf530a8c8fa78102d92ade

[...]

PR #258, opened 2017-01-09
Issue 257: Correct alpha batching with multiple ancestors

EthanEChristian:
Issue #257 

@WilliamParker  @kulkarnipushkar

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1493,19 +1493,24 @@
 
 ;; Wrap the fact-type so that Clojure equality and hashcode semantics are used
 ;; even though this is placed in a Java map.
-(deftype AlphaRootsWrapper [roots roots-hash]
+(deftype AlphaRootsWrapper [fact-type ^int fact-type-hash roots]
   Object
   (equals [this other]
-    ;; Since we memoize the function that retrieves these on the fact type, and since the alpha roots for
-    ;; any given type primary type T (excluding descendants and ancestors) are a disjoint
-    ;; set with respect to the alpha roots for any other primary type T2, we know that we will
-    ;; only have one AlphaRootsWrapper object for any given set of alpha roots per get-alphas-fn instance.
-    ;; Therefore we can just use reference equality in the interest of maximizing performance.
-    (identical? this other))
+    (let [other ^AlphaRootsWrapper other]
+      (cond
+        
+        (identical? fact-type (.fact-type other)) 
+        true
+        
+        (not (== fact-type-hash (.fact-type-hash other)))
+        false
+        
+        :else
+        (= fact-type (.fact-type other)))))
+  
   ;; Since know we will need to find the hashcode of this object in all cases just eagerly calculate it upfront
-  ;; and avoid extra calls to hash later.  The hashcode of the Clojure vector of roots would be cached but there
-  ;; is no need to even make the call to that cache.
-  (hashCode [this] roots-hash))
+  ;; and avoid extra calls to hash later.
+  (hashCode [this] fact-type-hash))
 
 (defn- create-get-alphas-fn
   "Returns a function that given a sequence of facts,
@@ -1547,7 +1552,7 @@
                                   ;; removing groups with no alpha nodes here will improve performance on subsequent calls
                                   ;; to the get-alphas-fn with the same fact type.
                                   (keep #(when-let [roots (not-empty (get alpha-roots %))]
-                                           (AlphaRootsWrapper. roots (hash roots))))
+                                           (AlphaRootsWrapper. % (hash %) roots)))
                                   ;; If a user-provided ancestors-fn returns a sorted collection, for example for
                                   ;; ensuring determinism, we respect that ordering here by conj'ing on to the existing
                                   ;; collection.
--- a/src/test/clojure/clara/test_rules.clj
+++ b/src/test/clojure/clara/test_rules.clj
@@ -21,7 +21,11 @@
            ISystemFact]
           [java.util TimeZone]
           [clara.tools.tracing
-           PersistentTracingListener]))
+           PersistentTracingListener]
+          [java.util
+           List
+           LinkedList
+           ArrayList]))
 
 (use-fixtures :once schema.test/validate-schemas)
 
@@ -5025,3 +5029,33 @@
     (is (= (query end-session-equal-facts lousy-weather-query)
            (query end-session-identical-facts lousy-weather-query)
            [{}]))))
+
+(deftest test-alpha-batching-with-multiple-ancestors
+  ;; Validate that two distinct fact types are batched together when passed to a condition on a
+  ;; common ancestor type. We use an accumulator because it provides a convenient way to directly
+  ;; access the facts passed to the condition without exposing Clara internals.
+  ;; see issue 257 for more details of the bug surrounding this test
+  (let [accum-state (atom [])
+        
+        stateful-accum (acc/accum
+                         {:initial-value []
+                          :reduce-fn conj
+                          :retract-fn (fn [items retracted] (remove #{retracted} items))
+                          :convert-return-fn (fn [items] 
+                                               (do 
+                                                 (swap! accum-state conj items)
+                                                 items))})
+        
+        common-ancestor-rule (dsl/parse-rule [[?lists <- stateful-accum :from [List]]]
+                                             ;; don't care about whats inserted
+                                             (->LousyWeather))
+        
+        array-list (ArrayList.)
+        linked-list (LinkedList.)
+        
+        ses (-> (mk-session [common-ancestor-rule])
+              (insert-all [array-list linked-list])
+              (fire-rules))]
+    
+    (is (not-any? #(= 1 (count %)) @accum-state)
+        "Facts with common ancestors should be batched together, expected either the initial accumulator value or a vector containing both lists but never a vector containing one list.")))


src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1493,19 +1493,22 @@
...
 ;; Wrap the fact-type so that Clojure equality and hashcode semantics are used
 ;; even though this is placed in a Java map.
-(deftype AlphaRootsWrapper [roots roots-hash]
+(deftype AlphaRootsWrapper [fact-type fact-type-hash roots]
   Object
   (equals [this other]
     ;; Since we memoize the function that retrieves these on the fact type, and since the alpha roots for
     ;; any given type primary type T (excluding descendants and ancestors) are a disjoint
     ;; set with respect to the alpha roots for any other primary type T2, we know that we will
     ;; only have one AlphaRootsWrapper object for any given set of alpha roots per get-alphas-fn instance.
```
WilliamParker: This comment about having only one AlphaRootsWrapper instance isn't accurate (and never was, which is why this is currently not behaving the way https://github.com/cerner/clara-rules/issues/236 intended it to) and should be removed.

src/test/clojure/clara/test_rules.clj
```diff
@@ -5025,3 +5029,31 @@
...
+                                 ;; don't care about whats inserted
+                                 (->LousyWeather))
+        
+        identity-list (ArrayList.)
+        linked-list (LinkedList.)
+        
+        ses (-> (mk-session [map-rule])
+              (insert-all [identity-list linked-list])
+              (fire-rules))]
+
```
WilliamParker: The issue should work for more a more detailed explanation - +1 to the comment referencing it - but could you add a brief comment here about why you're asserting this?  Something like "Validate that two distinct fact types are batched together when passed to a condition on a common ancestor type.  We use an accumulator because it provides a convenient way to directly access the facts passed to the condition without exposing Clara internals."

src/test/clojure/clara/test_rules.clj
```diff
@@ -5025,3 +5029,31 @@
...
+                          :convert-return-fn (fn [items] 
+                                               (do 
+                                                 (swap! accum-state conj items)
+                                                 items))})
+        
+        map-rule (dsl/parse-rule [[?maps <- stateful-accum :from [List]]]
+                                 ;; don't care about whats inserted
+                                 (->LousyWeather))
+        
+        identity-list (ArrayList.)
```
WilliamParker: This is an ArrayList, not an "identity-list".  I presume this is a remnant of your earlier approach of having this be an IdentityHashMap. :)

src/test/clojure/clara/test_rules.clj
```diff
@@ -5025,3 +5029,31 @@
...
+        stateful-accum (acc/accum
+                         {:initial-value []
+                          :reduce-fn conj
+                          :retract-fn (fn [items retracted] (remove #{retracted} items))
+                          :convert-return-fn (fn [items] 
+                                               (do 
+                                                 (swap! accum-state conj items)
+                                                 items))})
+        
+        map-rule (dsl/parse-rule [[?maps <- stateful-accum :from [List]]]
```
WilliamParker: This isn't a map rule; could you rename it?  Perhaps "common-ancestor-rule"?

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1493,19 +1493,22 @@
...
-    ;; Therefore we can just use reference equality in the interest of maximizing performance.
-    (identical? this other))
+    ;; Therefore we can reference equality in the interest of maximizing performance and fallback to equivalence,
+    ;; as fact-types can be defined by consumers and thus might not be identical.
+    (let [other ^AlphaRootsWrapper other]
+      (and (= fact-type-hash (.fact-type-hash other))
+           (or (identical? fact-type (.fact-type other))
```
mrrodriguez: The `identical?` check is faster than the hash check.  It should be first.

src/test/clojure/clara/test_rules.clj
```diff
@@ -5025,3 +5029,31 @@
...
+        
+        identity-list (ArrayList.)
+        linked-list (LinkedList.)
+        
+        ses (-> (mk-session [map-rule])
+              (insert-all [identity-list linked-list])
+              (fire-rules))]
+    
+    (is (and (not (contains? @accum-state [identity-list]))
+             (not (contains? @accum-state [linked-list])))
```
WilliamParker: I think we'll probably want to do something like

````
(let [accum-state (atom [])]
    (is (not-any? (fn [acc-value] (= (count acc-value) 1)))))
````
instead of this approach using a set and contains? checks.

My understanding is that Clojure considers the behavior of these sort of operations combining Clojure and Java collections to be non-guaranteed.  In this particular case, I don't see a lot of reasons to expect it to break but it would be better to just avoid the issue entirely IMO.

I'm looking at the discussion on http://dev.clojure.org/jira/browse/CLJ-1372 particular the comment that "Rich says "all bets should be off for hasheq/equiv of non-values" where Java collections obviously fall into the class of "non-values"."

File: src/main/clojure/clara/rules/compiler.clj [hunk 7]
```diff
@@ -1493,19 +1493,22 @@
...
     ;; Since we memoize the function that retrieves these on the fact type, and since the alpha roots for
     ;; any given type primary type T (excluding descendants and ancestors) are a disjoint
     ;; set with respect to the alpha roots for any other primary type T2, we know that we will
     ;; only have one AlphaRootsWrapper object for any given set of alpha roots per get-alphas-fn instance.
-    ;; Therefore we can just use reference equality in the interest of maximizing performance.
-    (identical? this other))
+    ;; Therefore we can reference equality in the interest of maximizing performance and fallback to equivalence,
+    ;; as fact-types can be defined by consumers and thus might not be identical.
+    (let [other ^AlphaRootsWrapper other]
+      (and (= fact-type-hash (.fact-type-hash other))
```
mrrodriguez: Using `=` is adding a lot of overhead to this comparison.  What is the type here?  I think it is `Long`.  I'd suggest to use `==` if you are sure neither are nil, otherwise do so nil-safely.

src/test/clojure/clara/test_rules.clj
EthanEChristian: yeah 😞

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1493,19 +1493,22 @@
...
+           (or (identical? fact-type (.fact-type other))
+               (= fact-type (.fact-type other))))))
   ;; Since know we will need to find the hashcode of this object in all cases just eagerly calculate it upfront
-  ;; and avoid extra calls to hash later.  The hashcode of the Clojure vector of roots would be cached but there
-  ;; is no need to even make the call to that cache.
-  (hashCode [this] roots-hash))
+  ;; and avoid extra calls to hash later.
+  (hashCode [this] fact-type-hash))
```
mrrodriguez: So you are sure the `roots` do not matter for the equal and hash code contract?

src/main/clojure/clara/rules/compiler.clj
WilliamParker: - I don't see any scenarios where the hash field would be nil.  We're eagerly population it at the time the AlphaRootsWrapper is created.
- Using == seems reasonable, that should have less overhead.
- Hasheq returns a primitive int: https://github.com/clojure/clojure/blob/clojure-1.8.0/src/jvm/clojure/lang/Util.java#L164
- I think that the call to the Clojure hash function will box the int to an Integer since it doesn't have any type hints preventing that, but I could be wrong.  https://github.com/clojure/clojure/blob/clojure-1.8.0/src/clj/clojure/core.clj#L4961

I agree with changing to ==.  At first glance it seems like the primitive type hints you suggest would be incorrect though.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: In the generic sense of an equals and hashcode contract, since both the equals and hash code depend on the fact type, the equals and hashcode contract would be upheld regardless.  However, in our particular use-case, the AlphaRootsWrapper is being used to logically represent the roots that correspond to a particular concrete fact type that is used in conditions.  There is one-to-one correspondence between a concrete fact type and its set of alpha roots.

mrrodriguez: I was thinking that was the case.  I was just double checking.

src/test/clojure/clara/test_rules.clj
```diff
@@ -5025,3 +5029,33 @@
...
+        stateful-accum (acc/accum
+                         {:initial-value []
+                          :reduce-fn conj
+                          :retract-fn (fn [items retracted] (remove #{retracted} items))
+                          :convert-return-fn (fn [items] 
+                                               (do 
+                                                 (swap! accum-state conj items)
+                                                 items))})
+        
+        common-ancestor-rule (dsl/parse-rule [[?maps <- stateful-accum :from [List]]]
```
WilliamParker: Could you name this something other than ?maps please?  Another leftover from your previous version where these were maps I presume.

src/test/clojure/clara/test_rules.clj
```diff
@@ -5025,3 +5029,33 @@
     (is (= (query end-session-equal-facts lousy-weather-query)
            (query end-session-identical-facts lousy-weather-query)
            [{}]))))
+
+(deftest test-alpha-batching-with-multiple-ancestors
+  ;; Validate that two distinct fact types are batched together when passed to a condition on a
+  ;; common ancestor type. We use an accumulator because it provides a convenient way to directly
+  ;; access the facts passed to the condition without exposing Clara internals.
+  ;; see issue 257 for more details of the bug surrounding this test
+  (let [accum-state (atom #{})
```
WilliamParker: With the new approach of adding all arguments provided to the convert-return-fn to this I think it would be more straightforward/obviously correct to use a sequence rather than a set to avoid discarding any entries.  Admittedly in this case it is safe to assume that a 1-item entry can only be replaced by another 1-item entry but I'd rather just avoid anyone reading the code having to think about it.

WilliamParker:
@EthanEChristian I made a couple minor comments in the tests on your new changeset but otherwise it looks good.  I'll look again before merging for anything I'm missing now.

nrayapati:
+1, looks good for me, I looked at it very high level.

kulkarnipushkar:
+1 looks good

src/test/clojure/clara/test_rules.clj
```diff
@@ -5025,3 +5029,33 @@
...
+        
+        array-list (ArrayList.)
+        linked-list (LinkedList.)
+        
+        ses (-> (mk-session [common-ancestor-rule])
+              (insert-all [array-list linked-list])
+              (fire-rules))]
+    
+    (is (not-any? #(= 1 (count %)) @accum-state)
+        "Facts with common ancestors should be batched together, expected either [] or [ArrayList LinkedList] but never one by itself.")))
```
WilliamParker: Could you clarify this to be "...[], [ArrayList LinkedList], or [LinkedList ArrayList]..."?  This isn't dependent on the ordering between the two lists.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1493,19 +1493,24 @@
...
+      (cond
+        
+        (identical? fact-type (.fact-type other)) 
+        true
+        
+        (not (== fact-type-hash (.fact-type-hash other)))
+        false
+        
+        :else
+        (= fact-type (.fact-type other)))))
```
mrrodriguez: If you know the types of these you are better off using the `Object.equals()` as well.  I don't think you need any features of `clojure.core/=` here.  It is just a waste of time for the non-`=` cases

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1493,19 +1493,24 @@
...
-    ;; only have one AlphaRootsWrapper object for any given set of alpha roots per get-alphas-fn instance.
-    ;; Therefore we can just use reference equality in the interest of maximizing performance.
-    (identical? this other))
+    (let [other ^AlphaRootsWrapper other]
+      (cond
+        
+        (identical? fact-type (.fact-type other)) 
+        true
+        
+        (not (== fact-type-hash (.fact-type-hash other)))
```
mrrodriguez: This would be faster if you give a primitive type hint `^int` on the `deftype` signature for `fact-type-hash`.

I mentioned this in a previous version.
```clj
(deftype AlphaRootsWrapper [fact-type ^int fact-type-hash roots] <etc>)
```

mrrodriguez: e.g.
```clj

(definterface MyI (doIt []))

(deftype MyX [x]
  MyI
  (doIt [this]
    (== x x)))

(deftype MyY [^int y]
  MyI
  (doIt [this]
    (== y y)))

(def ^MyX mx (->MyX (int 10)))
(def ^MyY my (->MyY (int 10)))

(time (dotimes [_ 1E6] (.doIt mx)))
;; "Elapsed time: 12.92941 msecs"

(time (dotimes [_ 1E6] (.doIt my)))
;; "Elapsed time: 5.768262 msecs"
```

These are fairly consistent results too if you run repeatedly.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: The fact-type here is an arbitrary object that is returned by the fact-type-fn.  Using Java rather than Clojure equality semantics seems like surprising/incorrect behavior to me.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: I agree that it is possible to improve the performance by using primitives and ensuring that the call to == is inlined.  My concern is that we'd need to foresee that we wouldn't have any problems from inadvertent boxing that would cause errors.  This is certainly doable, but the performance numbers I'm seeing don't show any noticeable impact from == (it literally doesn't show up in VisualVM for the use case I'm profiling) and very little impact from the Java map operations inside the get-alphas-fn which encapsulate the == calls under discussion.  I'm inclined to go with the simple option for now, i.e. just use == without type hints, and revisit it later if we find a need.

mrrodriguez: > My concern is that we'd need to foresee that we wouldn't have any problems from inadvertent boxing that would cause errors

This should only ever be an int.  The only trouble you could run into would be if there was a overflowing long (or similar) trying to be dumped into the field, which shouldn't happen.  However, it's up to you.  I'm just pointing out it is a doubling of the time.

I'm sure in most use-cases it wouldn't matter much.

WilliamParker: Summary: I've investigated and I'm OK with this change now.

When you add the int type hint to the fact-type-hash field of the AlphaRootsWrapper, AOT the clara.rules.compiler namespace by adding

````
:aot [clara.rules.compiler]
````

to the project.clj and decompile the AlphaRootsWrapper to Java syntax with JD-GUI, the constructor of the AlphaRootsWrapper looks like this:

````
public AlphaRootsWrapper(Object paramObject1, int paramInt, Object paramObject2)
  {
    this.fact_type = paramObject1;this.fact_type_hash = paramInt;this.roots = paramObject2;
  }
````

I then tried decompiling interop calls to the AlphaRootsWrapper constructor and found that the second argument to the constructor was wrapped in clojure.lang.RT/intCast [1].  The call looks like 

````
return new AlphaRootsWrapper(null, RT.intCast(x), null);
````

The Clojure compiler inserts this code at [2].  If \*unboxed-math\* is false, the default, a call to RT.intCast is inserted.  If \*unboxed-math\* is true then a call to RT.uncheckedIntCast [3] is inserted.  If the parameter type is not a primitive, then the JVM opcode for a checked cast is inserted. To validate my understanding of this I cross-referenced the 192 base 10 opcode the compiler inserts against [4]. In either case it looks like passing either boxed types or primitives should be safe.  This is why building an AlphaRootsWrapper like this works, with \*unchecked-math\* set to both true and false:

````
(AlphaRootsWrapper. nil (Integer/valueOf 1) nil)
````

Note that the first and third arguments to the AlphaRootsWrapper wouldn't actually be nil, I just didn't care about them for the purposes of this investigation.

1. https://github.com/clojure/clojure/blob/clojure-1.8.0/src/jvm/clojure/lang/RT.java#L1169
2. https://github.com/clojure/clojure/blob/clojure-1.8.0/src/jvm/clojure/lang/Compiler.java#L893
3. https://github.com/clojure/clojure/blob/clojure-1.8.0/src/jvm/clojure/lang/RT.java#L1429
4. https://en.wikipedia.org/wiki/Java_bytecode_instruction_listings

Comment (mrrodriguez) [inline src/main/clojure/clara/rules/compiler.clj]: @WilliamParker thanks for adding all the details.

I should have mentioned that I too did look a bit into what happens with primitive type hints in the emitted bytecodes.  That's why I mentioned that the only threat would be that Clojure will, unlike Java, throw overflow exceptions when coercing/downcasting by default.  Which is of course changeable with the setting of `*unchecked-math*` (you called it `*unboxed-math*` above, which I think was a typo).

Just for another future note, you can always do:
```clj
  (eval '(AlphaRootsWrapper. nil 1 nil)))
```

And look into the directory specified by the `*compile-path*` for the compiled class files.  It'll have a name derived from the current `*ns*` and probably the word "eval" in the class name.

You can use
```clj
javap -c <file name here>
```

Issue #268, opened 2017-02-20
Reduce transitions between transient and persistent memory and facilitate performance enhancements to fire-rules


WilliamParker:
Currently when a user wishes to retract some facts, insert other facts, and then fire the rules the user must do something like

````
(-> session
    (retract-facts facts1)
    (insert-facts facts2)
    fire-rules)
````

As a result there are 3 transitions between transient and persistent memory.  https://github.com/cerner/clara-rules/issues/184 improved performance for our use cases with large numbers of facts and Rete operations overall, but it did increase the cost of these transitions.  On average, I've seen us spend 1-3% of runtime in these transitions, which while not huge is not insignificant.  

We can cut this down to a single transition by caching the insertions and retractions and only actually executing them when fire-rules is called.  By using a Clojure persistent data structure for this cache we can efficiently maintain the existing semantics around insert and retract returning new sessions.  This also has the benefit of giving fire-rules visibility to and control of the insert and retract operations, which facilitates adding options to control the performance of the Rete operations.  For example, it would facilitate the enhancements discussed at https://github.com/cerner/clara-rules/issues/249.  There was some discussion on this in relation to that enhancement at https://github.com/WilliamParker/clara-rules/commit/af977444cfd345a1caf272df3a69fecc94b04532#commitcomment-20782287 The API I'm envisioning there would then look something like

````
(fire-rules session {:cancelling true})
````
on top of the changes proposed in this issue.

The main downsides I see are

- We're changing behavior in that querying on a session before firing the rules will consistently not work.  However, there was previously no guarantee on this anyway and we have broken scenarios for such query usage with extracted negations as discussed at https://github.com/cerner/clara-rules/issues/149.  In the long run it would likely be better to consistently return nothing from queries  than to inconsistently successfully query.
- I can envision cases where one could benefit from propagating to the LHS before firing the rules, say if a single session with some facts served as a "parent" to many other sessions that would have additional facts and then be fired.  This seems like a fairly small subset of cases though and could be addressed by adding an option to flush the cache without firing the rules if necessary, or even allowing an option to only fire through a particular activation group, which would be a much more general solution for such use cases.
[...]

Issue #249, opened 2016-12-13
Proposal for efficient updates of facts in the session


WilliamParker:
Currently, when we want to change a fact we have to retract the original version and then insert the modified version.  This works, but is inefficient in cases where the modification to the fact will have limited impact on the session.  In our case, one important problem here is that we have a ReferenceTime fact that contains the "current time" that is used in numerous rules in joins with data points that have data points on them; we then ask questions like "is this data point within the last year"?  When we retract ReferenceTime and replace it with a ReferenceTime that is, for example, a few hours later as a practical matter we would not expect much actual impact on the session.  Therefore, in principle, when we have a rule like

````
(defrule last-year-rule
     [ReferenceTime (= ?time time)]
     [?fact <- [FactWithTime (time-based-predicated? ?time time)]]
    =>
    (insert! (->FactThatCausesLotsOfRuleActivations ?fact)))
````

we should be able to avoid any work other than re-checking the join between the ReferenceTime fact and FactWithTime facts when the truth of the time-based-predicate? does not change due to the change in the time.  However, in our current pattern, we completely retract the ReferenceTime and insert a new ReferenceTime.  This causes a significant amount of work in the Rete network that shouldn't be necessary.

I've considered several approaches for reducing this work.

A. Create a special atom or other deref'able type whose changes would be detected by Clara and be reflected in the rule network.  This would be a similar idea to the specialized atoms used in [Reagent](https://github.com/reagent-project/reagent).  This would have the advantage of being fairly simple to implement but would fundamentally introduce a new logical structure to rules that doesn't belong and would only be present for performance reasons.  The appropriate usage of such a structure would be a confusing part of the API.

B. Retract the original facts, and then insert their replacements, but before propagating facts inserted due to activations by the replacements or sending retractions to the relevant alpha nodes, eliminate duplicates where the same fact was both inserted and retracted.  I'm concerned that searching for such pairs would be too costly in large rules sessions though.

C. Propagate a pair of [old-fact replacement-fact] through the Rete network and stop further propagation once we can determine that the replacement will not impact the network state further and all memory modifications necessary to ensure the accuracy of future truth maintenance operations are complete.

Overall at this point I favor option C.  It does not change the fundamental semantics of rules or rule design.  It would require adding an update method to ISession but this seems like an easy to understand operation if one already understands insertions and retractions.

[...]

PR #269, opened 2017-02-20
Issue 268: Cache insertions and retractions until the rules are fired…

WilliamParker:
… to avoid unnecessary transitions between persistent and transient memory

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -1748,50 +1748,82 @@
         (when (flush-updates *current-session*)
           (recur (mem/next-activation-group transient-memory) next-group))))))
 
-(deftype LocalSession [rulebase memory transport listener get-alphas-fn]
+(deftype LocalSession [rulebase memory transport listener get-alphas-fn pending-operations]
   ISession
   (insert [session facts]
-    (let [transient-memory (mem/to-transient memory)
-          transient-listener (l/to-transient listener)]
 
-      (l/insert-facts! transient-listener facts)
-
-      (binding [*pending-external-retractions* (atom [])]
-        ;; Bind the external retractions cache so that any logical retractions as a result
-        ;; of these insertions can be cached and executed as a batch instead of eagerly realizing
-        ;; them.  An external insertion of a fact that matches
-        ;; a negation or accumulator condition can cause logical retractions.
-        (doseq [[alpha-roots fact-group] (get-alphas-fn facts)
-                root alpha-roots]
-          (alpha-activate root fact-group transient-memory transport transient-listener))
-        (external-retract-loop get-alphas-fn transient-memory transport transient-listener))
+    (let [new-pending-operations (conj pending-operations (->PendingUpdate :insertion
+                                                                           ;; Preserve the behavior prior to https://github.com/cerner/clara-rules/issues/268
+                                                                           ;; , particularly for the Java API, where the caller could freely mutate a
+                                                                           ;; collection of facts after passing it to Clara for the constituent
+                                                                           ;; facts to be inserted or retracted.  If the caller passes a persistent
+                                                                           ;; Clojure collection don't do any additional work.
+                                                                           (if (coll? facts)
+                                                                             facts
+                                                                             (into [] facts))))]
 
       (LocalSession. rulebase
-                     (mem/to-persistent! transient-memory)
+                     memory
                      transport
-                     (l/to-persistent! transient-listener)
-                     get-alphas-fn)))
+                     listener
+                     get-alphas-fn
+                     new-pending-operations)))
 
   (retract [session facts]
 
-    (let [transient-memory (mem/to-transient memory)
-          transient-listener (l/to-transient listener)]
-
-      (l/retract-facts! transient-listener facts)
-
-      (binding [*pending-external-retractions* (atom facts)]
-        (external-retract-loop get-alphas-fn transient-memory transport transient-listener))
+    (let [new-pending-operations (conj pending-operations (->PendingUpdate :retraction
+                                                                           ;; As in insert above defend against facts being a mutable collection.
+                                                                           (if (coll? facts)
+                                                                             facts
+                                                                             (into [] facts))))]
 
       (LocalSession. rulebase
-                     (mem/to-persistent! transient-memory)
+                     memory
                      transport
-                     (l/to-persistent! transient-listener)
-                     get-alphas-fn)))
+                     listener
+                     get-alphas-fn
+                     new-pending-operations)))
 
   (fire-rules [session]
 
     (let [transient-memory (mem/to-transient memory)
           transient-listener (l/to-transient listener)]
+
+      ;; We originally performed insertions and retractions immediately after the insert and retract calls,
+      ;; but this had the downside of making a pattern like "Retract facts, insert other facts, and fire the rules"
+      ;; perform at least three transitions between a persistent and transient memory.  Delaying the actual execution
+      ;; of the insertions and retractions until firing the rules allows us to cut this down to a single transition
+      ;; between persistent and transient memory.  There is some cost to the runtime dispatch on operation types here,
+      ;; but this is presumably less significant than the cost of memory transitions.
+      ;;
+      ;; We perform the insertions and retractions in the same order as they were applied to the session since
+      ;; if a fact is not in the session, retracted, and then subsequently inserted it should be in the session at
+      ;; the end.
+      (doseq [{op-type :type facts :facts} pending-operations]
+
+        (case op-type
+
+          :insertion
+          (do
+            (l/insert-facts! transient-listener facts)
+
+            (binding [*pending-external-retractions* (atom [])]
+              ;; Bind the external retractions cache so that any logical retractions as a result
+              ;; of these insertions can be cached and executed as a batch instead of eagerly realizing
+              ;; them.  An external insertion of a fact that matches
+              ;; a negation or accumulator condition can cause logical retractions.
+              (doseq [[alpha-roots fact-group] (get-alphas-fn facts)
+                      root alpha-roots]
+                (alpha-activate root fact-group transient-memory transport transient-listener))
+              (external-retract-loop get-alphas-fn transient-memory transport transient-listener)))
+
+          :retraction
+          (do
+            (l/retract-facts! transient-listener facts)
+
+            (binding [*pending-external-retractions* (atom facts)]
+              (external-retract-loop get-alphas-fn transient-memory transport transient-listener)))))
+
       (fire-rules* rulebase
                    (:production-nodes rulebase)
                    transient-memory
@@ -1803,7 +1835,8 @@
                      (mem/to-persistent! transient-memory)
                      transport
                      (l/to-persistent! transient-listener)
-                     get-alphas-fn)))
+                     get-alphas-fn
+                     [])))
 
   ;; TODO: queries shouldn't require the use of transient memory.
   (query [session query params]
@@ -1849,7 +1882,8 @@
                  (if (> (count listeners) 0)
                    (l/delegating-listener listeners)
                    l/default-listener)
-                 get-alphas-fn))
+                 get-alphas-fn
+                 []))
 
 (defn local-memory
   "Returns a local, in-process working memory."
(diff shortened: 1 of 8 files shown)


src/test/clojure/clara/test_accumulators.clj
```diff
@@ -59,10 +59,15 @@
         session (-> empty-session
                     (insert (->Temperature 30 "MCI"))
                     (insert (->Temperature 10 "MCI"))
-                    (insert (->Temperature 80 "MCI")))
+                    (insert (->Temperature 80 "MCI"))
+                    fire-rules)
```
curtman16: a lot of tests to update to add a fire-rules call
do we have concerns around how this would impact consumers necessitating a call to fire rules?

src/test/clojure/clara/test_accumulators.clj
```diff
@@ -59,10 +59,15 @@
...
 
-        min-retracted (retract session (->Temperature 10 "MCI"))
-        max-retracted (retract session (->Temperature 80 "MCI"))]
+        min-retracted (-> session
```
curtman16: there are no assertions in this test on this
assuming we'd want an assertion on min-retracted

src/test/clojure/clara/test_rules.clj
```diff
@@ -5187,3 +5260,30 @@
                         (fire-rules)
                         (query temperature-query)))))
 
+(deftest test-stored-insertion-retraction-ordering
```
curtman16: can we add a more robust somewhere for these?  testing multiple insertions/retractions prior to firing rules and then repeating multiple insertions and retractions is what I'd be looking for.

src/test/clojure/clara/test_accumulators.clj
WilliamParker: You are correct, looks like we don't have an assertion on that currently at https://github.com/cerner/clara-rules/blob/0.13.0/src/test/clojure/clara/test_accumulators.clj#L64  That should be fixed (it can be with this or another quick commit if we don't make these changes)

src/test/clojure/clara/test_accumulators.clj
WilliamParker: Short version: I do have concerns, but the current behavior isn't reliable anyway as discussed in https://github.com/cerner/clara-rules/issues/268#issue-208938988 specifically the first bullet, isn't a public contract, and simply changing it seems like the lesser of evils to me.  Making using this sort of caching or non-caching a user-facing option would be one possibility, but I hesitate to permanently add the complication of another user-facing option solely for passivity when the original behavior was unreliable anyway, wasn't a part of the API contract, and the uplift for anyone who runs into the issue should be pretty trivial.  That said, it is possible that others think differently/see other concerns  which is a major reason why I solicited input on the parent issue.

src/test/clojure/clara/test_rules.clj
WilliamParker: OK.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1748,50 +1748,74 @@
...
-        ;; of these insertions can be cached and executed as a batch instead of eagerly realizing
-        ;; them.  An external insertion of a fact that matches
-        ;; a negation or accumulator condition can cause logical retractions.
-        (doseq [[alpha-roots fact-group] (get-alphas-fn facts)
-                root alpha-roots]
-          (alpha-activate root fact-group transient-memory transport transient-listener))
-        (external-retract-loop get-alphas-fn transient-memory transport transient-listener))
+    (let [new-pending-operations (conj pending-operations (->PendingUpdate :insertion (if (coll? facts)
+                                                                                        facts
+                                                                                        (into [] facts))))]
```
mrrodriguez: What is the purpose of putting the `facts` into a vector here?  It seems like it could just end up being a waste of time.  Are you concerned with the mutability of the collection?  I'm not sure if that is a concern that should be paid for with a perf penalty like this.  If the caller is mutating the same collection of `facts` they passed to the session, the expected semantics would be mysterious anyways.

I'd say do `seq` instead, but it'd still be affected by the backing collection beyond the first (32 element) chunk.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1748,50 +1748,74 @@
...
+
+      ;; We originally performed insertions and retractions immediately after the insert and retract calls,
+      ;; but this had the downside of making a pattern like "Retract facts, insert other facts, and fire the rules"
+      ;; perform at least three transitions between a persistent and transient memory.  Delaying the actual execution
+      ;; of the insertions and retractions until firing the rules allows us to cut this down to a single transition
+      ;; between persistent and transient memory.  There is some cost to the runtime dispatch on operation types here,
+      ;; but this is presumably less significant than the cost of memory transitions.
+      ;;
+      ;; We perform the insertions and retractions in the same order as they were applied to the session since
+      ;; if a fact is not in the session, retracted, and then subsequently inserted it should be in the session at
```
mrrodriguez: I dislike these semantics still.  I think the order-dependence is brittle and it is limiting as far as engine optimizations.  However, I accept you are just preserving what has been done before.

Just wanted to add a rant'y comment here.  😛

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1748,50 +1748,74 @@
...
   (fire-rules [session]
 
     (let [transient-memory (mem/to-transient memory)
           transient-listener (l/to-transient listener)]
+
+      ;; We originally performed insertions and retractions immediately after the insert and retract calls,
+      ;; but this had the downside of making a pattern like "Retract facts, insert other facts, and fire the rules"
+      ;; perform at least three transitions between a persistent and transient memory.  Delaying the actual execution
+      ;; of the insertions and retractions until firing the rules allows us to cut this down to a single transition
+      ;; between persistent and transient memory.  There is some cost to the runtime dispatch on operation types here,
```
mrrodriguez: The "runtime dispatch" cost could be minimized here.  I don't think it is worth it right now though.  It is probably very minimal in nearly all cases.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1748,50 +1748,74 @@
...
+      ;; but this is presumably less significant than the cost of memory transitions.
+      ;;
+      ;; We perform the insertions and retractions in the same order as they were applied to the session since
+      ;; if a fact is not in the session, retracted, and then subsequently inserted it should be in the session at
+      ;; the end.
+      (doseq [{op-type :type facts :facts} pending-operations]
+
+        (condp = op-type
```
mrrodriguez: This would be faster as `case`
```clj
(case op-type
   :insertion
<etc>)
```

src/main/clojure/clara/rules/engine.cljc
WilliamParker: Yeah, it would be almost nonexistent unless you were, say, inserting/retractions tons of facts one at a time which will already perform poorly anyway.  It is basically just a keyword equality check.  That can add up if you do enough of them; we've had cases where doing (identical? some-arg :some-keyword) rather than (= some-arg :some-keyword) in a hotspot sped things up considerably.  I doubt this is a meaningful concern in this particular case though.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1748,50 +1748,74 @@
...
+
+            (binding [*pending-external-retractions* (atom [])]
+              ;; Bind the external retractions cache so that any logical retractions as a result
+              ;; of these insertions can be cached and executed as a batch instead of eagerly realizing
+              ;; them.  An external insertion of a fact that matches
+              ;; a negation or accumulator condition can cause logical retractions.
+              (doseq [[alpha-roots fact-group] (get-alphas-fn facts)
+                      root alpha-roots]
+                (alpha-activate root fact-group transient-memory transport transient-listener))
+              (external-retract-loop get-alphas-fn transient-memory transport transient-listener)))
```
mrrodriguez: Is there any reason not to do the `*pending-external-retractions*` along with the other given `:retraction`s?

Perhaps it'd just be too tricky to implement for the sake of this PR.

src/test/clojure/clara/test_accumulators.clj
mrrodriguez: I agree, the behavior has never been reliable.  I don't like adding more options and tricky semantics for things like this either really.  If it is any consolation, prior to Drools 6, the entire engine was very eager in computing LHS logic and accumulators.  When Drools 6 released, it went nearly the complete opposite direction and it was necessary to fire rules to observe almost any behavior.

Drools has a large user-base too.

I don't like making "breaking" changes, but this is really undocumented and unreliable behavior prior to these changes.  If anything, the documentation should make it clearly stated that prior to firing rules, the state of working memory is undefined.  The engine reserves this as an implementation detail to facilitate better runtime optimizations.

mrrodriguez:
+1

I'm certainly a big fan of this change.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: Well on the keyword equality thing, see https://github.com/cerner/clara-rules/pull/269#discussion_r102490288

`case` would alleviate that anyways.  It creates a constant time "jump table".

src/main/clojure/clara/rules/engine.cljc
WilliamParker: You are correct that I'm defending against a mutable collection here.  Currently if a user passes in a mutable collection and then mutates it later it doesn't matter since Clara will insert/retract the facts in it before returning from the insert/retract call.  Now that this is delayed mutation of the collection would change the behavior unless we specifically do something to be passive.  Note that we only do this if (coll? facts) returns false.  From the [docs](https://clojuredocs.org/clojure.core/coll_q) this should return true for any Clojure persistent collection.  Basically, we'd only incur the copying costs if a user passes a non-Clojure collection; otherwise we'd just incur what [amounts to an instance check](https://github.com/clojure/clojure/blob/clojure-1.8.0/src/clj/clojure/core.clj#L5977).  Particularly since Clara has a [top-level Java API](https://github.com/cerner/clara-rules/blob/0.13.0/src/main/clojure/clara/rules/java.clj) being passive to Java usage seems desirable.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: Possibly, along with various other changes here to group things together.  However, we'd have to carefully evaluate the performance implications since you'd be altering how things propagate through the Rete network, potentially significantly.  We'd also need to assess the implications on unconditional operations of such grouping.  My intent was basically to preserve existing behavior for now apart from removing the memory persistent/transient transitions and any entirely new options added (such as the one discussed at https://github.com/cerner/clara-rules/issues/249).  I don't see any current decisions with regards to the API that depend on the outcome of such optimizations and thus I'd prefer to defer them until a later time if we do want them.

curtman16:
agreed +1
performance optimizations

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: Yeah, my concern was from the Java side that this could end up being excessive deep-copying.  Maybe it doesn't matter though.

WilliamParker:
I have updated the pull request to

- Use case instead of condp for dispatch on the :retraction and :insertion keywords in fire-rules.
- Have some more assertions in the new test for this as @curtman16 suggested.
- Have some additional code comments, particularly around the copying of the collection of facts passed.

As regards https://github.com/cerner/clara-rules/pull/269#discussion_r102289462 it looks there is some other cleanup to do in that file e.g. removing duplication.  It also probably makes sense to have that be in a different commit since it is logically unrelated to this change.  I thus don't think it should block this change although I'd like to address it in the near future.

WilliamParker:
Also, I discussed this off-thread with @rbrush ; as of then he was pretty busy and hadn't had time to look at the details here but was OK with the general idea of the change and the requirement that users call fire-rules for querying to work.

mrrodriguez:
Looks good.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: I commented below, but I have updated this to use case.

PR #270, opened 2017-02-23
Fix malformed test and add intellij files to .gitignore

Zylox:
Basically this test was passing because it was throwing an arity exception, not because of what it was testing.

Interestingly, cursive found this for me, which was neat.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ pom.xml.asc
 .lein-repl-history
 figwheel_server.log
 *.*~
+.idea/*
+*.iml
--- a/src/test/clojure/clara/test_rules.clj
+++ b/src/test/clojure/clara/test_rules.clj
@@ -3627,9 +3627,9 @@
                                  (fire-rules)))
 
         run-session-untraced (fn []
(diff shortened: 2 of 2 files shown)


WilliamParker:
To elaborate a bit, when we threw an arity exception we then looked for listeners on the ExceptionInfo data map using [ex-data](https://clojuredocs.org/clojure.core/ex-data), which returns nil when the exception doesn't implement IExceptionInfo.  We then validate that there are no listeners in the data map, and when this data map is nil that check passes.  The assertion that there are listeners would fail if a similar mistake in run-session-traced but in the case where there are no listeners as @Zylox says the test will pass when another exception is thrown.

WilliamParker:
Also another minor point @Zylox : In this case I don't think it matters since the .gitignore has no functional impact and the change is very minor anyway but in general we try to have commits reflect logically distinct changes.  If you were, say, updating the project.clj for something to do with IntelliJ we'd probably want that in a separate commit.

Zylox:
Fair enough, i realized the distinction not being there this morning after i  had put the commit up. I'll keep that in mind in the future.

PR #271, opened 2017-02-28
Issue 249: Introduce experimental support for insertions and retracti…

WilliamParker:
…ons of equal facts to cancel each other, preventing further work in the Rete rules network

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -6,7 +6,9 @@
             [clojure.string :as string]
             [clara.rules.memory :as mem]
             [clara.rules.listener :as l]
-            [clara.rules.platform :as platform]))
+            [clara.rules.platform :as platform]
+            [clara.rules.update-cache.core :as uc]
+            #?(:clj [clara.rules.update-cache.cancelling :as ca])))
 
 ;; The accumulator is a Rete extension to run an accumulation (such as sum, average, or similar operation)
 ;; over a collection of values passing through the Rete network. This object defines the behavior
@@ -72,14 +74,19 @@
 ;; Returns a new session with the additional facts inserted.
 (defprotocol ISession
 
-  ;; Inserts a fact.
-  (insert [session fact])
+  ;; Inserts facts.
+  (insert [session facts])
 
-  ;; Retracts a fact.
-  (retract [session fact])
+  ;; Retracts facts.
+  (retract [session facts])
 
   ;; Fires pending rules and returns a new session where they are in a fired state.
-  (fire-rules [session])
+  ;;
+  ;; Note that clara.rules/fire-rules, the public API for these methods, will handle
+  ;; calling the two-arg fire-rules with an empty map itself, but we add handle it in the fire-rules implementation
+  ;; as well in case anyone is directly calling the fire-rules protocol function or interface method on the LocalSession.
+  ;; The two-argument version of fire-rules was added for issue 249.
+  (fire-rules [session] [session opts])
 
   ;; Runs a query agains thte session.
   (query [session query params])
@@ -163,7 +170,6 @@
   (alpha-activate [node facts memory transport listener])
   (alpha-retract [node facts memory transport listener]))
 
-
 ;; Protocol for getting the type (e.g. :production and :query) and name of a
 ;; terminal node.
 (defprotocol ITerminalNode
@@ -191,10 +197,6 @@
          :children
          (map get-conditions-and-rule-names)
          (reduce (partial merge-with conj) {})))))
-
-
-;; Record indicating pending insertion or removal of a sequence of facts.
-(defrecord PendingUpdate [type facts])
 
 ;; Active session during rule execution.
 (def ^:dynamic *current-session* nil)
@@ -234,15 +236,12 @@
   [current-session]
   (letfn [(flush-all [current-session flushed-items?]
             (let [{:keys [rulebase transient-memory transport insertions get-alphas-fn listener]} current-session
-                  pending-updates @(:pending-updates current-session)]
-
-              ;; Remove the facts here so they are re-inserted if we flush recursively.
-              (reset! (:pending-updates current-session) [])
+                  pending-updates (-> current-session :pending-updates uc/get-updates-and-reset!)]
 
               (if (empty? pending-updates)
                 flushed-items?
                 (do
-                  (doseq [partition (partition-by :type pending-updates)
+                  (doseq [partition pending-updates
                           :let [facts (mapcat :facts partition)]
                           [alpha-roots fact-group] (get-alphas-fn facts)
                           root alpha-roots]
@@ -251,7 +250,7 @@
                       (alpha-activate root fact-group transient-memory transport listener)
                       (alpha-retract root fact-group transient-memory transport listener)))
 
-                  ;; There may be new :pending-updates due to the flush just
+                  ;; There may be new pending updates due to the flush just
                   ;; made.  So keep flushing until there are none left.  Items
                   ;; were flushed though, so flush-items? is now true.
                   (flush-all current-session true)))))]
@@ -307,12 +306,12 @@
         (mem/add-insertions! transient-memory node token facts)
         (l/insert-facts-logical! listener node token facts)))
 
-    (swap! (:pending-updates *current-session*) into [(->PendingUpdate :insert facts)])))
+    (-> *current-session* :pending-updates (uc/add-insertions! facts))))
 
 (defn retract-facts!
   "Perform the fact retraction."
   [facts]
-  (swap! (:pending-updates *current-session*) into [(->PendingUpdate :retract facts)]))
+  (-> *current-session* :pending-updates (uc/add-retractions! facts)))
 
 ;; Record for the production node in the Rete network.
 (defrecord ProductionNode [id production rhs]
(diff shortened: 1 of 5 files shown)


rbrush:
Looks good! 

The only pitfall that comes to mind is that users could see insertions and retractions propagate and fire in a different order than how they inserted them, which would could break rules with side effects, or unconditional assert/retract logic (as you point out already). A rule that fires missiles multiple times based on a collection of insertions followed by retractions may not fire at all if they are all collapsed together, or if some other fact is seen first a side effect may be taken that affects a subsequent side effect.

We might strengthen the disclaimer, e.g. "Do not use side effects or unconditional inserts or retracts in this instance." It's also tempting to actually enforce this as much as we can, e.g. throw a descriptive exception if unconditional inserts or retracts occur when this is set. (Of course, we don't have a good way to detect side effects.)

rbrush:
Overall I'm +1 to merge this in, just floated the above ideas as optional improvements we can kick around.

mrrodriguez:
> The only pitfall that comes to mind is that users could see insertions and retractions propagate and fire in a different order than how they inserted them, which would could break rules with side effects, or unconditional assert/retract logic (as you point out already). A rule that fires missiles multiple times based on a collection of insertions followed by retractions may not fire at all if they are all collapsed together, or if some other fact is seen first a side effect may be taken that affects a subsequent side effect.

This change seems to be about a logical truth-maintenance thing to me.

rbrush:
@mrrodriguez Agreed. My point is simply that this creates unpredictability even if users aren't using truth maintenance since items can be cancelled or re-ordered...and it changes the signature of the current behavior even with truth maintenance. But it's certainly worth doing even so.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1784,52 +1778,100 @@
...
+                                 {:session session :opts opts}))
+
+           :clj (do (let [insertions (into []
+                                           (comp (filter (fn [pending-op]
+                                                           (= (:type pending-op)
+                                                              :insertion)))
+                                                 (map :facts)
+                                                 cat)
+                                           pending-operations)
+                          retractions (into []
```
mrrodriguez: Minor: when building "throw away" collections that need a transformation like this, building a vector is heavier than necessary .  `sequence` is a good alternative here.

src/test/clojure/clara/test_rules.clj
```diff
@@ -27,7 +29,21 @@
            LinkedList
            ArrayList]))
 
-(use-fixtures :once schema.test/validate-schemas)
+(defn opts-fixture
+  ;; For operations other than replace-facts uc/get-ordered-update-cache is currently
```
mrrodriguez: Should you instead run all tests here twice, one for each backing cache? A :once fixture makes that easy enough.

mrrodriguez: Whoops. You are doing this. Ignore. Can't get it to remove on mobile version.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -72,14 +74,14 @@
...
 (defprotocol ISession
 
-  ;; Inserts a fact.
-  (insert [session fact])
+  ;; Inserts facts.
+  (insert [session facts])
 
-  ;; Retracts a fact.
+  ;; Retracts facts.
   (retract [session fact])
```
WilliamParker: Yes, will change.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -72,14 +74,14 @@
...
 
   ;; Fires pending rules and returns a new session where they are in a fired state.
-  (fire-rules [session])
+  (fire-rules [session opts])
```
WilliamParker: I did that in the fire-rules function in rules.cljc.  I've always thought of that as the public API for this.

````
rules.cljc
(defn fire-rules
  "in real version the doc is here."
  ([session] (eng/fire-rules session {}))
  ([session opts] (eng/fire-rules session opts)))
````

File: src/main/clojure/clara/rules/engine.cljc [hunk 5]
```diff
@@ -1784,52 +1778,100 @@
...
+
+                      (binding [*current-session* {:rulebase rulebase
+                                                   :transient-memory transient-memory
+                                                   :transport transport
+                                                   :insertions (atom 0)
+                                                   :get-alphas-fn get-alphas-fn
+                                                   :pending-updates update-cache
+                                                   :listener transient-listener}]
+
+                        (doseq [[alpha-roots fact-group] (get-alphas-fn retractions)
```
WilliamParker: I'm not sure why you think this.  The current-session is bound here so left-retract on ProductionNode will place retractions in the pending updates.  The cancelling cache is symmetric in its cancellations - insertions can cancel retractions and vice versa.  Part of the relevant code:

````
  (add-insertions! [this facts]
    (let [fact-iter (.iterator ^Iterable facts)]
      (loop []
        (when (.hasNext fact-iter)
          (let [fact (.next fact-iter)]
            (when-not (dec-fact-count! retractions fact)
              (inc-fact-count! insertions fact))
            (recur))))))

  (add-retractions! [this facts]
    (let [fact-iter (.iterator ^Iterable facts)]
      (loop []
        (when (.hasNext fact-iter)
          (let [fact (.next fact-iter)]
            (when-not (dec-fact-count! insertions fact)
              (inc-fact-count! retractions fact))
            (recur))))))
````

mrrodriguez: The alpha-retract will have a chain effect of removing partial join results from beta memory across the network. Those memory updates will not have visibility into the partial join results that are subsequently added by the insertions done here after retractions. 

I'm not talking about the eventual RHS inserts or retracts that may happen at the terminal nodes of the network.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: My thinking is that this alters the ISession interface though. However I guess an implement of would have to implement both stories anyways due to no default behaviors able to be assigned in CLJ.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: I'm still not clear on what your concerns are; could you give a more concrete example?  Obviously retracting will remove partial joins, but are you saying that you'd expect doing insertions first would end up doing less work overall?  Overall my goal here is to cut down on the work in the downstream Rete network, not eliminate the first level of work done in response to insertions and retractions.  Also, is this related to the optimization in AccumulateNode of delaying accumulation until there is a matching token?

WilliamParker:
I have addressed the comments on my last changeset and updated the PR.

- I have strengthened the warning to only use truth maintenance rather than unconditional operations with the :cancelling option as requested at https://github.com/cerner/clara-rules/pull/271#issuecomment-283716743  Unfortunately, actively enforcing that we aren't performing these operations inside a fire-rules call using the :cancelling option will some work to ensure that it doesn't degrade performance and so I'm inclined to leave enforcement of this for a future enhancement.
- I have addressed Mike's comment at https://github.com/cerner/clara-rules/pull/271#discussion_r104310853 by moving insertions before retractions in the :cancelling path.  I added a comment explaining why insertions need to come first and a test of the behavior.
- I updated the ISession protocol and LocalSession implementation to support a 1-argument version of fire-rules.  I had already done this in the public API in clara.rules, but some consumers might be using the protocol function directly and it is easy enough to be passive to such usage in this case.
- I addressed assorted minor style comments from Mike, most of which I agree with.  I didn't change to a seq as discussed at https://github.com/cerner/clara-rules/pull/271#discussion_r104310908 because I wanted to maintain ordering of the insertions relative to each other and of retractions relative to each other.  Clara isn't obligated to do this, and insertions have to be reordered relative to retractions for this functionality, but my immediate inclination is to avoid reordering when we can reasonably do so without causing other negative impacts.  Into will use transients to build the new vector and we're using transducers we shouldn't build any intermediate vectors.

@mrrodriguez, could you confirm that I've addressed your concerns that you expressed in https://github.com/cerner/clara-rules/pull/271#discussion_r104310853 ?  Otherwise the changes are relatively straightforward.

mrrodriguez:
>I have addressed Mike's comment at #271 (comment) by moving insertions before retractions in the :cancelling path. I added a comment explaining why insertions need to come first and a test of the behavior.

Looks good and good comment.

>I updated the ISession protocol and LocalSession implementation to support a 1-argument version of fire-rules. I had already done this in the public API in clara.rules, but some consumers might be using the protocol function directly and it is easy enough to be passive to such usage in this case.

Looks good.

>I didn't change to a seq as discussed at #271 (comment) because I wanted to maintain ordering of the insertions relative to each other and of retractions relative to each other.
This is concerning the line in this change https://github.com/cerner/clara-rules/pull/271/files#diff-00da05e66726186c05209662d990b97cR1849

`sequence` will not reorder anything.  It will not be the same as `(into () <etc>)` if that is what you are thinking.  `sequence` will just build a lazy seq destination collection for the "tranducible process" over the source collection.  So I'm not sure what your concerns were here.  

If later you were to `conj` to this data structure, then a vector would matter as far as appending to the correct ends.  This is why I called it a "throw away" collection.  You immediately pass it to `get-alphas-fn` (which has no expectations of its input being a vector) and disregard it beyond that (or at least just iterate over it).

So I just suggested it since `sequence` is typically the really cheap operation here, but it really isn't that important.  I just think it is an under-used transducer processing function, probably because it wasn't as popular as `into` was prior to transducers.

It doesn't bother me one way or another here though.  I just wanted to make it clear that I was not suggesting that we should be re-ordering `retractions` or `insertions` here.  I was being careful to not suggest doing that for the reasons you point out.

WilliamParker:
@mrrodriguez OK I understand now; I was indeed of thinking of the behavior of 

````
(into () ...)
````

I wasn't aware of the transducer arities of clojure.core/sequence.  I updated the PR to use sequence to build the insertions and retractions collections before merging.

Issue #267, opened 2017-02-17
Variable Binding Scope


JarrodCTaylor:
I'm attempting to understand how variable binding works within rules, and experiencing some unexpected behavior. For example in the code below `?b` disappears when printed with `?a`

```
(require '[clara.rules :as rule])

(rule/defrule wat
  [1 (= ?a "a")]
  [1 (= ?b "b")
   (nil? (println ?a))    ;; => "a"
   (nil? (println ?b))    ;; => "b"
   (nil? (println ?a ?b)) ;; => "a nil"
   ] => nil)

(-> (rule/mk-session [wat] :fact-type-fn (constantly 1))
    (rule/insert 1)
    (rule/fire-rules))
```

rbrush:
Thanks for reporting this.

This looks like a bug for a the specific case where if a variable is bound in a constraint, and then later used in that constraint in such a way that the compiler thinks the bound variable is used in a join that cannot be hashed joined, it's not set in the resulting code that gets compiled.

This is something of an unusual case, since usually we don't reuse a variable bound in a given constraint later in that constraint (since that variables source is visible there). However, we should still support it because this behavior is definitely surprising. 

Here's the quick debugging I did to see what's going on here. The pertinent area is the :join-filter-expression in node 2.  Notice the (println ?a ?b) was moved to the join-filter-expression (since it may be attempting a join between ?a and ?b), whereas the simple (printlin ?b) remains in the original expression (since there is no other variable in there to make the compiler think it's joining to something else.)

Anyway, it seems like we'll have to dig into the logic for that non-hash join node to make sure previously bound variables are visible to the join operation as well.

```clj
(-> [wat] clara.rules.compiler/to-beta-graph clojure.pprint/pprint)

{:forward-edges {0 #{1}, 1 #{2}, 2 #{3}},
 :backward-edges {1 #{0}, 2 #{1}, 3 #{2}},
 :id-to-condition-node
 {0 :clara.rules.compiler/root-condition,
  1
  {:node-type :join,
   :condition {:type 1, :constraints [(= ?a "a")]},
   :new-bindings #{:?a},
   :used-bindings #{:?a},
   :join-bindings #{}},
  2
  {:node-type :join,
   :condition
   {:type 1,
    :constraints [(= ?b "b") (nil? (println ?b))],
    :original-constraints
    [(= ?b "b")
     (nil? (println ?a))
     (nil? (println ?b))
     (nil? (println ?a ?b))]},
[...]

PR #273, opened 2017-03-11
Issue #267: local bindings not visible if used in a non-hash join.

rbrush:
Fix for #267, ensuring a bound variable later used in a join binding is properly visible there. 

A side note: my editor cleaned up some extraneous whitespace when making this change, which normally I'd strip out and do in another commit. But I don't think it's terribly distracting and we should clean it up anyway, so I've left the whitespace cleanup in this commit as well.

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -402,12 +402,16 @@
 
 (defn compile-join-filter
   "Compiles to a predicate function that ensures the given items can be unified. Returns a ready-to-eval
-   function that accepts a token, a fact, and an environment, and returns truthy if the given fact satisfies
-   the criteria."
-  [{:keys [type constraints args] :as unification-condition} ancestor-bindings env]
-  (let [accessors (field-name->accessors-used type constraints)
+   function that accepts the following:
 
-        binding-keys (variables-as-keywords constraints)
+   * a token from the parent node
+   * the fact
+   * a map of bindings from the fact, which was typically computed on the alpha side
+   * an environment
+
+   The function created here returns truthy if the given fact satisfies the criteria."
+  [{:keys [type constraints args] :as unification-condition} ancestor-bindings element-bindings env]
+  (let [accessors (field-name->accessors-used type constraints)
 
         destructured-env (if (> (count env) 0)
                            {:keys (mapv #(symbol (name %)) (keys env))}
@@ -424,18 +428,31 @@
                                              [name (list accessor '?__fact__)])
                                            accessors)))
 
-        token-assignments (mapcat #(list (symbol (name %)) (list 'get-in '?__token__ [:bindings %])) binding-keys)
+        ;; Get the bindings used in the join filter expression that are pulled from
+        ;; the token. This is simply the bindings in the constraints with the newly
+        ;; created element bindings for this condition removed.
+        token-binding-keys (remove element-bindings (variables-as-keywords constraints))
+
+        token-assignments (mapcat #(list (symbol (name %))
+                                         (list 'get-in '?__token__ [:bindings %]))
+                                  token-binding-keys)
+
+        new-binding-assignments (mapcat #(list (symbol (name %))
+                                               (list 'get '?__element-bindings__ %))
+                                        element-bindings)
 
         assignments (concat
                      fact-assignments
-                     token-assignments)
+                     token-assignments
+                     new-binding-assignments)
 
         equality-only-variables (into #{} (for [binding ancestor-bindings]
                                             (symbol (name (keyword binding)))))]
 
     `(fn [~'?__token__
          ~(add-meta '?__fact__ type)
-          ~destructured-env]
+         ~'?__element-bindings__
+         ~destructured-env]
        (let [~@assignments
              ~'?__bindings__ (atom {})]
          ~(compile-constraints constraints equality-only-variables)))))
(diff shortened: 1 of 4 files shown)


rbrush:
Minor update: This same problem could occur with accumulators as well, so I added a test for that too.

src/test/clojure/clara/test_rules.clj
```diff
@@ -5452,5 +5452,25 @@
                     (insert (->Cold -10))
                     (fire-rules {:cancelling true})
                     (query cold-query))))))
-    
-  
+
+;; Test for https://github.com/cerner/clara-rules/issues/267
+(deftest local-scope-visible-in-join-filter
```
WilliamParker: Nitpick: Almost all of the other tests in the file have names prefixed by "test" so we should probably make this "test-local-scope-visible-in-join-filter".

WilliamParker:
@rbrush I've looked over this; I had some suggestions for improvements but I basically agree with the changes.

On the whitespace issue, IMO ideally we'd have some sort of automated editor-agnostic way of taking care of such issues.  I haven't tracked down where this whitespace was inserted, but I've seen different editors format Clojure differently.

mrrodriguez:
I think Will's comments cover anything I'd add.  I did look over it though and it looks good to me.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -646,8 +646,8 @@
       (into [type] constraints))))
 
 (defn- join-node-matches
-  [node join-filter-fn token fact env]
-  (let [beta-bindings (try (join-filter-fn token fact {})
+  [node join-filter-fn token fact fact-bindings env]
+  (let [beta-bindings (try (join-filter-fn token fact fact-bindings {})
                            (catch #?(:clj Exception :cljs :default) e
                                (throw-condition-exception {:cause e
```
rbrush: Makes sense. I'm more tempted to go with token-bindings and fact-bindings here so we keep the distinction. 

Technically this could break a user explicitly catching an exception and checking the :bindings field, but I think that's low enough risk that it's worth using more clear names.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -424,18 +422,29 @@
                                              [name (list accessor '?__fact__)])
                                            accessors)))
 
-        token-assignments (mapcat #(list (symbol (name %)) (list 'get-in '?__token__ [:bindings %])) binding-keys)
+        ;; Binding keys that are not newly created are drawn from the token.
+        token-binding-keys (remove element-bindings (variables-as-keywords constraints))
```
rbrush: Yeah, we could explicitly compute this and add it as another field to the beta graph. If doing so makes things easier to debug (i.e., the small additional complexity added to the beta graph is justified), then it's worth pursuing, but I agree that is out of scope here.

rbrush: In either case, I'll expand the comments here a bit to clarify what it's doing.

src/test/clojure/clara/test_rules.clj
```diff
@@ -5452,5 +5452,25 @@
                     (insert (->Cold -10))
                     (fire-rules {:cancelling true})
                     (query cold-query))))))
-    
-  
+
```
rbrush: I'm adding a test with the following structure, which I think is what you're looking for.

```clj
        check-reuse-previous-binding (dsl/parse-query [] [[WindSpeed (= ?t windspeed) (= ?w windspeed)]
                                                          [Temperature (= ?t temperature)
                                                           (join-filter-equals ?w ?t 10)]])
```

src/main/clojure/clara/rules/engine.cljc
rbrush: After taking this for a spin (and looking at the unit tests that check for :bindings that broke), I decided to reverse direction and just merge into the existing :bindings field. That preserves the existing behavior in most cases, keeps the structure a bit simpler, and users can easily see where the bindings came from by looking at the rule itself.

rbrush:
The PR is updated with suggestions incorporated, and a couple new tests added as well.

WilliamParker:
A test like this would be helpful IMO:

````
;;; query in let binding
        check-accum-result-previous-binding (dsl/parse-query []
                                                             [[?t <- (acc/min :temperature) :from [Temperature]]
                                                              [ColdAndWindy (= ?t temperature) (join-filter-equals ?t windspeed)]])
;;; assertions

(is (= [{:?t -10}]
           (-> (mk-session [check-accum-result-previous-binding])
               (insert (->Temperature -10 "MCI"))
               (insert (->ColdAndWindy -10 -10))
               fire-rules
               (query check-accum-result-previous-binding))))

    (is (empty? (-> (mk-session [check-accum-result-previous-binding])
                    (insert (->Temperature -10 "MCI"))
                    (insert (->ColdAndWindy -20 -20))
                    fire-rules
                    (query check-accum-result-previous-binding)))
        "Validate that the ?t binding from the previous accumulator result is used, rather
         than the binding in the ColdAndWindy condition that would create a ?t binding if one were
         not already present")
````
Basically the idea is to validate that a result binding is considered a "previous binding" still even when the later condition could create the binding itself.  I was specifically thinking about a binding from a result rather than a constraint in a previous condition since the paths for result bindings through the compiler are slightly different than those for variables from constraints; see [here](https://github.com/cerner/clara-rules/blob/0.13.0/src/main/clojure/clara/rules/compiler.clj#L1031) which is used to determine what bindings come from ancestors [here](https://github.com/cerner/clara-rules/blob/0.13.0/src/main/clojure/clara/rules/compiler.clj#L822).  This worked in my REPL against the branch; https://gist.github.com/WilliamParker/7e8fae758582fb52a63084f4f404f280 is a full Gist of the deftest body I used including existing cases.

Otherwise this looks good to me; +1 to merging.

rbrush:
Alright, I went ahead and added Will's suggested test as he posted it. :)

rbrush:
Unless there are objections, I'll merge this today since I've gotten +1's, modulo the last test I added.

WilliamParker:
+1 to merging this @rbrush

rbrush:
Merged! Thanks!

PR #279, opened 2017-03-31
Issue 276: Add the accumulated-over facts to rule activation explanat…

WilliamParker:
…ions in the session inspection functionality

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -118,6 +118,24 @@
   ;; Right-activate the node with items reduced in the above pre-reduce step.
   (right-activate-reduced [node join-bindings reduced  memory transport listener]))
 
+(defprotocol IAccumInspect
+  "This protocol is expected to be implemented on accumulator nodes in the rules network.
+   It is not expected that users will implement this protocol, and most likely will not call
+   the protocol function directly."
+  (token->matching-elements [node memory token]
+    "Takes a token that was previously propagated from the node, 
+     or a token that is a descendant of such a token, and returns the facts in elements 
+     matching the token propagated from the node.  During rules firing
+     accumulators only propagate bindings created and the result binding
+     downstream rather than all facts that were accumulated over, but there
+     are use-cases in session inspection where we want to retrieve the individual facts.
+    
+     Example: [?min-temp <- (acc/min :temperature) :from [Temperature (= temperature ?loc)]]
+              [?windspeed <- [WindSpeed (= location ?loc)]]
+     
+     Given a token propagated from the node for the WindSpeed condition 
+     we could retrieve the Temperature facts from the matching location."))
+
 ;; The transport protocol for sending and retracting items between nodes.
 (defprotocol ITransport
   (send-elements [transport memory listener nodes elements])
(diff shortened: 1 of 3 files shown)


src/main/clojure/clara/rules/engine.cljc
```diff
@@ -118,6 +118,14 @@
   ;; Right-activate the node with items reduced in the above pre-reduce step.
   (right-activate-reduced [node join-bindings reduced  memory transport listener]))
 
+(defprotocol IAccumInspect
+  ;; Takes a token that was previously propagated from the node
```
mrrodriguez: Small thing, but `defprotocol` supports doc strings on both the protocol and each of the functions defined along with it.

mrrodriguez:
Looks like a pretty good addition to me. +1

src/main/clojure/clara/rules/engine.cljc
rbrush: Also, am I correct in understanding this accepts a token from this node, or a descendent of that token, and returns the corresponding elements? It might be worth making it more explicit what exactly the input and output of this would be.

src/main/clojure/clara/tools/inspect.clj
```diff
@@ -15,8 +15,10 @@
 ;; This is derived from the Rete-style tokens, but this token
 ;; is designed to propagate all context needed to easily inspect
 ;; the state of rules.
-(s/defrecord Explanation [matches :- [[(s/one s/Any "fact") (s/one schema/Condition "condition")]] ; Fact, condition tuples
-                           bindings :- {s/Keyword s/Any}]) ; Bound variables
+(s/defrecord Explanation [matches :- [{:result s/Any
```
rbrush: Is ```:result``` the best field name here? This could be an arbitrary fact from a non-accumulator node (rather than an accumulator result)...so perhaps we just call it ```:fact```? 

Either way, I think a more complete definition of what each of these fields could contained would be helpful to users.

rbrush:
Very nice, Will! My only comments were for more clear definition of the structures and the updated Explanation structure and the IAccumInspect function in the code. Otherwise, +1.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: "Also, am I correct in understanding this accepts a token from this node, or a descendent of that token, and returns the corresponding elements?"

That is mostly correct.  It returns the matching facts, not [Element records](https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/rules/engine.cljc#L26).

I'll try to clear this up; let me know if you have further suggestions.  I made some comments in the implementations on AccumulateNode and AccumulateWithJoinFilterNode that could be helpful in understanding this as well.

src/main/clojure/clara/tools/inspect.clj
WilliamParker: Unfortunately, I don't see a name that really fits well with both the accumulator and non-accumulator cases.  It is probably better to go with a name that fits the majority non-accumulator case better.  I'll change this to :fact and elaborate more in documentation on the schema.

WilliamParker:
I've addressed the comments on the previous iteration of this and updated the PR.

rbrush:
Looks good to me! +1 to merge.

PR #281, opened 2017-04-07
Issue 280:

WilliamParker:
- In clara.tools.inspect, return empty collections of matches when no match is present for a rule, query, or condition.
    In some cases we previously returned a collection with elements that were either nil or an explanation record with nothing in it.
  - Fix memory leak where bindings from facts that are retracted were held by the memory.


rbrush:
Looks good to me.

PR #283, opened 2017-04-13
Issue 277: Added a function in tools to create a directed graph relat…

WilliamParker:
…ing facts to the logical insertions they trigger

### DIFF
--- a/src/main/clojure/clara/tools/fact_graph.clj
+++ b/src/main/clojure/clara/tools/fact_graph.clj
@@ -1,0 +1,96 @@
+(ns clara.tools.fact-graph
+  (:require [clara.tools.inspect :as i]
+            [schema.core :as sc]))
+
+;; This node will have either facts or results of accumulations as its parents.
+;; Its children will be facts that the rule inserted.
+(sc/defrecord RuleActivationNode [rule-name :- sc/Str
+                                  id :- sc/Int])
+
+;; The parents of this node are facts over which an accumulation was run.
+;; It will have a single child, the result of the accumulation.  So, for example,
+;; with the condition [?t <- (acc/min :temperature) :from [Temperature]], if we have
+;; (->Temperature 50 "MCI") and (->Temperature 30 "MCI") the child of this node will be
+;; an AccumulationResult with the :result 30 and the parents will be the two Temperature facts.
+(sc/defrecord AccumulationNode [id :- sc/Int])
+
+;; As alluded to above, this node represents the result of an accumulation.  Its child will be a
+;; RuleActivationNode.  Note that there will be an AccumulationResult for each distinct rules firing.
+(sc/defrecord AccumulationResultNode [id :- sc/Int
+                                      result :- sc/Any])
+
+(def ^:private empty-fact-graph {:forward-edges {}
+                                 :backward-edges {}})
+
+(defn ^:private add-edge [graph from to]
+  (-> graph
+      (update-in [:forward-edges from] (fnil conj #{}) to)
+      (update-in [:backward-edges to] (fnil conj #{}) from)))
+
+(defn ^:private add-insertion-to-graph
+  [original-graph id-counter fact-inserted {:keys [rule-name explanation]}]
+  (let [facts-direct (sequence
+                      (comp (remove :facts-accumulated)
+                            (map :fact))
+                      (:matches explanation))
+        
+        activation-node (map->RuleActivationNode {:rule-name rule-name
+                                                  :id (swap! id-counter inc)})
+
+        accum-matches (filter :facts-accumulated (:matches explanation))]
+    
+    (as-> original-graph graph
+      (if (seq accum-matches)
+        (reduce (fn [reduce-graph accum-match]
+                  (let [accum-node (->AccumulationNode (swap! id-counter inc))
+                        accum-result (->AccumulationResultNode (swap! id-counter inc) (:fact accum-match))]
+                    (as-> reduce-graph g
+                      ;; Add edges from individual facts to an AccumulationResultNode.
+                      (reduce (fn [g accum-element]
+                                (add-edge g accum-element accum-node))
+                              g (:facts-accumulated accum-match))
+                      (add-edge g accum-node accum-result)
+                      (add-edge g accum-result activation-node))))
+                graph
+                accum-matches)
+        graph)
+      ;; Add edges to the rule activation node from the facts that contributed
+      ;; to the rule firing that were not in accumulator condition.
+      (reduce (fn [g f]
+                (add-edge g f activation-node))
+              graph
+              facts-direct)
+      (add-edge graph activation-node fact-inserted))))
+
+(defn session->fact-graph
+  "Given a session, return a graph structure connecting all facts to the facts
+   that they caused to be logically inserted.  Note that such connections will not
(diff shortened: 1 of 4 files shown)


src/main/clojure/clara/tools/fact_graph.clj
```diff
@@ -0,0 +1,90 @@
...
+(sc/defrecord AccumulationResultNode [id :- sc/Int
+                                      result :- sc/Any])
+
+(def ^:private empty-fact-graph {:forward-edges {}
+                                 :backward-edges {}})
+
+;; Use a counter, whose value will be added to internal nodes, to the ensure that
+;; these nodes are not equal to each other.  This ensures that the number of the internal
+;; nodes accurately reflects the cardinality of each fact in the session.
+(def ^:private ^:dynamic *id-counter* nil)
```
rbrush: Rather than using a dynamic variable, could the id-counter just be an atom created in session->fact-graph function and passed to its children? (Just a preference for explicitly passed parameters over dynamic scope when possible to make it easier to trace the logic through.)

src/main/clojure/clara/tools/fact_graph.clj
```diff
@@ -0,0 +1,90 @@
...
+              facts-direct)
+      (add-edge graph activation-node fact-inserted))))
+
+(defn session->fact-graph
+  "Given a session, return a graph structure connecting all facts to the facts
+   that they caused to be logically inserted.  Note that such connections will not
+   be made for unconditionally inserted facts."
+  [session]
+  (binding [*id-counter* (atom 0)]
+    (let [fact->explanations (@#'i/gen-fact->explanations session)
```
rbrush: So this is accessed via a var because gen-fact->explanations is private? I'm not sure there any harm in making that function public (and removing this workaround), but am open to discussion on the reasoning here.

src/main/clojure/clara/tools/fact_graph.clj
```diff
@@ -0,0 +1,90 @@
...
+          insertion-tuples (into []
+                                 (comp
+                                  (map (fn [[k v]]
+                                         (map (fn [v-entry]
+                                                [k (update v-entry
+                                                           :rule
+                                                           :name)])
+                                              v)))
+                                  cat)
+                                 fact->explanations)]
```
rbrush: Not a blocker, but a little explanation of this composed function with nested maps wouldn't hurt, or perhaps more descriptive variable names in the nested functions. I had to read it a couple times to untangle the logic here.

rbrush:
Made a few comments, but looks good overall!

src/main/clojure/clara/tools/fact_graph.clj
WilliamParker: +1, id-counter is only being used in add-insertion-to-graph so it is just as easy to use an atom argument and thereby be more explicit.

src/main/clojure/clara/tools/fact_graph.clj
WilliamParker: Correct.  My goal was to avoid complicating the existing inspection API by adding a new entry point, and particularly making one of the entries on the map returned by inspect different from the others (i.e. having its own function).  I considered adding this graphing functionality to the inspection namespace, but subjectively it seems different in intent from the existing inspection functionality and might be expensive to create. 

My thinking here is that the different entries in the session inspection map should each have their own function, like :fact->explanations would if gen-fact->explanations were made public, or that they should all be accessed only through the inspect function.  Basically I'd just like to be consistent; I don't have any problem with having a "single purpose inspection function" API, although we'd probably want to leave the inspection function as well for passivity.

src/main/clojure/clara/tools/fact_graph.clj
WilliamParker: I'll see what I can do to untangle this; I agree that clarity is wanting here.

src/main/clojure/clara/tools/fact_graph.clj
rbrush: Fair enough. In that case I might suggest logging a bug to make each entry in session inspection a public function so we can track on that change, and add a comment here linking to it so anyone reading this code will know the reasoning here. (Just using the general heuristic of using comments to explain chunks of code where the reasoning isn't immediately clear, which I think applies to the indirect use of a var here.)

curtman16:
+1
I had no specific comments

src/main/clojure/clara/tools/fact_graph.clj
```diff
@@ -0,0 +1,90 @@
...
+
+(def ^:private empty-fact-graph {:forward-edges {}
+                                 :backward-edges {}})
+
+;; Use a counter, whose value will be added to internal nodes, to the ensure that
+;; these nodes are not equal to each other.  This ensures that the number of the internal
+;; nodes accurately reflects the cardinality of each fact in the session.
+(def ^:private ^:dynamic *id-counter* nil)
+
+(defn ^:private add-edge [graph a b]
```
mrrodriguez: minor but, wouldn't `from` and `to` be better names than `a` and `b`, respectively?
I think that'd make it it more immediately clear the intent.

src/main/clojure/clara/tools/fact_graph.clj
mrrodriguez: Just as another note on this:
If the conflict here is that you want to have more than one "public API namespace" to make it more logical for consumers or just conceptually easier to parse through, but you want these public API namespaces to share some internal impl details with each other, some libraries just make a namespace with something like "impl" or "internal" in it to hold these common functions. This is similarly done sometimes in Java codebases as well, etc.
So like `clara.impl.inspect` or `clara.tools.impl.inspect`, etc.

Probably not worth it for one var though. I think commented on the use of an "internal" private var across namespaces is a good idea though, like Ryan said above.

WilliamParker: I'll add a comment and log an issue as Ryan suggests.  Creating an "impl" namespace would probably be overkill for a single relatively small function.

WilliamParker:
I have updated the PR in response to the comments.  The changes were pretty minor and didn't impact any existing functionality, but I'll want to take another look at this before merging.

src/main/clojure/clara/tools/fact_graph.clj
```diff
@@ -0,0 +1,96 @@
...
+
+;; As alluded to above, this node represents the result of an accumulation.  Its child will be a
+;; RuleActivationNode.  Note that there will be an AccumulationResult for each distinct rules firing.
+(sc/defrecord AccumulationResultNode [id :- sc/Int
+                                      result :- sc/Any])
+
+(def ^:private empty-fact-graph {:forward-edges {}
+                                 :backward-edges {}})
+
+(defn ^:private add-edge [graph to from]
```
WilliamParker: It looks like I reversed the names on "to" and "from" but used them correctly so the tests still passed; will fix.

WilliamParker: I have fixed this.

WilliamParker:
The changes after the first rounds of reviews on this were relatively minor and to lower-risk areas of the codebase, so I'm going to go ahead and merge this.  I don't think we're planning to release immediately anyway so if anyone has concerns we can just address them before a release.

Issue #285, opened 2017-04-22
Activation groups in Clojurescript?


alex-dixon:
Confused about whether `activation-group-fn` and `activation-sort-fn` has been implemented yet in CLJS.
 
I came across this in https://github.com/cerner/clara-rules/blob/master/src/main/clojure/clara/rules.cljc#L248:
```
    ;; ClojureScript implementation doesn't support salience yet, so
    ;; no activation group functions are used.
```
However, a few lines down, vars for `activation-group-fn` and `activation-sort-fn` are both passed to `(eng/local-memory)` inside the `LocalSession` being assembled. 

I'm fairly confident I've witnessed rules marked with `{salience: -100}` fire last in CLJS. Not positive but pretty sure.

Also saw this comment from Ryan https://github.com/cerner/clara-rules/issues/174#issuecomment-196838416 about the above comment being potentially out of data.

alex-dixon:
Went ahead and tested `activation-group-fn` and `activation-group-sort-fn` in CLJS and both appear to be working. 😃

WilliamParker:
Yeah, it looks like the comment is out date. 

- The memory uses a sorted map on the activation group sort fn for giving the next rule activation to fire.  See [to-transient](https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/rules/memory.cljc#L1001) and [pop-activation!]( https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/rules/memory.cljc#L781), among others.
- [ClojureScript sessions wrap](https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/rules.cljc#L240) the activation-group-fn and activation-group-sort-fn in the same way as [Clojure sessions | https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/rules/compiler.clj#L1714].

Assuming I'm not missing something, we should add some simple test case(s) validating that salience works in ClojureScript if we don't have them already and remove the comment.

alex-dixon:
Just tried to get a CLJS repl going for a couple hours but don't seem to be getting very far. I see `austin` and `piggieback` referenced in `test_rules.cljs` and removed the comment but they don't seem to be part of the project's dependencies. Should I be trying to add `austin` and/or `piggieback` dependencies to my lein profile? Any help would be greatly appreciated. Would love to contribute the tests for CLJS.

WilliamParker:
[...]

PR #288, opened 2017-05-02
Add salience test CLJS

alex-dixon:
Copied salience test from `test-rules.clj` and modified for CLJS. Addresses #285 regarding whether salience and activation groups work in CLJS as they do in CLJ.

### DIFF
--- a/src/test/clojurescript/clara/test.cljs
+++ b/src/test/clojurescript/clara/test.cljs
@@ -2,6 +2,7 @@
   (:require-macros [cljs.test :as test])
   (:require [clara.test-rules]
             [cljs.test]
+            [clara.test-salience]
             [clara.test-common]))
 
 (enable-console-print!)
@@ -12,4 +13,4 @@
     (println "FAIL")))
 
 (defn ^:export run []
-  (test/run-tests 'clara.test-rules 'clara.test-common))
+  (test/run-tests 'clara.test-rules 'clara.test-common 'clara.test-salience))
--- a/src/test/clojurescript/clara/test_salience.cljs
+++ b/src/test/clojurescript/clara/test_salience.cljs
@@ -1,0 +1,115 @@
+(ns clara.test-salience
+  (:require-macros [cljs.test :refer (is deftest run-tests testing)]
+    [clara.test-rules-data])
+  (:require [cljs.test :as t]
+    [clara.rules.engine :as eng]
+    [clara.rules.accumulators :as acc]
+    [clara.rules :refer [insert retract fire-rules query insert!]
+     :refer-macros [defrule defsession defquery]]
+    [clara.rules.testfacts :refer [->Temperature Temperature
+                                   ->WindSpeed WindSpeed
+                                   ->ColdAndWindy ColdAndWindy]]))
+
+(def salience-rule-output (atom []))
+
+(defrule salience-rule1
+  {:salience 100}
+  [Temperature]
+  =>
+  (swap! salience-rule-output conj 100))
+
+(defrule salience-rule2
+  {:salience 50}
+  [Temperature]
+  =>
+  (swap! salience-rule-output conj 50))
+
+(defrule salience-rule3
+  {:salience 0}
+  [Temperature]
+  =>
+  (swap! salience-rule-output conj 0))
+
+(defrule salience-rule4
+  {:salience -50}
+  [Temperature]
+  =>
+  (swap! salience-rule-output conj -50))
+
+
+(deftest test-salience
+  (doseq [[sort-fn
+           group-fn
+           expected-order]
+
+          [[:default-sort :default-group :forward-order]
+           [:default-sort :salience-group :forward-order]
+           [:default-sort :neg-salience-group :backward-order]
+
+           [:numeric-greatest-sort :default-group :forward-order]
+           [:numeric-greatest-sort :salience-group :forward-order]
+           [:numeric-greatest-sort :neg-salience-group :backward-order]
+
+
+           [:boolean-greatest-sort :default-group :forward-order]
+           [:boolean-greatest-sort :salience-group :forward-order]
+           [:boolean-greatest-sort :neg-salience-group :backward-order]
+
+
+           [:numeric-least-sort :default-group :backward-order]
+           [:numeric-least-sort :salience-group :backward-order]
+           [:numeric-least-sort :neg-salience-group :forward-order]
+
+           [:boolean-least-sort :default-group :backward-order]
+           [:boolean-least-sort :salience-group :backward-order]
+           [:boolean-least-sort :neg-salience-group :forward-order]]]
+
+    (let [numeric-greatest-sort (fn [x y]
+                                  (cond
+                                    (= x y) 0
+                                    (> x y) -1
+                                    :else 1))
+
+          numeric-least-sort (fn [x y]
+                               (numeric-greatest-sort y x))
+
+          salience-group-fn (fn [production]
+                              (or (some-> production :props :salience)
+                                0))
+
+          neg-salience-group-fn (fn [p]
+                                  (- (salience-group-fn p)))]
+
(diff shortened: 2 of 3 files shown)


rbrush:
Looks good to me! 

Ultimately I think we're overdue for some significant refactoring of the Clara tests. We hadn't done as much in the CLJS space since our usage is Clojure on the JVM almost exclusively...but if we're going to pick up usage here it's probably worthwhile to refactor the larger test suite and make it so as many tests as possible are run on both Clojure and ClojureScript using some cljc trickery. 

But, that desire shouldn't prevent us from making things better in the mean time, so this change is worthwhile.

src/test/clojurescript/clara/test.cljs
```diff
@@ -12,4 +12,4 @@
     (println "FAIL")))
 
 (defn ^:export run []
-  (test/run-tests 'clara.test-rules 'clara.test-common))
+  (test/run-tests 'clara.test-rules 'clara.test-common 'clara.test-salience))
```
alex-dixon: I checked Travis and `test-salience` did run but I can see I broke the pattern here of requiring the test file at the top of the ns...

WilliamParker: I'm not too familiar with ClojureScript's compilation, but I think something analogous to Clojure's behavior where namespaces can be referenced with full qualification when they have already been compiled is happening here. I had to add the test-salience require to test.cljs for the build of this branch to pass on my local machine.  In any case it seems safest to explicitly require namespaces on which a namespace directly depends.

src/test/clojurescript/clara/test_salience.cljs
```diff
@@ -0,0 +1,115 @@
...
+                               (numeric-greatest-sort y x))
+
+          salience-group-fn (fn [production]
+                              (or (some-> production :props :salience)
+                                0))
+
+          neg-salience-group-fn (fn [p]
+                                  (- (salience-group-fn p)))]
+
+      (defsession test-salience-session 'clara.test-salience
```
WilliamParker: Defsession exists since rulebase compilation in ClojureScript has to happen at compile time, rather than at runtime.  Creating sessions at runtime in Clojure makes heavy use of eval, which was historically absent in ClojureScript.  I know there has been some work on adding eval to ClojureScript; it would be nice if that was mature enough to use but I haven't looked into it enough to assess that.

I think what's happening here is that the actual rule network construction is being done at compilation time, and the macro is working out so that the functionality controlled by the options, which isn't reliant on eval, gets done at runtime.  This seems to work, but is a little odd.  I'm not sure what we should do here, I'll have to give it some thought.  Thoughts @rbrush ?

src/test/clojurescript/clara/test.cljs
alex-dixon: @WilliamParker Fixed in c0ce9a9f22451ea50acded7e4c073943f8ce33ca

src/test/clojurescript/clara/test_salience.cljs
WilliamParker: On reflection, I think this is OK for the moment.  It is a bit odd, but I don't foresee the structure of the defsession macro changing in the immediate future in a way that would break this usage.

It does bring up the interesting point of whether we should expose a macro like

````
(mk-compile-time-session rule-sources :option option-value)
````

in ClojureScript that would behave like this without the requirement of defining a var, either as a test helper internal to Clara's tests or as a public API.  It wouldn't be too difficult to refactor the [defsession implementation](https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/macros.clj#L251) to share the logic in that way.  Alternatively, perhaps eval in ClojureScript has matured to the point that we could use it in our tests.  I don't think we should make decisions on these points in haste, so I'm in favor of leaving the defsession usage in these tests as is.   We should probably have an issue to address these points; I'll plan to log an issue (or you can) and then we can add a comment next to the defsession use here referencing that issue.

File: src/test/clojurescript/clara/test_salience.cljs [hunk 3]
```diff
@@ -0,0 +1,115 @@
...
+                                                 :boolean-greatest-sort  >
+                                                 :boolean-least-sort <)
+                     :activation-group-fn (condp = group-fn
+                                            :default-group nil
+                                            :salience-group salience-group-fn
+                                            :neg-salience-group neg-salience-group-fn))
+
+      ;; Ensure the rule output reflects the salience-defined order.
+      ;; independently of the order of the rules.
+      (dotimes [n 10]
```
WilliamParker: I don't see any variation of the rule order happening here, as suggested by the comment above.  I don't think varying the order is really necessary though since you have tests where the ordering is expected to be "backward" so if Clara were just firing these rules in their order they are defined the test would fail.  A few quick experiments suggest that defining sessions based on individual rules doesn't work in ClojureScript, which makes some sense to me based on the compilation model in ClojureScript, although I'd need to dig further into it to be sure I'm not just missing something.  If this is true, you'd need to create different namespace sources to vary the rule order, but I don't think that's really necessary since you test the "backwards" case.

I also don't see what the dotimes is accomplishing - it looks to me like it is just executing the same test case 10 times.  Am I missing something?

WilliamParker:
Please squash the two commits into one; the motivation for this is to have each commit in the history reflect a distinct logical change.

WilliamParker:
Since we're testing that salience works in ClojureScript with this change we'll want to update any docs/comments that say that salience doesn't work in ClojureScript.  I don't mind merging this and having those updates be done later by either you or someone else though.

WilliamParker:
@alex-dixon I left a few comments, but overall I think this is close to ready for merging.

I also agree with @rbrush 's comment at https://github.com/cerner/clara-rules/pull/288#issuecomment-298907118 that some refactoring of Clara's tests is in order but, as Ryan said, that shouldn't block this.

src/test/clojurescript/clara/test_salience.cljs
alex-dixon: Re: dotimes and the comment, both were copied from the CLJ test for salience in `test-rules.clj`. My impression was the author ran it 10 times be extra sure it worked. 😅 Will check to make sure I didn't miss something.

alex-dixon:
Looked at the docs for salience on the clara-rules site as well as docs generated by codox and didn't see any reference to lack of salience support in Clojurescript. Removed the comment here: https://github.com/cerner/clara-rules/blob/master/src/main/clojure/clara/rules.cljc#L248. Commits should be squashed now.

src/test/clojurescript/clara/test_salience.cljs
WilliamParker: I logged https://github.com/cerner/clara-rules/issues/292; could you add a comment referencing that issue here?

src/test/clojurescript/clara/test_salience.cljs
WilliamParker: In [the Clojure test-salience test](https://github.com/cerner/clara-rules/blob/0.14.0/src/test/clojure/clara/test_rules.clj#L2887) each individual execution will create a session with a [new call to shuffle](https://github.com/cerner/clara-rules/blob/0.14.0/src/test/clojure/clara/test_rules.clj#L2951), and presumably 10 different calls to shuffle will not all yield the same order.  That isn't the case here since you're using the namespace symbol as the source of rules for the session.

WilliamParker:
I see two commits in the PR - it looks like you squashed the comment change for https://github.com/cerner/clara-rules/pull/288#issuecomment-300252923 onto the fix for https://github.com/cerner/clara-rules/pull/288#discussion_r115033257 but didn't squash onto the original commit.

alex-dixon:
Added comment, removed dotimes, and hopefully squashed correctly :)

WilliamParker:
This looks good to me now and I don't think @rbrush had any outstanding concerns so I'm going to merge it.  Thanks for this @alex-dixon .

Issue #292, opened 2017-05-10
Investigate non-compile-time sessions in ClojureScript in tests


WilliamParker:
A [pull request](https://github.com/cerner/clara-rules/pull/288) was submitted in relation to [issue 285](https://github.com/cerner/clara-rules/issues/285) to add tests validating that salience works in ClojureScript.  In this PR, there was this usage of defsession:

````
(deftest test-salience ......
    (defsession test-salience-session 'clara.test-salience
                     :cache false
                     :activation-group-sort-fn (condp = sort-fn
                                                 :default-sort nil
                                                 :numeric-greatest-sort numeric-greatest-sort
                                                 :numeric-least-sort numeric-least-sort
                                                 :boolean-greatest-sort  >
                                                 :boolean-least-sort <)
                     :activation-group-fn (condp = group-fn
                                            :default-group nil
                                            :salience-group salience-group-fn
                                            :neg-salience-group neg-salience-group-fn))
````

The point of having defsession in ClojureScript is that due to the lack of eval in ClojureScript, which is used heavily by [mk-session](https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/rules/compiler.clj#L1734) we can't generate sessions at runtime in ClojureScript as we can in Clojure.  However, because the rule sources work at compile time and only the options needed to fall through to runtime evaluation, and the defsession macro was organized in a way that inadvertently facilitated this, this actually worked.

I didn't want to derail adding the tests with this larger issue, but I'd like to consider a couple options here.  This was also discussed in a comment on the PR at https://github.com/cerner/clara-rules/pull/288#discussion_r115544060.

- Should we create a macro that doesn't create a definition for the session?  Basically it would just be a macro that has this same behavior, but intentionally.  If we add this, should it be for internal test use only or a public API?
- There has been some work to add eval to ClojureScript; perhaps this is mature enough that we could use it in Clara's tests.
- Other options to consider?

alex-dixon:
@WilliamParker Should this be closed?

WilliamParker:
Yes, I don't see anything left to do here after the changes for https://github.com/cerner/clara-rules/issues/296, closing.

PR #295, opened 2017-05-11
Issue 241: Wrap the fact-type-fn and ancestors-fn to add the fix for …

WilliamParker:
…issue 149 to the ClojureScript version

### DIFF
--- a/src/main/clojure/clara/rules.cljc
+++ b/src/main/clojure/clara/rules.cljc
@@ -142,9 +142,24 @@
 
     ;; We preserve a map of fact types to alpha nodes for efficiency,
     ;; effectively memoizing this operation.
-    (let [alpha-map (atom {})]
+    (let [alpha-map (atom {})
+          wrapped-fact-type-fn (if (= fact-type-fn type)
+                                 type
+                                 (fn [fact]
+                                   (if (isa? (type fact) :clara.rules.engine/system-type)
+                                     ;; Internal system types always use ClojureScript's type mechanism.
+                                     (type fact)
+                                     ;; All other types defer to the provided function.
+                                     (fact-type-fn fact))))
+          wrapped-ancestors-fn (fn [fact-type]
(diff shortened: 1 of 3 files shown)


mrrodriguez:
The change looks fine. However, I'm starting to wonder why the cljs has to have so much repetitive code with clj side for this stuff?

It looks to me like there could be quite a bit more sharing in these building session steps, unless I'm missing something?

That could be a separate concern though. I'm just curious the thoughts on that are. Looks like a lot of duplication going on.

alex-dixon:
@mrrodriguez Ryan also mentioned duplication of CLJ/CLJS (specifically in tests)  https://github.com/cerner/clara-rules/pull/288#issuecomment-298907118. There's also a proposal for a CLJS macro that behaves more like `mk-session` (#292). 

William has mentioned Clara's CLJ code uses `eval` which hasn't worked well in CLJS in the past, but that it may have changed recently. I looked around a bit and think he may be referring to [`cljs.js/eval`](http://cljs.github.io/api/cljs.js/eval) released in CLJS 1.7.10.

mrrodriguez:
@alex-dixon thanks for the reminder. I read that from Ryan before and didn't realize it was the same code we were talking about.

I think the `eval` suggestion may be fine to explore for testing, however, I don't think that'd be something that we'd actually want to have happening in real production use.

However, there are a few places here that look like it there could just be the correct functions pulled out to share code between the clj vs cljs impl. The stuff that isn't involved with code gen or macros.

e.g. the alpha fn wrapper logic doesn't strike me as something that would need much separation frm the cljs vs clj side.

Anyways, it isn't a big issue for the sake of this PR. I appreciate your input though.

WilliamParker:
I've looked over this again and I don't see any problems; Mike looked at it as well from https://github.com/cerner/clara-rules/pull/295#issuecomment-301111945.  We should give some further thought to merging our Clojure and ClojureScript code bases to have as much common code as possible but this can be addressed on other issues.

Merging.

Issue #300, opened 2017-05-31
Resolution of ClojureScript symbols for condition types to Java classes


WilliamParker:
The following ClojureScript rule fails to compile:

````
(defrecord Number [])

(defrule num-rule
  [Number]
  =>
  (println "Hi number!"))
````
with the following exception:

````
Caused by: java.lang.IllegalArgumentException: No method in multimethod 'emit-constant' for dispatch value: class java.lang.Class
````
The reason for this can be found [here](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/rules/dsl.clj#L49).  In Clojure, we can embed classes into the returned rule structure, while in ClojureScript we have to return a symbol.  The problem arises when the symbol in question resolves to something in the Clojure environment of the ClojureScript compiler; we then try to embed Java objects, in this case a Class instance, in JavaScript code.  I believe the fix here is to only attempt resolution in the Clojure environment if we are in fact compiling ClojureScript as determined by the clara.rules.compiler/compiling-cljs? function.

This came up when I was working on https://github.com/cerner/clara-rules/issues/296 and the classes for our [Clojure test facts](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/rules/testfacts.clj) existed in the ClojureScript compiler environment when compiling rulesets using our [ClojureScript test facts](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/rules/testfacts.clj).  In the case of the simplified example above, Number will resolve to java.lang.Number and we'll attempt to compile the class object into JavaScript, which we can't do, so an exception is thrown.

WilliamParker:
This was fixed by commit https://github.com/cerner/clara-rules/commit/03ded9f626ae4034c4563418bb44f36daede3ce5 in PR https://github.com/cerner/clara-rules/pull/301

Closing.

Issue #296, opened 2017-05-17
Create a rules test definition macro that can be shared between Clojure and ClojureScript


WilliamParker:
Our tests for the ClojureScript version of Clara are far less extensive than our tests for the Clojure version. A major reason for this is that testing is easier in Clojure than in ClojureScript since Clojure supports on-the-fly creation of rules sessions, while the ClojureScript requires compile-time creation of rules sessions.  This prevents us from writing tests like this for ClojureScript:

````

(deftest test-simple-query
  (let [cold-query (dsl/parse-query [] [[Temperature (< temperature 20) (= ?t temperature)]])

        session (-> (mk-session [cold-query])
                    (insert (->Temperature 15 "MCI"))
                    (insert (->Temperature 10 "MCI"))
                    (insert (->Temperature 80 "MCI"))
                    fire-rules)]

    (is (= #{{:?t 15} {:?t 10}}
           (set (query session cold-query))))))
````

The alternative in ClojureScript of creating a ClojureScript namespace or creating a Clojure namespace containing the rules like [test-rules-data](https://github.com/cerner/clara-rules/blob/0.14.0/src/test/clojurescript/clara/test_rules_data.clj) is inconvenient and, multiplied over the large number of test cases we have, reduces readability of the tests considerably.

The end result is that most of our tests only run for Clojure.  While this works to a large degree since most of the codebase is shared between Clojure and ClojureScript, this is not always the case.  In particular, a variety of performance optimizations we've made for the Clojure version rely on JVM interop, which obviously won't work in ClojureScript.  In particular, the [memory](https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/rules/memory.cljc) uses Java data structures extensively.  The holes this creates in our testing of the ClojureScript version are problematic; there are any number of ways in which we could cause a regression in the ClojureScript version without impacting the Clojure version.

Ideally we'd have a common testing codebase between Clojure and ClojureScript that kept the desirable characteristics of our current Clojure tests.  I think this can be achieved with a custom test definition macro that would look very similar to our current tests but would create the sessions in ClojureScript at compile time and at runtime in Clojure.  I'm envisioning something like

````
(def-rules-test simple-test
  :rules [cold-rule (....)
          hot-rule (....)]
  :queries [cold-query]
  :sessions [session1 [cold-rule hot-rule] {opts-here}
             session2 [cold-rule] {opts-here}]
  (body-of-tests-here))
````

[...]

PR #301, opened 2017-05-31
Issues 296 and 300 - Add a macro for common tests between Clojure and ClojureScript

WilliamParker:


WilliamParker:
For some reason Travis CI doesn't seem to be picking up /src/test/common although my local builds do.  This looks like it is an issue with master too though.

src/main/clojure/clara/rules/dsl.clj
```diff
@@ -47,7 +47,12 @@
   "Creates a condition with the given optional result binding when parsing a rule."
   [condition result-binding expr-meta]
   (let [type (if (symbol? (first condition))
-               (if-let [resolved (resolve (first condition))]
+               (if-let [resolved (and
+                                  ;; If we are compiling ClojureScript rules we don't want
+                                  ;; to resolve the symbol in the ClojureScript compiler's
+                                  ;; Clojure environment.  See issue 300.
+                                  (not (com/compiling-cljs?))
+                                  (resolve (first condition)))]
```
mrrodriguez: It does seem odd to me that the clj impl would embed a class object in the data structure at this point. A fully-qualified symbol seems more appropriate and a "data representation". The fact that clj lets you embed class objects in `eval` code etc just seems to be non-documented behaviors.

This is a broader topic though, so not really suitable to be fixed in this issue.

src/main/clojure/clara/tools/testing_utils.clj
```diff
@@ -0,0 +1,66 @@
...
+                                                     (map (fn [[session-name productions session-opts]]
+                                                            [session-name (if (com/compiling-cljs?)
+                                                                            (m/productions->session-assembly-form (map eval productions) session-opts)
+                                                                            `(com/mk-session ~(into [(vec productions)]
+                                                                                                    cat
+                                                                                                    session-opts)))]))
+                                                     cat)))
+        
+        test-form `(~(if (com/compiling-cljs?)
+                       'cljs.test/deftest
```
mrrodriguez: This is sort of a macro-anti-pattern to return code that has implicit dependencies on namespaces being pre-required by the caller.

If this macro was used by a caller that did not have a `require` on `cljs.test` (or clj) and it wasn't already loaded elsewhere, then it would fail.

src/test/common/clara/test_testing_utils.cljc
```diff
@@ -0,0 +1,56 @@
+#?(:clj
```
mrrodriguez: It is somewhat non-idiomatic to use a read conditional at this scope.

It is more common to see them used just in the places where the forms differ. Here there is duplication and even a chance the `ns` name could be changed, etc. 

It's not a huge issue though.

src/main/clojure/clara/tools/testing_utils.clj
WilliamParker: I discussed this some in the docstring here.  I agree that it is undesirable; unfortunately, my understanding is that ClojureScript doesn't have good support for macros causing additional namespaces to be required since all requires have to be at the beginning of the file.  I'd be happy to find out I'm wrong about that or that there is a way around this problem though, and I'm certainly not as well-versed in ClojureScript's compilation model as in Clojure's.  Practically speaking, for this particular use-case it probably isn't a big deal since I'd expect the namespaces consuming this to have those requires anyway and this is currently targeted toward consumption within Clara, not as an external API.

src/main/clojure/clara/rules/dsl.clj
WilliamParker: Yeah, I agree it is a bit odd but probably outside the scope of https://github.com/cerner/clara-rules/issues/296 or https://github.com/cerner/clara-rules/issues/300

src/main/clojure/clara/tools/testing_utils.clj
mrrodriguez: Oh, wow, you explained this already thoroughly in the doc string. I somehow overlooked it.
That makes sense what you are saying. 

It'd be pretty tricky to pull of the require on the cljs side I'd think so seems fine for now.

src/test/common/clara/test_testing_utils.cljc
WilliamParker: Yeah, doing it all in one body with reader conditionals is also an option.  In this case I just thought that so much differed for such a small ns header that it would be more readable to just have entirely separate forms.  Looking back at it I'm thinking perhaps make the namespace name common and just have the rest be separate.

src/main/clojure/clara/rules/dsl.clj
rbrush: Yeah, we can drop this in another change. I originally had wanted to validate the presence of a type early in the process to fail fast, but it doesn't make sense to include in our structure.

rbrush:
Looks good to me!

src/test/common/clara/test_testing_utils.cljc
WilliamParker: Having everything other than the name be separate is complicated by having multiple top-level forms beneath the declaration for clj and cljs.  I think I'll just keep this as is; by my count only the clara.rules require is identical between clj and cljs so splicing in lots of reader conditions would hurt readability more than it would help it in this instance IMO.

Issue #298, opened 2017-05-22
:exists conditions perform downstream work when their state didn't change


WilliamParker:
We added support for :exists conditions that only fire once per binding group with https://github.com/cerner/clara-rules/issues/130.  Since they use a [count accumulator](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/rules/accumulators.cljc#L161) behind the scenes, Clara will consider a change to have taken place each time a match is added or removed, even if it doesn't change whether a match exists.  This implementation can be found in [extract-exists](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/rules/compiler.clj#L612). For example, suppose we have a condition like

````
[:exists [Cold]]
````
and 2 Cold facts in the session.  We shouldn't perform any downstream work in the rules network when we retract or add 1 Cold fact since we'd have 1 or 3 Cold facts afterward.  That is, we have a Cold fact before and a Cold fact after in these cases, so the answer to the question "Does a Cold fact exist in the session" did not change.  However, we will current retract downstream work based on a count of 2 and then perform that work again based on a count of 3, even though the count is only used by a :test condition added by extract-exists that determines whether the count is positive.

We can improve this by adding a :convert-return-fn to the accumulator in this case that returns true if the count is greater than 0 and nil if is 0.  Due to the enhancements in https://github.com/cerner/clara-rules/issues/182 if the final result of the accumulator does not change then work will not be performed downstream in the rules network.

WilliamParker:
PR at https://github.com/cerner/clara-rules/pull/411

WilliamParker:
Resolved by https://github.com/cerner/clara-rules/pull/411  , closing

PR #314, opened 2017-06-18
Refactor the tests for :exists conditions from test_rules.clj to thei…

WilliamParker:
…r own .cljc file


WilliamParker:
Admittedly this test file is relatively small, but I think these tests are logically distinct enough from tests of other functionality to merit separation into their own file.  It is also likely that https://github.com/cerner/clara-rules/issues/298 will require new tests for :exists conditions.

PR #323, opened 2017-07-06
Move tests focused on DSL functionality such as binding visibility in…

WilliamParker:
… edge cases, fact field access, etc. to their own namespace shared between Clojure and ClojureScript

### DIFF
--- a/src/main/clojure/clara/tools/testing_utils.clj
+++ b/src/main/clojure/clara/tools/testing_utils.clj
@@ -75,4 +75,3 @@
   (f)
   (with-redefs [uc/get-ordered-update-cache ca/get-cancelling-update-cache]
     (f)))
-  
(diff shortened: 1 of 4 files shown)


WilliamParker:
This PR creates a test file, shared between Clojure and ClojureScript, containing tests that are focused on edge cases of ways in which the user can write rules rather than the substance of different rules.  For example, tests that variations in ways bindings can be created don't incorrectly impact outcomes, that fields with strange names work correctly, etc.  Admittedly this is a somewhat nebulous category of tests but looking over test-rules I saw enough tests along such lines that the thematic grouping made sense to me.

Unfortunately some of the tests failed when moved over to ClojureScript, so we'll want to investigate those and log appropriate issues to fix the divergence from the Clojure version, so I don't think that needs to block merging this.  We also need to land on a strategy for testing what exceptions are thrown in ClojureScript before moving over some error handling test cases.

rbrush:
Looks good to me. It is interesting that some of those test don't work in CLJS, but I agree that troubleshooting those should be independent of restructuring the tests.

PR #326, opened 2017-07-10
Issue 325: Fix retraction function of the distinct accumulator to res…

WilliamParker:
…pect the existing cardinality of the accumulated values


mrrodriguez:
Seems like a good find.

rbrush:
+1 to merge

PR #328, opened 2017-07-23
Refactor tests for AccumulateNode and AccumulateWithJoinFilterNode to…

WilliamParker:
… be cross-platform (clj and cljs)


WilliamParker:
This refactors tests for support for accumulators in the engine (as opposed to specific accumulators provided by Clara) into their own cross-platform test file.  Unfortunately several tests fail; I've put those in clj-only code blocks for now pending a fix.  I haven't investigated the underlying cause of the failures.

WilliamParker:
I'm going to go ahead and merge this since

- There is no direct impact on consumers since this is just test code.
- Despite the volume of the changes in terms of lines of code, this is fundamentally a very straightforward uplift using established patterns.

In any case, if there is anything anyone wants to discuss on this, we can revisit it later with another commit.

Issue #303, opened 2017-06-06
IllegalAccessError from lazy use of TransientLocalMemory


WilliamParker:
I'd expect the following test to pass:

````
(def-rules-test test-negation-complex-join-with-numerous-non-matching-facts-inserted-after-descendant-negation
  {:rules []
   :queries [query1 [[]
                     [[Hot (= ?t temperature)]
                      [:not [Cold (join-filter-equals ?t temperature)]]]]]
   :sessions [empty-session [query1] {}]}
  (let [lots-of-hot (doall (for [_ (range 100)]
                             (->Hot 20)))]
    (is (= (repeat 100 {:?t 20})
           (-> empty-session
               (insert (->Cold 10))
               fire-rules
               (insert-all lots-of-hot)
               fire-rules
               (query query1))))))
````

Instead, on the JVM it throws:

````
java.lang.IllegalAccessError: Transient used after persistent!
 at clojure.lang.PersistentArrayMap$TransientArrayMap.ensureEditable (PersistentArrayMap.java:457)
    clojure.lang.ATransientMap.valAt (ATransientMap.java:78)
    clojure.lang.RT.get (RT.java:762)
    clara.rules.memory.TransientLocalMemory.get_elements (memory.cljc:441)
    clara.rules.engine.NegationWithJoinFilterNode$iter__16861__16865$fn__16866$fn__16867.invoke (engine.cljc:829)
    clara.rules.engine.NegationWithJoinFilterNode$iter__16861__16865$fn__16866.invoke (engine.cljc:826)
    clojure.lang.LazySeq.sval (LazySeq.java:40)
    clojure.lang.LazySeq.seq (LazySeq.java:49)
    clojure.lang.RT.seq (RT.java:521)
    clojure.core$seq__4357.invokeStatic (core.clj:137)
````

The full stack trace is in [this gist](https://gist.github.com/WilliamParker/6b927972aac07dd83ac98677e26ca589).

[...]

PR #331, opened 2017-07-30
Refactor tests focused on edge cases in the memory to be cross-platfo…

WilliamParker:
…rm using def-rules-test


WilliamParker:
This PR refactors tests focused on edge cases in the memory to their own file.  Looking over test-rules I thought that these sorts of tests were different enough from the tests focused on the rule networks to merit separation, even though they don't test the memory directly, but rather indirectly via rules scenarios chosen to do so with knowledge of the Clara internals.  In addition to the ones already here I anticipate adding tests for https://github.com/cerner/clara-rules/issues/303 here when that is done (likely in the near future).

rbrush:
+1; LGTM

PR #332, opened 2017-08-08
Issue 330: Fix ClojureScript accumulation and memory tests that fail …

WilliamParker:
…due to different semantics of the core type function between Clojure and ClojureScript

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -53,6 +53,10 @@
                               "phantom-advanced" ["phantomjs"
                                                   "src/test/js/runner.js"
                                                   "src/test/html/advanced.html"]}}
+
+  :repl-options {;; The large number of ClojureScript tests is causing long compilation times
+                 ;; to start the REPL.
+                 :timeout 120000}
   
   ;; Factoring out the duplication of this test selector function causes an error,
   ;; perhaps because Leiningen is using this as uneval'ed code.
--- a/src/test/common/clara/test_accumulation.cljc
+++ b/src/test/common/clara/test_accumulation.cljc
@@ -69,6 +69,16 @@
   [& args]
   (apply = args))
 
+;; In ClojureScript the type function does not use the :type key in metadata as in Clojure, but we had
+;; tests in Clojure that were relying on this.  Creating a custom function for the fact type along these
+;; lines was the easiest way to get the tests relying on this behavior to pass in ClojureScript.  I used
+;; data on the actual fact rather than metadata and updated the tests accordingly.  This is desirable since
+;; Clara's value-based semantics won't consider the metadata part of the fact for the purpose of memory operations.
+(defn type-or-class
+  [fact]
+  (or (:type fact)
+      (type fact)))
+
 (def-rules-test test-simple-binding-variable-ordering
 
   {:rules [rule1 [[[Temperature (< temperature 20) (= ?t temperature)]]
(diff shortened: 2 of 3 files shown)


WilliamParker:
Very brief summary: The tests in test_accumulation and test_memory that failed in ClojureScript but passed in Clojure did so due to flaws in the tests themselves rather than a bug in Clara.

[Clojure type function:](https://github.com/clojure/clojure/blob/clojure-1.8.0/src/clj/clojure/core.clj#L3357)
````
(defn type 
  "Returns the :type metadata of x, or its Class if none"
  {:added "1.0"
   :static true}
  [x]
  (or (get (meta x) :type) (class x)))
````
[ClojureScript type function:](https://github.com/clojure/clojurescript/blob/r3308/src/main/cljs/cljs/core.cljs#L206)
````
(defn type
  "Return x's constructor."
  [x]
  (when-not (nil? x)
    (.-constructor x)))
````

In these tests we have facts to which we added the :type key on their metadata and expected its value to become the type for the rules/queries to match the fact.  This works since the default :fact-type-fn is the type function.  Despite the large number of LOC impacted, this change is actually quite simple - I just added :type to the facts in question and added a custom :fact-type-fn to the tests in question that look at it.  Some cases needed to have the :type key added to the fact expectations.  I went with literally adding the key rather than continuing to use metadata since making the fact metadata impact its semantics is dangerous since the Clara memory won't look at it when performing operations.

File: src/test/common/clara/test_accumulation.cljc [hunk 1]
```diff
@@ -1540,9 +1551,10 @@
                             (cond (false? a) false
                                   (false? b) true
                                   :else (throw
-                                         (IllegalStateException. (str "This test should only compare false and numeric values, "
-                                                                      \newline not
-                                                                      "a numeric value to another numeric value")))))
+                                         (ex-info (str "This test should only compare false and numeric values, "
+                                                       \newline not
+                                                       "a numeric value to another numeric value")
+                                                  {:a a :b b}))))
```
WilliamParker: We can't throw IllegalStateException in ClojureScript since it is a JVM class.  This would have failed with another exception if we actually reached this in a ClojureScript/JavaScript environment.

mrrodriguez:
> I went with literally adding the key rather than continuing to use metadata since making the fact metadata impact its semantics is dangerous since the Clara memory won't look at it when performing operations.

Just thinking through this, shouldn't it not be dangerous and something that Clara memory should be robust to? Are we just hiding a problem by making the test avoid the "equal facts, different instances" problem?

project.clj
```diff
@@ -53,6 +53,10 @@
                               "phantom-advanced" ["phantomjs"
                                                   "src/test/js/runner.js"
                                                   "src/test/html/advanced.html"]}}
+
+  :repl-options {;; The large number of ClojureScript tests is causing long compilation times
```
mrrodriguez: This is a little surprising to me. Maybe something to look into at some point.
What is your "start repl" workflow that you experienced such long build times, just for reference?

src/test/common/clara/test_accumulation.cljc
```diff
@@ -69,6 +69,16 @@
   [& args]
   (apply = args))
 
+;; In ClojureScript the type function does not use the :type key in metadata as in Clojure, but we had
```
mrrodriguez: This is a difference I was not aware of. I know I wrote some tests adding "type metadata" before, so thanks for the cleanup!

src/test/common/clara/test_accumulation.cljc
```diff
@@ -69,6 +69,16 @@
...
+;; tests in Clojure that were relying on this.  Creating a custom function for the fact type along these
+;; lines was the easiest way to get the tests relying on this behavior to pass in ClojureScript.  I used
+;; data on the actual fact rather than metadata and updated the tests accordingly.  This is desirable since
+;; Clara's value-based semantics won't consider the metadata part of the fact for the purpose of memory operations.
```
mrrodriguez: See https://github.com/cerner/clara-rules/pull/332#issuecomment-320998627 for related to this.

src/test/common/clara/test_accumulation.cljc
mrrodriguez: I'm surprised it compiled in cljs.

project.clj
WilliamParker: When I start my REPL with "lein repl" it compiles all of the ClojureScript test files for each build(figwheel, simple, and advanced) if the compiled JS doesn't exist already.  I agree that it shouldn't need to do that, at most it should need to compile the Figwheel profile only.

PR #334, opened 2017-08-15
Issue 303: IllegalAccessError from lazy use of TransientLocalMemory

ris5438:
Brief summary: Replaced all instances of `for` in engine with an `eager-for` macro which forces realization. Also moved the `get-elements` calls in `NegationWithJoinFilterNode` out of the `:when` predicate forcing them to occur before passing the lazy sequence of tokens on.


src/main/clojure/clara/rules/platform.cljc
```diff
@@ -75,4 +75,10 @@
            ~@body
            (finally
              ~@(for [[tl] binding-pairs]
-                 `(.remove ~tl)))))))
\ No newline at end of file
+                 `(.remove ~tl)))))))
+
+#?(:clj
```
mrrodriguez: There isn't anything platform-specific here. `defmacro` is fine in cljs's compiler. So you shouldn't need to wrap it in this read conditional `:clj`.

Other places in this file use read conditionals due to using types that are platform-specific.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -6,9 +6,12 @@
             [clojure.string :as string]
             [clara.rules.memory :as mem]
             [clara.rules.listener :as l]
-            [clara.rules.platform :as platform]
+            #?(:clj [clara.rules.platform :as platform :refer [eager-for]]
```
mrrodriguez: I'm always just not a fan of `:refer` and I don't think we really use it much in this ns.
This could be written as:
```clj
#?(:clj [clara.rules.platform :as platform]
   :cljs [clara.rules.platform :as platform :include-macros true])
```
I think you can even get away with
```clj
[clara.rules.platform :as platform #?@(:cljs [:include-macros true])]
```
but I don't tend to see that for whatever reason.

See http://blog.fikesfarm.com/posts/2016-03-01-clojurescript-macro-sugar.html for more on the `ns` "sugar"

But perhaps others like the `:refer` and if that is the case, so be it. I just think "alias everything" improves readability overall since it is immediately clear when a var is "foreign" to a ns or not.

And could then just use the alias as always.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -6,9 +6,12 @@
...
+               :cljs [clara.rules.platform :as platform])
             [clara.rules.update-cache.core :as uc]
-            #?(:clj [clara.rules.update-cache.cancelling :as ca])))
+            #?(:clj [clara.rules.update-cache.cancelling :as ca]))
+  #?(:cljs
+     (:require-macros [clara.rules.platform :refer [eager-for]])))
```
mrrodriguez: This would go away if the above `:include-macros` pattern was adopted instead.

src/test/common/clara/test_memory.cljc
```diff
@@ -43,10 +44,48 @@
 
 (use-fixtures :once st/validate-schemas #?(:clj tu/opts-fixture))
 
+(defn join-filter-equals
```
mrrodriguez: Seems like this could be added to `clara.tools.testing-utils` and then used in the (now) various testing ns's?

src/test/common/clara/test_memory.cljc
```diff
@@ -43,10 +44,48 @@
...
+  "Intended to be a test function that is the same as equals, but is not visible to Clara as such
+  and thus forces usage of join filters instead of hash joins"
+  [& args]
+  (apply = args))
+
 ;; While the memory is tested through rules and queries, rather than direct unit tests on the memory,
 ;; the intent of these tests is to create patterns in the engine that cover edge cases and other paths
 ;; of concern in clara.rules.memory.
 
+(def-rules-test test-negation-complex-join-with-numerous-non-matching-facts-inserted-after-descendant-negation
```
mrrodriguez: Perhaps a link to the GitHub issue containing more details about what is happening? 
I wouldn't say it is at all obvious what this test is really meant to test. It just looks like a really specific scenario with no rationale given when looking at this later once we've forgotten the initial reason.

src/test/common/clara/test_memory.cljc
```diff
@@ -43,10 +44,48 @@
...
+                             (->Hot 20)))]
+    (is (= (repeat 100 {:?t 20})
+           (-> empty-session
+               (insert (->Cold 10))
+               fire-rules
+               (insert-all lots-of-hot)
+               fire-rules
+               (query query1))))))
+
+(def-rules-test test-negation-complex-join-with-numerous-non-matching-facts-inserted-after-descendant-negation-with-equals
```
mrrodriguez: Sort of unfortunate the whole test body has to be repeated just to change `join-filter-equals` to `=`.
Maybe that is an artifact of `def-rules-test` and perhaps it could be improved later.

Maybe a comment here saying this is the same as above though would be good to make it clear they are supposed to be in sync.

src/test/common/clara/test_memory.cljc
WilliamParker: +1

src/main/clojure/clara/rules/platform.cljc
WilliamParker: I'd have thought this would be a problem since ClojureScript macros are evaluated by the JVM compiler, at least if you're not using self-hosted ClojureScript; I told @ris5438 to do it this way.  Perhaps I'm incorrect or out of date though.  Do you know offhand of any good references, blog posts, etc. on the subject?

src/main/clojure/clara/rules/engine.cljc
ris5438: Good point. I prefer aliasing as well. Updated.

src/main/clojure/clara/rules/engine.cljc
ris5438: Updated.

src/main/clojure/clara/rules/platform.cljc
mrrodriguez: I guess the best explanation is just https://clojurescript.org/about/differences#_macros but that doesn't really get directly to the point.

In cljs macros are done in a separate compilation stage and do need to be in a clj or cljc file. However, they are not a feature exclusive to a single platform - clj vs cljs.

The read conditionals are for forms that are exclusive to the particular platform, like `java.lang.String` vs `js/String`. 

`defmacro` is defined in both the clj and cljs compiler. Cljs macros must (of course) emit code that is meant for cljs compilation though.

Can't put macros in .cljs file because the cljs compiler expects the macros to have already be defined in a prior "compilation stage".

ris5438: That clears it up. Updated.

src/test/common/clara/test_memory.cljc
ris5438: Refactored to use `join-filter-equals` from `clara.tools.testing-utils`

src/main/clojure/clara/tools/testing_utils.clj
```diff
@@ -75,3 +75,9 @@
   (f)
   (with-redefs [uc/get-ordered-update-cache ca/get-cancelling-update-cache]
     (f)))
+
```
WilliamParker: This is a clj file, if you want to use a function in it in ClojureScript it will need to be turned into a cljc file.  Note the [failed tests in your Travis build](https://travis-ci.org/cerner/clara-rules/builds/264850588?utm_source=github_status&utm_medium=notification).

ris5438: Would you recommend turning it into a cljc file or moving that function to a different namespace?

src/main/clojure/clara/tools/testing_utils.cljc
```diff
@@ -0,0 +1,89 @@
...
+     (:require [clara.rules.update-cache.core :as uc]
+               [clara.rules.update-cache.cancelling :as ca]
+               [clara.rules.compiler :as com]
+               [clara.macros :as m]
+               [clara.rules.dsl :as dsl]))
+   :cljs
+   (ns clara.tools.testing-utils
+     (:require [clara.rules.update-cache.core :as uc])))
+
+#?(:clj
```
mrrodriguez: Similar to our past discussion on macros in cljs, this shouldn't be specified in a read-conditional. It isn't platform specific. It is used in cljs testing.

src/main/clojure/clara/tools/testing_utils.cljc
```diff
@@ -0,0 +1,89 @@
...
+   (ns clara.tools.testing-utils
+     "Internal utilities for testing clara-rules and derivative projects.  These should be considered experimental
+  right now from the perspective of consumers of clara-rules, although it is possible that this namespace 
+  will be made part of the public API once its functionality has proven robust and reliable.  The focus, however,
+  is functionality needed to test the rules engine itself."
+     (:require [clara.rules.update-cache.core :as uc]
+               [clara.rules.update-cache.cancelling :as ca]
```
mrrodriguez: This should be the only `:clj` only ns from the ones I see here.
This relates to https://github.com/cerner/clara-rules/pull/334/files#r133448443

So the whole `ns` header should be
```clj
(ns clara.tools.testing-utils
   "Internal utilities for testing clara-rules and derivative projects.  These should be considered experimental
  right now from the perspective of consumers of clara-rules, although it is possible that this namespace 
  will be made part of the public API once its functionality has proven robust and reliable.  The focus, however,
  is functionality needed to test the rules engine itself."
   (:require [clara.rules.update-cache.core :as uc]
             #?(:clj [clara.rules.update-cache.cancelling :as ca])
             [clara.rules.compiler :as com]
             [clara.macros :as m]
             [clara.rules.dsl :as dsl]))
```

src/main/clojure/clara/tools/testing_utils.clj
WilliamParker: I'd prefer to turn it into a .cljc file rather than create another namespace entirely.  That said I'm OK with addressing the duplication of the join-filter-equals function as part of a later commit rather than as part of these changes.

PR #335, opened 2017-08-17
Issue 303: Clara transient memory and memory getter mutable return values should not be used lazily

ris5438:
Issue 303: Replaced all instances of for in engine to eager-for

Issue 303: moved macro to platform, added cljc support, added tests

Issue 303: Fix indentation of eager-for's

Issue 303: Moved the get-elements calls in NegationWithJoinFilterNode out of the :when predicate

Issue 303: Fixed ns header and made the macro non platform specific

Issue 303: Moved join-filter-equals to a common ns clara.tools.testing-utils

Issue 303: Made testing_utils from clj to cljc

Brief summary: Replaced all instances of for in engine with an eager-for macro which forces realization. Also moved the get-elements calls in NegationWithJoinFilterNode out of the :when predicate forcing them to occur before passing the lazy sequence of tokens on.


WilliamParker:
Quick update for anyone tracking on this: this is undergoing some testing inside Cerner for performance impacts, hence the extended time without activity on the PR.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1700,10 +1704,10 @@
 (defn variables-as-keywords
   "Returns symbols in the given s-expression that start with '?' as keywords"
   [expression]
-  (into #{} (for [item (tree-seq coll? seq expression)
-                  :when (and (symbol? item)
-                             (= \? (first (name item))))]
-              (keyword item))))
+  (into #{} (platform/eager-for [item (tree-seq coll? seq expression)
```
WilliamParker: @ris5438 I'm OK with leaving this for the sake of the pattern of using eager-for in the engine namespace, but just FYI putting the contents of a LazySeq in a set as here will force realization of the lazy sequence anyway.  Also variables-as-keywords takes s-expression input, so it isn't exposed to the mutability problems in the memory that prompted the overall shift to eager-for.

src/test/common/clara/test_memory.cljc
```diff
@@ -47,8 +49,41 @@
...
+  (let [lots-of-hot (doall (for [_ (range 100)]
+                             (->Hot 20)))]
+    (is (= (repeat 100 {:?t 20})
+           (-> empty-session
+               (insert (->Cold 10))
+               fire-rules
+               (insert-all lots-of-hot)
+               fire-rules
+               (query query1))))))
+;; These tests are here to verify https://github.com/cerner/clara-rules/issues/303
```
WilliamParker: Can we merge these two new tests into a single test body as discussed at https://github.com/cerner/clara-rules/pull/334#discussion_r133224727? Mainly that will clarify what these comments apply to.

src/test/common/clara/test_memory.cljc
```diff
@@ -47,8 +49,41 @@
...
+    (is (= (repeat 100 {:?t 20})
+           (-> empty-session
+               (insert (->Cold 10))
+               fire-rules
+               (insert-all lots-of-hot)
+               fire-rules
+               (query query1))))))
 
+;; This is the same test as above except using `=` istead of `join-filter-equals`.
+;; The two tests are supposed to be in sync.
```
WilliamParker: This comment looks like it applies to the preexisting test below.  Addressing https://github.com/cerner/clara-rules/pull/335/files#r136121745 should take care of this.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -6,7 +6,8 @@
             [clojure.string :as string]
             [clara.rules.memory :as mem]
             [clara.rules.listener :as l]
-            [clara.rules.platform :as platform]
+            #?(:clj [clara.rules.platform :as platform]
+               :cljs [clara.rules.platform :as platform :include-macros true])
```
WilliamParker: Nice, I wasn't aware of the :include-macros option.

WilliamParker:
The functional changes look good to me.  Our performance testing of this on Cerner's production data has revealed no noticeable performance degradation.

There is some minor cleanup left to do here.
- See https://github.com/cerner/clara-rules/pull/335#discussion_r136121745 in test-memory.
- The commit message has various items that seem to refer to steps in the development process e.g. fixing indentation.  I'd rather have a more concise commit message that summarizes the change as a whole rather than an explanation of the intermediate steps during development; from the point of view of the Git history that is what matters, not how the PR was devised.

Otherwise I think we're good to merge this in the next couple days and then cut a release afterward if there are no objections.  I'm planning on cutting this as a 0.16.0 release since in principle some use cases could experiences slowdowns, and so a 0.15.3 release seems inappropriate, but I don't see enough uncertainty to warrant doing this via an RC.  I've spoken off-thread with @rbrush and I believe he is OK with both the changes and this release plan.

For reference, a previous PR https://github.com/cerner/clara-rules/pull/334 for this that was closed and replaced by this PR contains discussion of previous iterations of this.

src/test/common/clara/test_memory.cljc
ris5438: Merged the two tests --> https://github.com/ris5438/clara-rules/commit/848f342f9544e630adb4d40f55af8dae7753e4d3

PR #339, opened 2017-08-31
Issue 307 (Preliminary refactoring): Remove the dependency on the com…

WilliamParker:
…piler from session inspection functionality


src/main/clojure/clara/tools/inspect.clj
```diff
@@ -55,27 +59,37 @@
 
 (defn- get-condition-matches
   "Returns facts matching each condition"
-  [beta-graph memory]
-  (let [join-node-ids (for [[node-id beta-node] (:id-to-condition-node beta-graph)
-                            :let [node-type (:node-type beta-node)]
+  [nodes memory]
+  (let [node-class->node-type (fn [node]
+                                (get {ExpressionJoinNode :join
```
mrrodriguez: The map could just be stored in a ns var since it is constant.

Issue #338, opened 2017-08-24
Inconsistent behavior when using the 'all' accumulator


EthanEChristian:
The [all](https://github.com/cerner/clara-rules/blob/master/src/main/clojure/clara/rules/accumulators.cljc#L186) accumulator has an initial value of vector and values are accumulated into this vector. 

However if there is a retraction, due to [drop-one-of](https://github.com/cerner/clara-rules/blob/master/src/main/clojure/clara/rules/accumulators.cljc#L38) the vector will become a sequence. 

I am not sure this would classify as defect, however it does make me wonder if this accumulator should be consistent with and without retractions.

Clojure Version: 1.8
Clara Version: 0.15.2

mrrodriguez:
I thought that this was fixed long ago, but I think I just did that in a branch at some point.

There is already a fn very similar to `clara.rules.accumulators/drop-one-of`, which is `clara.rules.memory/remove-first-of-each`. It has the added benefit being a bit more optimized. Maybe that should be moved to some common location and used instead? It doesn't take the same signature as `c.r.a/drop-one-of`, but it could be used. Or we could duplicate this sort of logic, I'm just pointing it out due to the similarity in the "remove one" match part of the logic.

I agree with this issue though that `c.r.a/all` has odd `:retract-fn` behavior. It is nearly "incorrect" in that an insertion, retraction, then insertion of the same fact against the accumulator would result in different ordering if the first insertion was done against a vector (append as last item) and the second insertion was done against a generic seq/list (append as first item).

EthanEChristian:
As it seems that `remove-first-of-each` has a bit more overhead, could we go with something like:

```
(defn- drop-one-of
  [^PersistentVector items value]
  (let [idx (.indexOf items value)]
    (if (> idx -1)
      (into (subvec items 0 idx) (subvec items (inc idx)))
      items)))
```
The `if` is probably unnecessary considering the usage would be retracting an item from an accumulator where it should be present.

thoughts @mrrodriguez @WilliamParker ?

mrrodriguez:
`subvec` is really efficient in time, but it retains all the original vector's memory. That is why we typically can't really use it for efficient splicing of items out of vector.

There are actually persistent data libs out there that do allow this sort of behavior, but not going to bring in more dependencies here.

[...]

PR #341, opened 2017-09-20
Issue 338: Inconsistent behavior when using the 'all' accumulator

EthanEChristian:


EthanEChristian:
#338 
@WilliamParker @mrrodriguez

src/test/clojure/clara/test_rules.clj
```diff
@@ -3004,3 +3004,23 @@
...
+                            first
+                            :?all-cold)
+        session (-> (mk-session [cold-query])
+                    (insert cold-fact warmer-fact)
+                    fire-rules
+                    (retract cold-fact)
+                    fire-rules)
+        cold-facts (get-cold-facts session)]
+
+    (is (and (= [warmer-fact] cold-facts)
```
WilliamParker: This isn't a big deal since this isn't the focus of this test, but ideally we'd validate that there is only one query result as well IMO.  So

````
(= 1 (count (query session cold-query)))
````

File: src/test/clojure/clara/test_rules.clj [hunk 2]
```diff
@@ -3004,3 +3004,23 @@
 
       (is (= {:?w 10, :?t nil}
              (-> e (ex-data) :bindings))))))
```
WilliamParker: Please move this test to [test-accumulators](https://github.com/cerner/clara-rules/blob/master/src/test/common/clara/test_accumulators.cljc) and change it to use def-rules-test so that we'll run it in both Clojure and ClojureScript.  For reference, the test-accumulators namespace focuses on tests of individual accumulators, while test-accumulation focuses on the ability of Clara to run arbitrary accumulators.

src/test/clojure/clara/test_rules.clj
```diff
@@ -3004,3 +3004,23 @@
...
+
+;; Validate that the inital vector will remain as a vector even after a retraction occurs
+;; See https://github.com/cerner/clara-rules/issues/338
```
WilliamParker: Should we do similar validation that if a user adds an empty list as the initial value (instead of the default empty vector) that we'll keep that type as well?  I'm inclined toward yes, but don't feel strongly about it right now.  By "add" I mean something like:

````
(assoc (acc/all) :initial-value ())
````

File: src/main/clojure/clara/rules/accumulators.cljc [hunk 4]
```diff
@@ -39,8 +39,10 @@
   "Removes one instance of the given value from the sequence."
   [items value]
   (let [pred #(not= value %)]
-    (into (take-while pred items)
-          (rest (drop-while pred items)))))
+    (into (empty items)
```
WilliamParker: +1 to returning whatever type was passed in here.

WilliamParker:
@EthanEChristian I made some comments on the test but the actual change looks good.

EthanEChristian:
Made recommended changes

Issue #304, opened 2017-06-06
Extracted complex negation rules do not respect binding groups


WilliamParker:
I'd expect the following test case, added to [test-complex-negation](https://github.com/cerner/clara-rules/blob/0.15.0/src/test/clojure/clara/test_rules.clj#L1215), to pass:

````
    ;; Match the nested negation for location ORD but not for MCI.
    (let [end-session (-> s-with-nested
                          (insert
                           (->WindSpeed 10 "MCI")
                           (->Temperature 10 "MCI")
                           (->Cold 20)
                           (->WindSpeed 20 "ORD")
                           (->Temperature 20 "ORD"))
                          (fire-rules))]
      (is (= [{:?l "ORD"}]
             (query end-session nested-negation-with-prior-bindings)))
      (is (no-system-types? end-session)))
````

However, it fails:

````
FAIL in (test-complex-negation) (test_rules.clj:1324)
expected: (= [{:?l "ORD"}] (query end-session nested-negation-with-prior-bindings))
  actual: (not (= [{:?l "ORD"}] ()))
````

Regarding the test case construction, essentially I combined the last two tests in test-complex-negation with different locations and the temperature changed so one set of WindSpeed and Temperature should have the same temperature as the Cold fact as in the last test, but one set wouldn't as in the second to last.

I believe the root cause of this is that the :not condition created by [extract-negation](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/rules/compiler.clj#L986) to fix https://github.com/cerner/clara-rules/issues/149 doesn't look at the bindings, so a NegationResult from the same rule but a different binding group will still be used.

PR #342, opened 2017-09-20
Issue 304: Respect binding groups in complex extracted negations

WilliamParker:
This is a draft of changes to fix https://github.com/cerner/clara-rules/issues/304.  We'll want to uplift these tests to be cross-platform using def-rules-test eventually, but since the relevant tests haven't been uplifted yet I just added a case to the existing test body in test-rules.

I'd like to think on this a bit more and make sure I haven't missed anything before merging but I wanted to go ahead and make this visible to anyone interested.

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -983,16 +983,38 @@
                              "__"
                              (gensym))
 
+          ;; Insert the bindings from ancestors that are used in the negation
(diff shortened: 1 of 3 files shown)


WilliamParker:
I've looked over this again now and think it is correct. 

@rbrush do you have any thoughts on this?

rbrush:
Yes, I like this approach. +1 to merge.

Issue #276, opened 2017-03-16
Provide information on facts accumulated over in session inspection


WilliamParker:
Clara's [session inspection functionality](https://github.com/cerner/clara-rules/blob/0.13.0/src/main/clojure/clara/tools/inspect.clj) currently doesn't provide visibility into the facts that were accumulated over.  So, for example, if we have a condition like

````
[?coldest-temp <- (acc/min :temperature :returns-fact true) :from [Cold]]
````

and we had a session with the facts

````
[(->Cold -10) (->Cold -20)]
````

session inspection would provide the user with the value of ?coldest-temp when this condition matched, but would not provide visibility into the individual Cold facts that were accumulated over.  In this case the logic is trivial, but in cases where there are nontrivial checks on facts before accumulation it would be helpful to know which facts passed the checks and were thus accumulated on.  Furthermore, while simple cases like the example above are easy to reason about for a direct human consumer, tooling built on top of session inspection needs the information to be explicitly provided by Clara.

WilliamParker:
Note that after https://github.com/cerner/clara-rules/issues/190 the engine keeps references to the facts that match accumulators (rather than discarding them after calculating the accumulator result as previously; see that issue for discussion of why the changes were made.

WilliamParker:
I'm looking at adding the accumulated facts to the [Explanation record](https://github.com/cerner/clara-rules/blob/0.13.0/src/main/clojure/clara/tools/inspect.clj#L18).

However, there is a question of passitivity.  If I were doing this from a blank slate, I'd go with a schema like

````
(s/defrecord Explanation [matches :- [{:fact s/Any 
                                       :condition schema/Condition 
                                       (s/optional-key :accumulated-facts) [s/Any]]
                          bindings :- {s/Keyword s/Any}])
````

However, if anyone is consuming Explanation records programmatically, this would obviously be a non-passive change, although all of the data currently present on the explanation would still be present.  This could be made passive by adding the accumulated-facts onto the end of each match tuple if present, but this would be problematic if we ever decided to add something else to the :matches.

On balance, I'm currently learning toward making the non-passive uplift for the :matches to be maps.  If anyone has built tooling on top of session inspection, there would need to be a relatively small uplift, but turning the matches into maps will facilitate future improvements, since any more additions that need to be made will just be additional keys on the map (and thus their addition would be passive).

Thoughts?

rbrush:
[...]

PR #346, opened 2017-10-05
Fix exception in explain-activation

dgoeke:
Previously, when iterating over a list of matches, each match
was assumed to be a vector of `[fact condition]` but was actually
a map of `{:fact ... :condition ...}`. Destructuring the map into
a vector threw an UnsupportedOperationException.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 This is a history of changes to clara-rules.
 
+### 0.17.0-SNAPSHOT
+* Fix exception in explain-activations.
+
 ### 0.16.0
 * Eliminate laziness that broke internal contracts around order of execution, causing an exception to be thrown when executing queries with negation conditions in some edge cases.  See [issue 303](https://github.com/cerner/clara-rules/issues/303) for details.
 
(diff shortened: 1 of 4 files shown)


WilliamParker:
The changes basically look correct to me; looks like I broke this in https://github.com/cerner/clara-rules/issues/276/.    Thanks for submitting a PR for this.

- There aren't any tests of explain-activations, even just that it returns without failing, so whether as part of this PR or not we should add something.  A test could be as basic as validating the explanations for something like

````
(defrule cold-temp
     [?t <- Temperature (< 0 temperature)
     =>
    (insert (->LousyWeather))
````
have "< 0 temperature" in the string somewhere using a regex.  Unfortunately testing against strings is messy (which is why we tested against the underlying inspection API instead) but even a very simple test could help prevent errors like this in the future.  Clojure's [with-out-str](http://clojuredocs.org/clojure.core/with-out-str) would allow such a test to capture the output.
- Adding a 0.17.0 entry to the changelog implies to me that we've released 0.17.0.  Usually we've just added the changelog entries when we actually cut a release, but I can see some advantages to adding changelog entries as we go along for visibility to the community.  Perhaps we should have an "In master, not released" entry at the top whose content we transfer when a release is cut.  Thoughts, anyone?
- @rbrush: You set up the initial contributors page; is there a standard template for Cerner-sponsored open source projects that you copied that from?  Should we have some sort of additional entries besides the "Cerner Corporation" heading?  @dgoeke : To be clear, contributions from the larger community are most welcome and something we encourage but (at least as far as I know) you've never worked for Cerner and the "Cerner Corporation" heading with bullet points underneath it seems to imply otherwise to me.

FYI I'll be otherwise occupied next week so I may not reply to anything here until the week after that.  I can help with the testing on this then if desired.

dgoeke:
RE: Changelog/contributors -- I was trying to follow the CONTRIBUTING.md guidelines and existing patterns there.  I'm happy to remove either or both; my only interest here is having the function not crash.

dgoeke:
The logic around ensuring the data underlying explanations is correct is already tested elsewhere, so I've added a simple test that the function doesn't crash.  This may need to be expanded, but it's a start and will ensure this specific issue doesn't regress.

WilliamParker:
@dgoeke please squash your commits together; we prefer to have each commit in the history reflect a distinct logical change to Clara rather than a step in development.  Otherwise I'm OK with merging this.  I'll leave it open for now in case anyone else wants to look at this, but I'll plan to merge this in the week after next when I'm back and don't object if someone else merges it before then.

rbrush:
Looks good to me, modulo the squash that Will mentioned. Thanks for fixing and contributing this to the project!

dgoeke:
Updated with a squashed commit.  I'm accustomed to just using the github "squash and merge" feature on PRs.

WilliamParker:
Merged.  I logged https://github.com/cerner/clara-rules/issues/350 for future improvements to the tests of explain-activations.  Thanks for the PR @dgoeke.

PR #349, opened 2017-10-11
Add constraint to parser for empty :test conditions

Zylox:
- Previously an empty test condition (aka [:test]) was valid to the
  parse and would generate pointless beta nodes. This change throws an
  exception when parsing that.

This is something we noticed when doing an internal code review for something else and it seemed like an easy enough fix. I can't think of a situation in which we would want an empty test condition. Additionally, it looks like the parser would have intercepted it had`[:test]` been intended as a type anyway, so i dont think this changes anything functionally there.

### DIFF
--- a/src/main/clojure/clara/rules/dsl.clj
+++ b/src/main/clojure/clara/rules/dsl.clj
@@ -136,7 +136,7 @@
 
    (contains? #{'test :test} (first expression))
    (if (= 1 (count expression))
-     (throw-dsl-ex (str "Empty test conditions are not allowed.")
+     (throw-dsl-ex (str "Empty :test conditions are not allowed.")
                    {}
                    expr-meta)
      {:constraints (vec (rest expression))})
--- a/src/main/clojure/clara/rules/dsl.clj
+++ b/src/main/clojure/clara/rules/dsl.clj
@@ -135,7 +135,12 @@
            (parse-expression nested-expr expr-meta)))
 
    (contains? #{'test :test} (first expression))
-   {:constraints (vec (rest expression))}
+   (if (= 1 (count expression))
+     (throw-dsl-ex (str "Empty test conditions are not allowed.")
+                   {}
+                   expr-meta)
+     {:constraints (vec (rest expression))})
+
 
    :default
    (parse-condition-or-accum expression expr-meta)))
--- a/src/test/clojure/clara/test_rules.clj
+++ b/src/test/clojure/clara/test_rules.clj
@@ -1134,15 +1134,14 @@
                (query not-different-temps))))))
 
 (deftest test-empty-test-condition
-
-  (is (thrown-with-msg?
-        clojure.lang.ExceptionInfo
-        #"line.*123.*column.*456"
-        (dsl/parse-query*
-          []
-          [[:test]]
-          {}
-          {:line 123 :column 456}))))
+  (let [exception-data {:line 123 :column 456}]
+    (is (assert-ex-data
+          exception-data
+          (dsl/parse-query*
+            []
+            [[:test]]
+            {}
+            exception-data)))))
 
 (deftest test-multi-insert-retract
 
--- a/src/test/clojure/clara/test_rules.clj
+++ b/src/test/clojure/clara/test_rules.clj
@@ -1133,6 +1133,17 @@
                fire-rules
                (query not-different-temps))))))
 
+(deftest test-empty-test-condition
+
+  (is (thrown-with-msg?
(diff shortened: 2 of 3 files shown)


src/test/clojure/clara/test_rules.clj
```diff
@@ -1133,6 +1133,17 @@
                fire-rules
                (query not-different-temps))))))
 
+(deftest test-empty-test-condition
+
+  (is (thrown-with-msg?
+        clojure.lang.ExceptionInfo
+        #"line.*123.*column.*456"
```
WilliamParker: Is there a particular reason why you didn't use [assert-ex-data](https://github.com/cerner/clara-rules/blob/0.16.0/src/test/clojure/clara/test_rules.clj#L76) here or were you just not aware of it?  It has the advantages of:

- Using normal Clojure data equality semantics rather than messy string comparisons.
- Not failing if something else wraps the exception later since it walks the exception chain to look for a match.

Side note: Clojure has an [IExceptionInfo interface](https://github.com/clojure/clojure/blob/clojure-1.8.0/src/jvm/clojure/lang/IExceptionInfo.java), of which [clojure.lang.ExceptionInfo](https://github.com/clojure/clojure/blob/clojure-1.8.0/src/jvm/clojure/lang/ExceptionInfo.java) is a concrete implementation.  Any instance checks you do should generally use IExceptionInfo for all the usual reasons to favor coding against an interface rather than the implementation; in this particular case my understanding is that user programs might define implementations of IExceptionInfo themselves.  The [ex-data function](http://clojuredocs.org/clojure.core/ex-data) just returns nil if the exception is of the wrong type so that can be used as well in many cases in logic like

````
(when (ex-data e) ...
````

File: src/main/clojure/clara/rules/dsl.clj [hunk 2]
```diff
@@ -135,7 +135,12 @@
            (parse-expression nested-expr expr-meta)))
 
    (contains? #{'test :test} (first expression))
-   {:constraints (vec (rest expression))}
+   (if (= 1 (count expression))
+     (throw-dsl-ex (str "Empty test conditions are not allowed.")
```
WilliamParker: I think this might be clearer with a colon in front of the "test" to denote that it is a keyword, i.e. 

````
"Empty :test conditions are not allowed."
````

I realize that the symbol "test" is allowed as well, but I suspect that :test is the more common usage since that is what the [docs](http://www.clara-rules.org/docs/expressions/#tests) use.

Thoughts?

File: src/test/common/clara/test_dsl.cljc [hunk 3]
```diff
@@ -337,7 +337,6 @@
                           {:r :r4
                            :v :good}])
             (->> (-> empty-session
-                     (insert ^{:type :test} {})
```
WilliamParker: I propose fixing this by changing the fact type inserted here to something other than :test rather than eliminating the LHS of the rule.  My reasoning is that a test for shadowing seems a bit stronger if we force the evaluation to wait until the insert call is made (and not at any time after the call to create the session).  Not a big deal, but it is also an easy change.

src/main/clojure/clara/rules/dsl.clj
```diff
@@ -135,7 +135,12 @@
...
+                   {}
+                   expr-meta)
+     {:constraints (vec (rest expression))})
+
```
WilliamParker: Throw-dsl-ex won't include useful info here like what the offending rule or query is, but that looks like a shortcoming of the current usage of throw-dsl-ex in this namespace that impacts other malformed productions as well, so I think changing that is out of scope here.  I'll log another issue.

WilliamParker:
I made some comments on the changes to the test namespaces, but fundamentally I agree that this is a desirable change.

src/main/clojure/clara/rules/dsl.clj
Zylox: Ya good point

src/test/clojure/clara/test_rules.clj
Zylox: I wasnt aware of it. I copied a test elsewhere in this file and modified it for my needs. I'll work it out to use this.

src/test/common/clara/test_dsl.cljc
Zylox: can do

src/main/clojure/clara/rules/dsl.clj
Zylox: For reference the issue logged is https://github.com/cerner/clara-rules/issues/351

PR #353, opened 2017-10-22
ISSUE-352: Durability fails to deserialize rules containing condition…

EthanEChristian:
…s with an empty constraint

### DIFF
--- a/src/main/clojure/clara/rules/durability/fressian.clj
+++ b/src/main/clojure/clara/rules/durability/fressian.clj
@@ -235,6 +235,23 @@
                 (read [_ rdr tag component-count]
                   (read-with-meta rdr #(apply list %))))}}
 
+   "clj/emptylist"
+   {:class clojure.lang.PersistentList$EmptyList
+    :writer (reify WriteHandler
+              (write [_ w o]
+                (let [m (meta o)]
+                  (do
+                    (.writeTag w "clj/emptylist" 1)
+                    (if m
+                      (.writeObject w m)
+                      (.writeNull w))))))
+    :readers {"clj/emptylist"
+              (reify ReadHandler
+                (read [_ rdr tag component-count]
+                  (let [m (read-meta rdr)]
+                    (cond-> '()
+                            m (with-meta m)))))}}
+
    "clj/aseq"
    {:class clojure.lang.ASeq
     :writer (reify WriteHandler
@@ -264,7 +281,7 @@
               (reify ReadHandler
                 (read [_ rdr tag component-count]
                   (read-with-meta rdr #(into {} %))))}}
-   
+
    "clj/treeset"
    {:class clojure.lang.PersistentTreeSet
     :writer (reify WriteHandler
@@ -289,7 +306,7 @@
                     (if m
                       (with-meta s m)
                       s))))}}
-   
+
    "clj/treemap"
    {:class clojure.lang.PersistentTreeMap
     :writer (reify WriteHandler
@@ -347,7 +364,7 @@
                   (let [s (symbol (.readObject rdr) (.readObject rdr))
                         m (read-meta rdr)]
                     (cond-> s
-                      m (with-meta m)))))}}
+                            m (with-meta m)))))}}
 
    "clj/record"
    {:class clojure.lang.IRecord
--- a/src/test/clojure/clara/durability_rules.clj
+++ b/src/test/clojure/clara/durability_rules.clj
@@ -29,7 +29,12 @@
 
 (defrule find-wind-speeds-without-temp
   "Rule using NegationNode"
-  [?w <- WindSpeed (= ?loc location)]
+  [?w <- WindSpeed
+   ;; Empty constraint and a constraint containing an empty list to test serializing an EmptyList,
+   ;; see Issue 352 for more information
+   ()
+   (not= this ())
+   (= ?loc location)]
   [:not [Temperature (= ?loc location)]]
   =>
   (insert! (->UnpairedWindSpeed ?w)))
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -259,8 +259,8 @@
          (throw (ex-info (str "Malformed variable binding for " variables ". No associated value.")
                          {:variables (map keyword variables)})))
 
-       (if binds-variables?
-
+       (cond
+         binds-variables?
          ;; Bind each variable with the first value we encounter.
          ;; The additional equality checks are handled below so which value
          ;; we bind to is not important. So an expression like (= ?x value-1 value-2) will
@@ -287,8 +287,14 @@
                compiled-rest)
             )
 
+         ;; A contraint that is empty doesn't need to be added as a check,
+         ;; simply move on to the rest
+         (empty? exp)
+         compiled-rest
+
          ;; No variables to unify, so simply check the expression and
          ;; move on to the rest.
+         :else
          `(if ~exp ~compiled-rest nil))))))
 
 
--- a/src/test/common/clara/test_dsl.cljc
+++ b/src/test/common/clara/test_dsl.cljc
@@ -10,7 +10,8 @@
                                     query]]
                [clara.rules.accumulators :as acc]
                [clara.rules.testfacts :refer [->Temperature ->Cold ->WindSpeed
-                                              ->ColdAndWindy map->FlexibleFields]]
+                                              ->ColdAndWindy map->FlexibleFields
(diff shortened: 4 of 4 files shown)


EthanEChristian:
Issue 352

EthanEChristian:
@mrrodriguez @WilliamParker

src/test/clojure/clara/durability_rules.clj
```diff
@@ -29,7 +29,10 @@
 
 (defrule find-wind-speeds-without-temp
   "Rule using NegationNode"
-  [?w <- WindSpeed (= ?loc location)]
+  [?w <- WindSpeed
+   ;; Empty constraint to test the serializing EmptyList, see Issue 352 for more information
+   ()
```
EthanEChristian: I'm not sure that its right to accept this, however removing it as a valid constraint would probably be seen as non-passive

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -235,6 +235,16 @@
                 (read [_ rdr tag component-count]
                   (read-with-meta rdr #(apply list %))))}}
 
+   "clj/emptylist"
+   {:class clojure.lang.PersistentList$EmptyList
```
mrrodriguez: This is a sneaky case. The case for `clojure.lang.ASeq` is meant to catch the Clojure collection types not otherwise specified here, but  `clojure.lang.PersistentList$EmptyList` is a special inner class that apparently only implements the interfaces, which differs from how the mos to the other Clojure data structures are implemented.

However, it is concerning. Classes like this are really "implementation details" of Clojure. I don't know that there is a good way around this though, other than to keep it consistent with Clara's supported versions of Clojure.

One, potentially obvious, sounding fix for this sort of issue would be to implement readers and writers against the interfaces instead of these concrete classes. However, this can lead to issues with multiple handlers matching for some objects, which will fail. This is similar to the multiple inheritance sort of issues that happen with protocol extensions. Many concrete collection types will implement many common interfaces. It tends to cause problems quickly. The recommendation I've seen is to only target extending concrete classes instead of interfaces. (Technically the "lookup" mechanism of Fressian is extensible and could be used to implement something similar to multimethod "prefers", but that still gets complicated and brittle.)

The downside, is that we have to be sure to cover all possible types that Clara may be serializing in its own network and working memory structures.

I don't see a better way right now other than doing what you did here and just adding the classes. This made me look around in Clojure to recall if there are any other special classes used to represent "empty" collections. It doesn't see to be the case for vectors, sets, or maps. Clojure eventually backed off an earlier [approach to using different classes for certain "smaller" collection sizes](https://dev.clojure.org/jira/browse/CLJ-1517) too, so that shouldn't be a problem.

src/test/clojure/clara/durability_rules.clj
mrrodriguez: I think it is a bit odd to have the empty constraint along with something else. However, I agree that it doesn't really make sense to make it start failing now. It seems like something that could be done by people in practice fairly easily or maybe even intentionally for some visual preference.

Also, the empty list is "truthy" in Clojure, so I'd expect it to be evaluated that way.

EthanEChristian: Yes this is an odd scenario, however i wanted to minimize the change to the rules that we use to test serialization, because one small change can vastly change the tests

src/main/clojure/clara/rules/durability/fressian.clj
WilliamParker: Yeah, as far as I can tell there is no consistent convention or API in Clojure that would allow us to serialize and deserialize these collections while maintaining their concrete class without special-casing each concrete type.  It is admittedly unfortunate that something like [Java's collection constructor](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#ArrayList-java.util.Collection-) isn't available for Clojure's data structures.

src/test/common/clara/test_simple_rules.cljc
```diff
@@ -194,3 +194,14 @@
     (is (has-fact? (:cold @side-effect-holder) (->Temperature -10 "MCI") ))
 
     (is (has-fact? (:subzero @side-effect-holder) (->Temperature -10 "MCI")))))
+
+(def-rules-test test-condition-with-empty-constraint
+  {:rules [simple-rule [[[Cold ()]]
```
WilliamParker: Can we put the empty list somewhere also where serialization of it will be required even if the empty constraint is filtered out?  Say [Cold () (not= temperature ()]

EthanEChristian: Added this to the basic and deserialization test

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -235,6 +235,16 @@
...
+    :writer (reify WriteHandler
+              (write [_ w o]
+                (write-with-meta w "clj/emptylist" o)))
```
WilliamParker: Since this type is, by definition, empty, we don't actually need to write out or read in the list itself - we could just write out the metadata if present (or null if absent).  See the implementation of write-with-meta and read-with-meta.

src/test/common/clara/test_simple_rules.cljc
```diff
@@ -194,3 +194,14 @@
     (is (has-fact? (:cold @side-effect-holder) (->Temperature -10 "MCI") ))
 
     (is (has-fact? (:subzero @side-effect-holder) (->Temperature -10 "MCI")))))
+
+(def-rules-test test-condition-with-empty-constraint
+  {:rules [simple-rule [[[Cold () (not= this ())]]
```
WilliamParker: This is good, but looking over compile-constraints I think we should have a test where we validate that a rule with only empty constraints, preferably for both the case of 1 empty constraint and 2 empty constraints, works correctly.  The case of no constraints (also a recursion termination case here) is specifically handled with this:

````
(if (empty? exp-seq)
     `(deref ~'?__bindings__)
````

but I'd like to make sure that we're testing the recursion properly both when there are later non-empty constraints and when we bottom out without finding any meaningful constraints.

File: src/test/common/clara/test_simple_rules.cljc [hunk 6]
```diff
@@ -194,3 +194,14 @@
     (is (has-fact? (:cold @side-effect-holder) (->Temperature -10 "MCI") ))
 
     (is (has-fact? (:subzero @side-effect-holder) (->Temperature -10 "MCI")))))
```
WilliamParker: This doesn't really seem like a "simple rule" sanity test to me.  I suppose we could put it here, but IMO it fits better in test-dsl since it revolves around edge cases of ways to represent the logical relationships implied by rules, not edge cases of the logical relationships themselves.  Test-dsl is a cross-platform test file as well.

src/test/clojure/clara/durability_rules.clj
```diff
@@ -29,7 +29,12 @@
 
 (defrule find-wind-speeds-without-temp
   "Rule using NegationNode"
-  [?w <- WindSpeed (= ?loc location)]
+  [?w <- WindSpeed
+   ;; Empty constraint and a constraint containing an empty list to test the serializing EmptyList,
```
WilliamParker: minor grammar: "the serializing" => "serializing the"

WilliamParker:
@EthanEChristian the idea looks good, I left some comments with a few suggestions.

EthanEChristian:
I have address the concerns you mentioned @WilliamParker

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -235,6 +235,23 @@
                 (read [_ rdr tag component-count]
                   (read-with-meta rdr #(apply list %))))}}
 
+   "clj/emptylist"
+   {:class clojure.lang.PersistentList$EmptyList
+    :writer (reify WriteHandler
+              (write [_ w o]
+                (let [m (meta o)]
+                  (do
+                    (.writeTag w "clj/emptylist" 2)
```
WilliamParker: Now that there's only one component written out this should be 1, not 2.  See the [Fressian custom handler documentation](https://github.com/clojure/data.fressian/wiki/Creating-custom-handlers); the implementation of this method in [FressianWriter](https://github.com/Datomic/fressian/blob/fressian-0.6.6/src/org/fressian/FressianWriter.java#L252) might be good to glance at as well.

WilliamParker:
@EthanEChristian looks good to me apart from https://github.com/cerner/clara-rules/pull/353/files#r147421772

EthanEChristian:
+1 good catch, updated

PR #358, opened 2017-11-16
Set up Travis to build with the most recent Clojure and ClojureScript…

WilliamParker:
… versions as well as the minimum required versions

I may still modify/clean this up some later but it seems to work now and I wanted to go ahead and raise visibility in case anyone else is looking at Clara in Clojure 1.9 as well.  There were some incorrect ns forms in the tests that spec fails in Clojure 1.9.0-RC1 but otherwise it appears everything build properly.  The problems being limited to tests probably explains why nobody has reported problems with Clara and Clojure 1.0 yet.


project.clj
```diff
@@ -7,11 +7,17 @@
...
+             :recent-clj {:dependencies [^:replace [org.clojure/clojure "1.9.0-RC1"]
+                                         ^:replace [org.clojure/clojurescript "1.9.946"]]}}
+  :plugins [[lein-codox "0.10.3" :exclusions [org.clojure/clojure
+                                              org.clojure/clojurescript]]
+            [lein-javadoc "0.3.0" :exclusions [org.clojure/clojure
+                                               org.clojure/clojurescript]]
+            [lein-cljsbuild "1.1.7" :exclusions [org.clojure/clojure
+                                                 org.clojure/clojurescript]]
+            [lein-figwheel "0.5.14" :exclusions [org.clojure/clojure
+                                                 org.clojure/clojurescript]]]
```
WilliamParker: I should probably investigate what exclusions here are actually needed.  I just added blanket ClojureScript exclusions for all plugins.

Some old versions of the plugins had ns forms that don't compile with 1.9.  Anyway, I don't see any reason not to update these plugins to the most recent versions since I don't expect that to impact users.

src/test/common/clara/test_common.cljc
```diff
@@ -25,7 +25,8 @@
   []
   [:temperature [{temperature :temperature}] (< temperature 20) (= ?t temperature)])
 
-(defsession my-session [test-rule cold-query] :fact-type-fn :type)
+(defsession my-session [test-rule
+                        cold-query] :fact-type-fn :type)
```
WilliamParker: The changes here are because on my machine I have some weird compilation issues around this line that aren't consistently reproducible and don't show up in Travis, and I was experimenting to see if anything changed that.  I can probably remove these changes.

.travis.yml
```diff
@@ -3,7 +3,8 @@ lein: 2.7.1
 script:
   - lein test
   - lein test :generative
+  - lein with-profile dev,recent-clj test
 # Workaround for https://github.com/travis-ci/travis-ci/issues/4691
 sudo: required
 jdk:
-  - openjdk7
+  - openjdk8
```
WilliamParker: The most recent ClojureScript seemed to need Java 8 to work in Travis where the previous didn't, since the recent-clj build failed with openjdk7 and the main build succeeded.  I haven't yet investigated what changed.

project.clj
WilliamParker: Actually on reflection this won't do any harm so I think I'll leave it.

WilliamParker:
I've pushed a bump to the released Clojure 1.9 version.  I plan to merge this soon unless there are objections.  If there is any further work we want to do here it will be easy enough to change later but I'd like to go ahead and get testing on recent Clojure/ClojureScript versions working (in addition to the tests on the existing versions of the same, this does not raise Clara's minimum version of either).

.travis.yml
mrrodriguez: It may just be due to some dev time stuff in cljs. It requires a jdk7 minimum according to docs. However, for some extra jdk nashorn features it uses jdk8. I'm not sure the specifics.

project.clj
mrrodriguez: Yeah, I was planning to give a PR soon that refactors some of this cljs setup. There are a few things I think that could be improved. I'm not sure adding extra `:exclusions` really matters. However, you shouldn't need to exclude transitive dependencies too often if you are explicitly declaring the version of those deps you want anyways (in JDK space at least).

PR #360, opened 2017-11-27
ISSUE-359: Errors when session, rules, and fact types are defined in different namespaces and not explicitly referred by name.

### LINKED ISSUE #359: Errors when session, rules, and fact types are defined in different namespaces and not explicitly referred by name
I've augmented the tests to illustrate the issue. Suppose I have the following two namespace definitions:

```clojure
(ns clara.rule-defs
  (:require-macros [cljs.test :refer (is deftest run-tests testing)]
                   [clara.test-rules-data])
  (:require [cljs.test :as t]
            [clara.rules.engine :as eng]
            [clara.rules.accumulators :as acc]
            [clara.rules :refer [insert retract fire-rules query insert!]
             :refer-macros [defrule defsession defquery]]
            [clara.rules.testfacts :as tf]))

(def simple-defrule-side-effect (atom nil))
(def other-defrule-side-effect (atom nil))

(defrule test-rule
         [tf/Temperature (< temperature 20)]
         =>
         (reset! other-defrule-side-effect ?__token__)
         (reset! simple-defrule-side-effect ?__token__))

(defquery cold-query
          []
          [tf/Temperature (< temperature 20) (== ?t temperature)])

;; Accumulator for getting the lowest temperature.
(def lowest-temp (acc/min :temperature))

(defquery coldest-query
          []
          [?t <- lowest-temp :from [tf/Temperature]])

(defrule is-cold-and-windy
         "Rule to determine whether it is indeed cold and windy."

         (tf/Temperature (< temperature 20) (== ?t temperature))
         (tf/WindSpeed (> windspeed 30) (== ?w windspeed))
         =>
         (insert! (tf/->ColdAndWindy ?t ?w)))

(defquery find-cold-and-windy
          []
          [?fact <- tf/ColdAndWindy])
```
```clojure
(ns clara.test-rules-require
  (:require-macros [cljs.test :refer (is deftest run-tests testing)]
                   [clara.test-rules-data])
  (:require [cljs.test :as t]
            [clara.rules.engine :as eng]
            [clara.rules.accumulators :as acc]
            [clara.rules :refer [insert retract fire-rules query insert!]
             :refer-macros [defrule defsession defquery]]
            [clara.rule-defs :as rd]
            [clara.rules.testfacts :as facts]))

(comment
  ;; Launch browser repl.
  (cemerick.piggieback/cljs-repl :repl-env (cemerick.austin/exec-env))
  )

(defn- has-fact? [token fact]
       (some #{fact} (map first (:matches token))))

(defsession my-session 'clara.rule-defs)
(defsession my-session-map 'clara.rule-defs :fact-type-fn :type)

(deftest test-simple-defrule
         (let [session (insert my-session (facts/->Temperature 10 "MCI"))]

              (fire-rules session)

              (is (has-fact? @rd/simple-defrule-side-effect (facts/->Temperature 10 "MCI")))
              (is (has-fact? @rd/other-defrule-side-effect (facts/->Temperature 10 "MCI")))))

(deftest test-simple-query
         (let [session (-> my-session
                           (insert (facts/->Temperature 15 "MCI"))
                           (insert (facts/->Temperature 10 "MCI"))
                           (insert (facts/->Temperature 80 "MCI"))
                           fire-rules)]

              ;; The query should identify all items that wer einserted and matchd the
              ;; expected criteria.
              (is (= #{{:?t 15} {:?t 10}}
                     (set (query session rd/cold-query))))))

(deftest test-simple-accumulator
         (let [session (-> my-session
                           (insert (facts/->Temperature 15 "MCI"))
                           (insert (facts/->Temperature 10 "MCI"))
                           (insert (facts/->Temperature 80 "MCI"))
                           fire-rules)]

              ;; Accumulator returns the lowest value.
              (is (= #{{:?t 10}}
                     (set (query session rd/coldest-query))))))

(deftest test-simple-insert
         (let [session (-> my-session
                           (insert (facts/->Temperature 15 "MCI"))
                           (insert (facts/->WindSpeed 45 "MCI"))
                           (fire-rules))]

              (is (= #{{:?fact (facts/->ColdAndWindy 15 45)}}
                     (set
                       (query session rd/find-cold-and-windy))))))
```

When executed with `lein test`, the compiler gives these warnings:

> WARNING: Use of undeclared Var clara.test-rules-require/other-defrule-side-effect at line 17 src\test\clojurescript\clara\test_rules_require.cljs
> WARNING: Use of undeclared Var clara.test-rules-require/simple-defrule-side-effect at line 18 src\test\clojurescript\clara\test_rules_require.cljs
> WARNING: No such namespace: tf, could not locate tf.cljs, tf.cljc, or Closure namespace "" at line 37 src\test\clojurescript\clara\test_rules_require.cljs
> WARNING: Use of undeclared Var tf/->ColdAndWindy at line 37 src\test\clojurescript\clara\test_rules_require.cljs
> WARNING: Use of undeclared Var clara.test-rules-require/other-defrule-side-effect at line 17 src\test\clojurescript\clara\test_rules_require.cljs
> WARNING: Use of undeclared Var clara.test-rules-require/simple-defrule-side-effect at line 18 src\test\clojurescript\clara\test_rules_require.cljs
> WARNING: No such namespace: tf, could not locate tf.cljs, tf.cljc, or Closure namespace "" at line 37 src\test\clojurescript\clara\test_rules_require.cljs
> WARNING: Use of undeclared Var tf/->ColdAndWindy at line 37 src\test\clojurescript\clara\test_rules_require.cljs

Along with several runtime errors such as

> ERROR in (test-simple-accumulator) (:)
> Uncaught exception, not in assertion.
> expected: nil
>   actual: #error {:message "Exception in clara.rule-defs/test-rule with bindings {}", :data {:bindings {}, :name "clara.rule-defs/test-rule", :rhs (do (reset! other-defrule-side-effect ?__token__) (reset! simple-defrule-side-effect ?__token__)), :batched-logical-insertions [], :batched-unconditional-insertions [], :batched-rhs-retractions [], :listeners []}, :cause #object[Error Error: No protocol method IReset.-reset! defined for type undefined: ]}

> ERROR in (test-simple-insert) (:)
> Uncaught exception, not in assertion.
> expected: nil
>   actual: #error {:message "Exception in clara.rule-defs/is-cold-and-windy with bindings {:?t 15, :?w 45}", :data {:bindings {:?t 15, :?w 45}, :name "clara.rule-defs/is-cold-and-windy", :rhs (do (insert! (tf/->ColdAndWindy ?t ?w))), :batched-logical-insertions [], :batched-unconditional-insertions [], :batched-rhs-retractions [], :listeners []}, :cause #object[ReferenceError ReferenceError: Can't find variable: tf]}

This issue is possibly related to #327.

sparkofreason:
Fixes #359.


src/main/clojure/clara/macros.clj
```diff
@@ -180,7 +180,8 @@
               (com/compile-action all-bindings
                                   ;; Using private function for now as a workaround.
                                   (if (:ns-name production)
-                                    (binding [*ns* (the-ns (:ns-name production))]
+                                    (binding [*ns* (the-ns (:ns-name production))
+                                              cljs.analyzer/*cljs-ns* (:ns-name production)]
```
WilliamParker: [Clojure's the-ns](http://clojuredocs.org/clojure.core/the-ns) throws an exception if the namespace isn't found.  I was very surprised that this didn't fail, since I expected the test rules ClojureScript namespace to not exist in the Clojure environment of the ClojureScript compiler.  I put in some debug statements and it appears that the ClojureScript compiler creates an empty Clojure namespace of the same name, at least in that case.  I dug around a bit and found some calls like this in Clojure in the [ClojureScript analyzer](https://github.com/clojure/clojurescript/blob/master/src/main/clojure/cljs/analyzer.cljc):

````
(create-ns *cljs-ns*)
````

I haven't dug around in the ClojureScript compiler to understand why it would do this, but it seems like an implementation detail that could break.  Since direct use of this namespace is deprecated in favor of use through rules.cljc, we probably won't have usage of it in Clojure; see [defsession in clara.rules/defsession](https://github.com/cerner/clara-rules/blob/0.16.1/src/main/clojure/clara/rules.cljc#L357) and its switching on whether Clojure or ClojureScript is being compiled.  Nevertheless it is easy enough to support both here in a safe way that my inclination is to keep doing so for the moment pending a more complete discussion of fully "unsupporting" this namespace in Clojure.

That is:

````
(if (:ns-name production)
     (if (compiling-cljs?)
         (binding [cljs.analyzer/*cljs-ns* (:ns-name-production)]
                 (resolve-vars)
         (binding [*ns* (the-ns (:ns-name production))]
                 (resolve-vars))
   .....
````

File: src/test/clojurescript/clara/test_rules_require.cljs [hunk 2]
```diff
@@ -0,0 +1,63 @@
...
+            [clara.rules.accumulators :as acc]
+            [clara.rules :refer [insert retract fire-rules query insert!]
+             :refer-macros [defrule defsession defquery]]
+            [clara.rule-defs :as rd]
+            [clara.rules.testfacts :as facts]))
+
+(comment
+  ;; Launch browser repl.
+  (cemerick.piggieback/cljs-repl :repl-env (cemerick.austin/exec-env))
+  )
```
WilliamParker: Did you mean to remove this piggieback call before merging?

src/test/clojurescript/clara/rule_defs.cljs
```diff
@@ -0,0 +1,42 @@
+(ns clara.rule-defs
+  (:require-macros [cljs.test :refer (is deftest run-tests testing)]
+                   [clara.test-rules-data])
+  (:require [cljs.test :as t]
+            [clara.rules.engine :as eng]
```
WilliamParker: I don't see any direct usage of the clara.rules.engine in this namespace, so I think this require can be removed.  I wouldn't expect such usage since that namespace isn't intended to be a public API.

src/test/clojurescript/clara/test_rules_require.cljs
```diff
@@ -0,0 +1,63 @@
+(ns clara.test-rules-require
```
WilliamParker: Could you add comment in this namespace and the test-rules-data namespace explaining why these namespaces exist, including a link to the GitHub issue?  Perhaps one comment could simply refer to the other to avoid content duplication.

src/test/clojurescript/clara/test_rules_require.cljs
```diff
@@ -0,0 +1,63 @@
...
+(deftest test-simple-insert
+         (let [session (-> my-session
+                           (insert (facts/->Temperature 15 "MCI"))
+                           (insert (facts/->WindSpeed 45 "MCI"))
+                           (fire-rules))]
+
+              (is (= #{{:?fact (facts/->ColdAndWindy 15 45)}}
+                     (set
+                       (query session rd/find-cold-and-windy))))))
+
```
WilliamParker: I don't want to block on this, particularly since this is a problem specific to ClojureScript, but when functionality isn't platform-specific, even if it came up on one platform (JVM vs JS) the use of [def-rules-test](https://github.com/cerner/clara-rules/blob/0.16.1/src/main/clojure/clara/tools/testing_utils.cljc#L17) to create shared tests between Clojure and ClojureScript is preferred.  There are a variety of examples under [test-common](https://github.com/cerner/clara-rules/tree/0.16.1/src/test/common/clara).  I've been gradually working to get the test suite as cross-platform as possible, which has the primary effect of improving test coverage in ClojureScript.  See [this project board](https://github.com/cerner/clara-rules/projects/1).  Not critical to merging this, but a good thing to keep in mind for any future contributions.   The "contributors" readme should probably mention this transition.

src/test/clojurescript/clara/test_rules_require.cljs
sparkofreason: Done

src/test/clojurescript/clara/test_rules_require.cljs
sparkofreason: I'll convert these tests after I get some experience updating the tests on #361.

WilliamParker: Hmm.. so looking at this again it seems like using def-rules-test would defeat the point of testing the real path of verifying how defrule/defquery work with namespace qualification.  I don't believe I was thinking clearly/paying enough attention there.  Perhaps these could just be cljc tests in the common directory, but look more or less the same as now, just with any necessary changes to the ns header?

sparkofreason: Moved these tests to common as you suggested, but interestingly the clj version fails. Need to dig in.

sparkofreason:
Similar issue to that seen in #361, where there's a difference in how the default fact-types work in CLJ and CLJS. CLJS "types" are just JS constructor functions, which means they can be referenced as a namespace-qualified symbol, with an ns alias, etc. just like any other function. In CLJ, the type really is a type, ultimately being a Java class name. So the code in `clara.rule-defs` which referenced facts like `tf/Temperature` work fine in CLJS, but go haywire in CLJ. As with #361, solution was to use a reader conditional like `#?(:clj Temperature :cljs tf/Temperature)`.

I'm wondering if these differences between CLJ and CLJS could be unified? The only thing I can think is to have the rule DSL work with symbols instead of type references, but that feels like it would touch a lot of code.

sparkofreason:
FWIW, the other place the CLJ/CLJS difference in fact-types rears its head is in the reload scenario. If fact-types are defined in their own ns, when they get reloaded in CLJS they now refer to a different instance of the constructor function. So if any rule def namespaces use those facts and are not reloaded, the rules will not fire when facts are inserted.

src/test/common/clara/rule_defs.cljc
```diff
@@ -0,0 +1,42 @@
...
+;; Rule definitions used for tests in clara.test-rules-require.
+
+(def simple-defrule-side-effect (atom nil))
+(def other-defrule-side-effect (atom nil))
+
+(defrule test-rule
+           [#?(:clj Temperature :cljs tf/Temperature) (< temperature 20)]
+           =>
+           (reset! other-defrule-side-effect ?__token__)
+           (reset! simple-defrule-side-effect ?__token__))
```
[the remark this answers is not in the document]
sparkofreason: I just copied these tests from existing tests, and broke up the rule and session definitions into separate namespaces. I thought the `?__token__` things was testing something special, but it looks like it's just an alternate way of getting fact binding? I'll just switch it to a regular fact binding.

src/test/common/clara/rule_defs.cljc
```diff
@@ -0,0 +1,42 @@
...
+            [clara.rules.testfacts :as tf]
+    #?(:clj [clara.rules :refer [defrule defquery insert!]])
+    #?(:cljs [clara.rules :refer-macros [defrule defquery] :refer [insert!]]))
+  #?(:clj
+     (:import [clara.rules.testfacts Temperature WindSpeed ColdAndWindy])))
+
+;; Rule definitions used for tests in clara.test-rules-require.
+
+(def simple-defrule-side-effect (atom nil))
+(def other-defrule-side-effect (atom nil))
```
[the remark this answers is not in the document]
sparkofreason: This also came from copying from `clara.test-rules`. I'll change it to work with `clara.tools.test-utils`. I don't think we need multiple atoms for this particular test anyway.

WilliamParker:
This looks good to me now at @sparkofreason 

I plan to leave this open until Thursday morning in the US in case anyone else in the community wants to look this over, but plan to merge it then unless anyone has concerns or requests a delay, in which case we can discuss it.

Issue #307, opened 2017-06-13
Support session inspection and fact-graph in ClojureScript


WilliamParker:
[Session inspection](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/tools/inspect.clj) and the [fact-graph](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/tools/fact_graph.clj) functionality added in https://github.com/cerner/clara-rules/issues/277 are currently Clojure-only, but for the most part I don't think they need to be, although [the :condition-matches are generated using Clojure-only functionality](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/tools/inspect.clj#L181).  In particular, I don't see any dependence of [gen-fact->explanations](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/tools/inspect.clj#L114) on Clojure-only functionality, and this what the fact-graph functionality uses.

In addition to being useful for ClojureScript users, this enhancement would make it much easier to prototype visualizations using fact graphs in a browser or other JS-based environment since the problem of serializing a JVM fact graph, making it available via a web server, etc. could be put off.

alex-dixon:
@WilliamParker Not seeing how `:condition-matches` are CLJ-only. When you have time if you'd be able to elaborate this might be something I could help with.

WilliamParker:
@alex-dixon the key line is [here](https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/tools/inspect.clj#L161).  Basically it takes the productions (rules and queries) in the session, invokes a function in the Clara compiler on them, and then uses the result of that computation in the compiler to [create the condition matches]( https://github.com/cerner/clara-rules/blob/0.15.0/src/main/clojure/clara/tools/inspect.clj#L181).  Since the Clara compiler is Clojure-only a straight port of that to ClojureScript wouldn't work.  Note that the Clojure-only compiler is invoked as part of macroexpansion in ClojureScript.  Perhaps this could be implemented in a more portable way though; I just haven't had time to think about it yet.  Ideally we wouldn't need to bring the compiler into session inspection anyway IMO.  I'm certainly open to proposals on how to accomplish this.

alex-dixon:
[...]

PR #361, opened 2017-11-28
ISSUE-307: Support session inspection and fact-graph in ClojureScript

sparkofreason:
Convert the clara.rules.schema, clara.tools.inspect, and clara.tools.fact-graph namespaces to CLJC. Provides support for session inspection and fact-graph in ClojureScript (#307).


WilliamParker:
The PR looks like what I'd expect, but I'd like to have at least a small sampling of the [session inspection tests](https://github.com/cerner/clara-rules/blob/0.16.1/src/test/clojure/clara/tools/test_inspect.clj) and [fact graph tests](https://github.com/cerner/clara-rules/blob/0.16.1/src/test/clojure/clara/tools/test_fact_graph.clj) uplifted to be shared between Clojure and ClojureScript by using [def-rules-test](https://github.com/cerner/clara-rules/blob/0.16.1/src/main/clojure/clara/tools/testing_utils.cljc#L17); you can find example usage in the common tests directory.  I realize that uplifting all of the tests would be time-consuming, but I would like to have some validation that this isn't just totally broken in ClojureScript in our tests if we are going to say that inspection and fact-graph support it.

sparkofreason:
I can update the tests, however `def-rules-test` does not currently correctly initialize the rule name, which some of the tests require. The fix is easy. Should I just include it in the PR for this issue?

WilliamParker:
Ah yes, I overlooked that @sparkofreason ; [defrule adds the name after calling dsl/parse-rule*](https://github.com/cerner/clara-rules/blob/0.16.1/src/main/clojure/clara/rules.cljc#L389) but [def-rules-test doesn't do the same](https://github.com/cerner/clara-rules/blob/0.16.1/src/main/clojure/clara/tools/testing_utils.cljc#L37).  I assume you're planning to just assoc the name onto the productions in def-rules-test in the same way that defrule/defquery do?  That fix can go in the PR for this issue.  Thanks.

sparkofreason:
Looks like I'll also need to modify `def-test-rules` to expose the rule and query names as part of the let-binding.

sparkofreason:
Ported all the tests to common, everything passes. Realized after I pushed that cursive reformatted some things, let me know if you want me to reformat to be closer to what it was before. The `clara.rules.testfacts` namespace had `:refer :all` in the original `clara.tools.test-inspect` file, and to make it work in cljs I wound up with a `:refer` that's a little out of hand. Thinking now it would be better to just use a namespace alias.

CONTRIBUTORS.md
```diff
@@ -16,3 +16,4 @@ Community
 [@EthanEChristian]: https://github.com/EthanEChristian
 [@kulkarnipushkar]: https://github.com/kulkarnipushkar
 [@dgoeke]: https://github.com/dgoeke
+[@sparkofreason]: https://github.com/sparkofreason
```
WilliamParker: I think you need to add your name to the list above for it to render in GitHub. The list here at the bottom looks to be a store of handles to links that isn't actually in the rendered markdown except in that it is joined with the rendered sections to create links.  I didn't find an entry for this sort of thing in the [GitHub markdown links section](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links) but that appears to be what's going on (and when I made the change locally Atom rendered it as expected).

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 This is a history of changes to clara-rules.
 
+### 0.17.0
```
WilliamParker: Minor thing I missed before, but we should probably label this "0.17.0-SNAPSHOT" to avoid giving the impression that we've released 0.17.0.  It will then be changed when we do a release of course.

WilliamParker:
@sparkofreason Looks good apart from https://github.com/cerner/clara-rules/pull/361#discussion_r154379866 in the testing-utils ns and some minor/easy fixes on the changelog and contributors page.  After that this should be ready to merge.

Regarding the reformatting, I'm not too worried about that personally, especially since the biggest change in terms of LOC is deletion/creation of test namespaces associated with moving to cljc.  Ideally we'd have some kind of editor-independent linter e.g. a Leiningen plugin that would take care of things like whitespace formatting, but it isn't something I've spent time looking into yet.

sparkofreason:
Ack - I forgot to add the commonized tests to clara.test to get the CLJS versions to run. They are not passing, will need to do some more work on this.

sparkofreason:
A couple of things had to be fixed. The use of `(.startsWIth <string>)` does not work in CLJS, replaced with `goog.string/startsWith` since clara is still on clojure 1.7. I think that can be cleaned up using `clojure.string/starts-with?` when the clojure version gets bumped. Also had to explicitly refer to fact types using the quoted symbol in CLJS.

src/main/clojure/clara/tools/inspect.cljc
```diff
@@ -4,16 +4,24 @@
...
-            NegationWithJoinFilterNode]))
+  #?(:clj
+    (:import [clara.rules.engine
+              RootJoinNode
+              HashJoinNode
+              ExpressionJoinNode
+              NegationNode
+              NegationWithJoinFilterNode]))
+  #?(:cljs
+     (:import goog.string)))
```
mrrodriguez: Although this tends to work, the proper cljs way to load in a closure library is via `:require`.
Along with that, I'd recommend a ns alias to refer to symbols from it.
```clj
(:require [goog.string :as gstr])

;; later

(gstr/startsWith (name k) "?__gen__")
```

See http://clojurescriptmadeeasy.com/blog/when-do-i-use-require-vs-import.html

src/main/clojure/clara/tools/testing_utils.cljc
```diff
@@ -62,14 +62,15 @@
                                                                                                        cat
                                                                                                        session-opts)))]))
                                                         cat)))
-           
+
            test-form `(~(if (com/compiling-cljs?)
```
mrrodriguez: I almost commented here that this `if` case is unnecessary since in [CLJS v1.9.198](https://github.com/clojure/clojurescript/blob/master/changes.md#19198) you can refer to namespaces via `clojure.*` that do not exists, but do exists in `cljs.*` and it will automatically resolve it. This was done in https://dev.clojure.org/jira/browse/CLJS-1692

However, then I realized that Clara is building against a really old CLJS now (by CLJS standards). No action to take here and I'm not saying we should move all the way up to a 1.9.198 minimum. However, I did log #363 when I realized this situation.

src/main/clojure/clara/tools/inspect.cljc
mrrodriguez: @sparkofreason 

The issue I logged at #363 is related to this, but on the CLJ-side of things instead:

Clara is intentionally building and supporting an older version of CLJ in order to just have broader reach. It is unfortunate though when it makes certain features not exists yet, like `clojure.string/starts-with?` (along with probably quite a bit of fn's designed to make writing CLJC easier/less-host-dependent/with less read-conditionals.
However, I do think it'd be good to have profiles for newer versions of CLJ to ensure Clara is building and passing tests against them.

src/test/common/clara/tools/test_fact_graph.cljc
```diff
@@ -0,0 +1,141 @@
+(ns clara.tools.test-fact-graph
+  (:require
+    #?(:clj [clara.tools.testing-utils :refer [def-rules-test
+                                               side-effect-holder] :as tu])
+    #?(:cljs [clara.tools.testing-utils :refer [side-effect-holder]
+              :refer-macros [def-rules-test] :as tu])
```
mrrodriguez: In a more modern cljs version ([v1.9.198 specifically](https://github.com/clojure/clojurescript/blob/master/changes.md#19198))This namespace shouldn't have to distinguish between `:refer-macros` and `:refer`. However, I think that the ns `clara.tools.testing-utils` needs to require it's own macros though in it's `:cljs` branch though to allow for this implicit macro inference to kick in (a practice we should use in all relevant places in this codebase).
e.g.
```clj
:cljs
(ns clara.tools.testing-utils
  (:require [clara.rules.update-cache.core :as uc])
  (:require-macros clara.tools.testing-utils))
```

Since Clara is still supporting an old version of CLJS (see #363 ), I think we should stick to using ns aliases to refer to all symbols. Macro inference in CLJS has been supported for a long time for aliases I believe.

So:
```clj
#?(:cljs [clara.tools.testing-utils :as tu])

;; later
(tu/def-rules-test <etc>)
```

This applies to the `clojure.test` vs `cljs.test` stuff too, and maybe more places.

I think these `ns` declarations are getting a bit heavy and just trying to simplify before things get too messy.

This is mostly a suggestion though. If you are strongly opposed to doing it in this PR, I could understand that.

src/test/common/clara/tools/test_inspect.cljc
```diff
@@ -0,0 +1,554 @@
+(ns clara.tools.test-inspect
+  (:require
+    #?(:clj [clara.tools.testing-utils :refer [def-rules-test
```
mrrodriguez: Same comment here as above. I think using as many ns aliases and as few `:refer` as possible would help cut down on the noise and redundancy in the `ns` form.

sparkofreason:
I'm happy to use aliases instead of `:refer`, that's my personal preference anyway. I generally only use `:refer` when it's extremely/idiomatically clear, e.g. with core.async.

WilliamParker:
This looks good to me.  Note that it doesn't make public API changes beyond a simple extension of Clojure functionality to ClojureScript functionality.

@mrrodriguez : Do you have any further concerns here?  Thanks for the input on ClojureScript practices and the impact of more recent developments in that ecosystem (which I'm not as familiar with as you are).  Feel free to merge this if it looks good to you.

PR #364, opened 2017-12-02
ISSUE-316: Implement clear-ns-productions! to clear rule/query definitions which...

sparkofreason:
...have been removed, including defs under the old name of a renamed def.


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 This is a history of changes to clara-rules.
 
+### 0.17.0-SNAPSHOT
+* Add clear-ns-productions! functionality to support clean reloading of rule and query defintitions. See [issue 316](https://github.com/cerner/clara-rules/issues/316) for details.
```
mrrodriguez: Typo: 
"defintitions" > "definitions"

src/main/clojure/clara/macros.clj
```diff
@@ -17,6 +17,12 @@
             [clara.rules.schema :as schema]
             [clojure.set :as s]))
 
+;;; Clear productions stored in cljs.env/*compiler* for current namespace.
+;;; Only exists for it's side-effect, hence returns nil.
```
alex-dixon: it's -> its 😅

src/main/clojure/clara/rules.cljc
```diff
@@ -410,3 +410,21 @@ See the [query authoring documentation](http://www.clara-rules.org/docs/queries/
            (cond-> ~(dsl/parse-query* binding definition {} (meta &form))
              ~name (assoc :name ~(str (clojure.core/name (ns-name *ns*)) "/" (clojure.core/name name)))
              ~doc (assoc :doc ~doc)))))))
+
+#?(:clj
+   (defmacro clear-ns-productions!
+     "Ensures that any rule/query definitions which have been cached will be cleared from the associated namespace.
+      Rule and query definitions can be cached such if there definitions are not explicitly overwritten with the same
```
alex-dixon: there -> their

src/main/clojure/clara/rules.cljc
```diff
@@ -410,3 +410,21 @@ See the [query authoring documentation](http://www.clara-rules.org/docs/queries/
            (cond-> ~(dsl/parse-query* binding definition {} (meta &form))
              ~name (assoc :name ~(str (clojure.core/name (ns-name *ns*)) "/" (clojure.core/name name)))
              ~doc (assoc :doc ~doc)))))))
+
+#?(:clj
+   (defmacro clear-ns-productions!
+     "Ensures that any rule/query definitions which have been cached will be cleared from the associated namespace.
+      Rule and query definitions can be cached such if their definitions are not explicitly overwritten with the same
```
WilliamParker: nitpick - Did you mean "such that" here?

sparkofreason: Yes.

sparkofreason:
Will the proposed test work? The rules/sessions are defined at compile time, while `clear-ns-productions!` happens at run time, right? I think you have to force a recompile to actually test this, and I can't think of a way to do this in a unit test.

src/main/clojure/clara/rules.cljc
rbrush: Looks good to me overall.

WilliamParker:
@sparkofreason regarding

````
Will the proposed test work? The rules/sessions are defined at compile time, while clear-ns productions! 
happens at run time, right? I think you have to force a recompile to actually test this,
and I can't think of a way to do this in a unit test.
````

I'd expect the clearing logic to happen at compile time, specifically once the compiler reaches the clear-ns-productions! form.  Macros basically function as functions that take code and return code, but they can also perform arbitrary additional operations, with the caveat that in ClojureScript these operations will run in the environment of the Clojure compiler, rather than at initialization of the ClojureScript.  Some illustrative examples:

````
user=> (defmacro my-macro [] (println "Hello") `(println "World"))
#'user/my-macro
;; Compile a function that executes the code returned by my-macro.
user=> (defn run-macro-code [] (my-macro))
Hello
#'user/run-macro-code
;; Call that function.
user=> (run-macro-code)
World
nil
;; Macroexpand my-macro and immediately run the resulting code
user=> (my-macro)
Hello
World
nil
;; Directly call macroexpand on code calling my-macro without actually doing anything with the
;; returned code.
user=> (macroexpand `(my-macro))
Hello
(clojure.core/println "World")
````

Suppose we have logic like

````
(defrule rule1 ...)
(clear-ns-productions!)
(defrule rule2)
````

I'd expect the Clojure and ClojureScript compilers to create the rule1 definition, then purge it from the cache once they reach clear-ns-productions!.  After this, they would then create rule2 and add it to the cache.

I don't exclude the possibility that I'm missing something, particularly on the ClojureScript side, but my understanding is that the compiler simply takes the statements and evaluates them in order, so that the code in the clear-ns-productions! macro (as opposed to the code it returns) will be executed before the call to the defrule macro with the contents of rule2.

**Comment by sparkofreason** [comment]
Sorry, you're right, had my wires crossed and had gotten it into my head that clojure did a macro pass before compiling and executing. Which is sort of what happens in CLJS, I guess, but it doesn't matter here because we're doing everything with macros. I have a version of the tests which will be pushed momentarily.

File: src/test/common/clara/test_clear_ns_productions.cljc [hunk 5]
```diff
@@ -0,0 +1,75 @@
...
+(defsession cleared-session 'clara.test-clear-ns-productions :fact-type-fn identity)
+
+;;; Then tests validating what productions the respective sessions have.
+(deftest cleared?
+  (let [uncleared (-> uncleared-session (insert :a) (fire-rules))
+        cleared (-> cleared-session (insert :a) (fire-rules))]
+    (testing "cleared-session should not contain any productions before (clear-ns-productions!)"
+      (is (= :before-clearing @before-clearing))
+      (is (empty? (query cleared-session query-before-clearing)))
+      (is (empty? (query cleared-session query-after-clearing)))
```
WilliamParker: Clara sessions are immutable;

````
(-> cleared-session (insert :a) (fire-rules))
````
returns a new session (bound to cleared) with the fact :a inserted and the rules fired, but the original cleared-session still doesn't have any facts.  Query-after-clearing actually does return results on the "cleared" binding.  As written these assertions pass even if you remove the clear-ns-productions! call.

File: src/test/common/clara/test_clear_ns_productions.cljc [hunk 6]
```diff
@@ -0,0 +1,75 @@
...
+       :lhs '[{:type :b
+               :constraints []}]
+       :rhs '(clara.rules/insert! :after-clearing-seq)}]))
+
+(defsession cleared-session 'clara.test-clear-ns-productions :fact-type-fn identity)
+
+;;; Then tests validating what productions the respective sessions have.
+(deftest cleared?
+  (let [uncleared (-> uncleared-session (insert :a) (fire-rules))
+        cleared (-> cleared-session (insert :a) (fire-rules))]
```
WilliamParker: You have rules that look for :b facts but don't insert :b here, nor do I see it being inserted in a rule.  I think you need to insert :b as well and adjust your assertions accordingly?  Those assertions can be Clojure-only since ClojureScript doesn't support :production-seq.

src/test/common/clara/test_clear_ns_productions.cljc
```diff
@@ -0,0 +1,75 @@
...
+                      [cljs.test :refer [is deftest run-tests testing use-fixtures]])))
+
+(def before-clearing (atom nil))
+(defrule rule-to-be-cleared
+         [:a]
+         =>
+         (reset! before-clearing :before-clearing)
+         (insert! :before-clearing))
+
+(defquery query-to-be-cleared [] [?f <- :before-clearing])
```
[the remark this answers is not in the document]
sparkofreason: I should just delete it, because if you try to reference it after the call to `clear-ns-productions~`, you get a compiler error about an undefined symbol. I thought about adding a test to call `resolve` and validate that it doesn't exist anymore, which I'm happy to add if you think it's useful.

src/test/common/clara/test_clear_ns_productions.cljc
sparkofreason: Copy/paste error from an earlier version, that should have been `:a`, not `:b`.

src/test/common/clara/test_clear_ns_productions.cljc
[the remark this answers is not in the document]
WilliamParker: I forgot to mention - you can query on the name of the query as well, which won't throw an exception.   See [here in the compiler](https://github.com/cerner/clara-rules/blob/0.16.1/src/main/clojure/clara/rules/compiler.clj#L1662) and the [query implementation in the engine](https://github.com/cerner/clara-rules/blob/0.16.1/src/main/clojure/clara/rules/engine.cljc#L1960).

src/test/common/clara/test_clear_ns_productions.cljc
sparkofreason: Yikes, good catch. Will fix.

src/test/common/clara/test_clear_ns_productions.cljc
```diff
@@ -0,0 +1,75 @@
...
+       :rhs '(clara.rules/insert! :after-clearing-seq)}]))
+
+(defsession cleared-session 'clara.test-clear-ns-productions :fact-type-fn identity)
+
+;;; Then tests validating what productions the respective sessions have.
+(deftest cleared?
+  (let [uncleared (-> uncleared-session (insert :a) (fire-rules))
+        cleared (-> cleared-session (insert :a) (fire-rules))]
+    (testing "cleared-session should not contain any productions before (clear-ns-productions!)"
+      (is (= :before-clearing @before-clearing))
```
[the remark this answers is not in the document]
sparkofreason: Done

src/test/common/clara/test_clear_ns_productions.cljc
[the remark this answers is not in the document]
sparkofreason: RIght, cool. Done.

Issue #362, opened 2017-11-28
Refactor defrule and defquery for ClojureScript to better support customization


sparkofreason:
Currently, if I were to write a custom rule/query definition DSL in ClojureScript, the only choice would be to author macros that convert the new DSL into `defrule`'s or `defquery`'s in terms of the clara DSL. It would be easier to simply emit the data structures which are the result of the parsing and transformation performed by `defrule` and `defquery`. But both macros currently require a side-effect in CLJS to interact with the compiler, an implementation detail which could potentially change. Therefore it would be useful to refactor the macros such that they will accept the data structure, perform any side-effects, and emit the appropriate `def` statement.

Another scenario is when using the clara DSL, but there is a need to transform the resultant data structures. An example I encountered was using custom fact types, where the `defrecord` type was not immediately available to the clara compiler and thus default destructuring was not applied. Without explicit destructuring the CLJS compiler gives lots of warnings about undefined symbols, and there are cases where bugs occur when the local symbol collides with some other symbol in scope. This can be avoided with a custom macro that transforms the the rule/query definition, and again the easiest path is to transform the data-structure rather than emitting another version of `defrule` or `defquery`, which is enabled by making the rule-parsing logic available as a public function to be consumed by the custom macros, in addition to the change described in the first paragraph.

The proposed refactoring for `defrule` would look something like this:

```clojure
(defn build-rule
  [name & body]
  (let [doc (if (string? (first body)) (first body) nil)
        body (if doc (rest body) body)
        properties (if (map? (first body)) (first body) nil)
        definition (if properties (rest body) body)
        {:keys [lhs rhs]} (dsl/split-lhs-rhs definition)

        production (cond-> (dsl/parse-rule* lhs rhs properties {})
                           name (assoc :name (str (clojure.core/name (com/cljs-ns)) "/" (clojure.core/name name)))
                           doc (assoc :doc doc))]
    production))

(defn defrule!
  [name production]
  (add-production name production)
  `(def ~name
     ~production))

(defmacro defrule
  [name & body]
  (defrule! name (apply build-rule name body)))
```
Similar refactoring would apply to `defquery`.

WilliamParker:
[...]

PR #365, opened 2017-12-14
ISSUE-362: Refactor defrule and defquery for ClojureScript to better support...

sparkofreason:
...customization of rule/query DSL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ This is a history of changes to clara-rules.
 * Add clear-ns-productions! functionality to support clean reloading of rule and query definitions. See [issue 316](https://github.com/cerner/clara-rules/issues/316) for details.
 * Support session inspection and fact-graph in ClojureScript. See [issue 307](https://github.com/cerner/clara-rules/issues/307) for details.
 * Fix issue when compiling CLJS and session, rules, and fact-types are not in the same namespace or explicitly referred. See [issue 359](https://github.com/cerner/clara-rules/issues/359).
+* Refactor defrule and defquery for ClojureScript to better support customization. See [issue 362](https://github.com/cerner/clara-rules/issues/362) for details.
 
 ### 0.16.1
 * Fix deserialization failure when the rulebase contains a clojure.lang.PersistentList$EmptyList. See [issue 352](https://github.com/cerner/clara-rules/issues/352) for details.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@ This is a history of changes to clara-rules.
 * Fix issue with incorrect namespace qualification of rule and query code in ClojureScript. See [issue 359](https://github.com/cerner/clara-rules/issues/359).
 * Add clear-ns-productions! functionality to support clean reloading of rule and query definitions. See [issue 316](https://github.com/cerner/clara-rules/issues/316) for details.
 * Support session inspection and fact-graph in ClojureScript. See [issue 307](https://github.com/cerner/clara-rules/issues/307) for details.
-* Fix issue when compiling CLJS and session, rules, and fact-types are not in the same namespace or explicitly referred. See [issue 359](https://github.com/cerner/clara-rules/issues/359).
 * Refactor defrule and defquery to better support customization. See [issue 362](https://github.com/cerner/clara-rules/issues/362) for details.
 
 ### 0.16.1
--- a/src/main/clojure/clara/macros.clj
+++ b/src/main/clojure/clara/macros.clj
@@ -43,19 +43,6 @@
    (coll? source) (seq source)
    :else (throw (IllegalArgumentException. "Unknown source value type passed to defsession"))))
 
-(defn build-rule
-  [name & body]
-  (let [doc (if (string? (first body)) (first body) nil)
-        body (if doc (rest body) body)
-        properties (if (map? (first body)) (first body) nil)
-        definition (if properties (rest body) body)
-        {:keys [lhs rhs]} (dsl/split-lhs-rhs definition)
-
-        production (cond-> (dsl/parse-rule* lhs rhs properties {})
-                           name (assoc :name (str (clojure.core/name (com/cljs-ns)) "/" (clojure.core/name name)))
-                           doc (assoc :doc doc))]
-    production))
-
 (defn defrule!
   [name production]
   (add-production name production)
@@ -64,18 +51,7 @@
 
 (defmacro defrule
   [name & body]
-  (defrule! name (apply build-rule name body)))
-
-(defn build-query
-  [name & body]
-  (let [doc (if (string? (first body)) (first body) nil)
-        binding (if doc (second body) (first body))
-        definition (if doc (drop 2 body) (rest body))
-
-        query (cond-> (dsl/parse-query* binding definition {})
-                      name (assoc :name (str (clojure.core/name (com/cljs-ns)) "/" (clojure.core/name name)))
-                      doc (assoc :doc doc))]
-    query))
+  (defrule! name (dsl/build-rule name body)))
 
 (defn defquery!
   [name query]
@@ -85,7 +61,7 @@
 
 (defmacro defquery
   [name & body]
-  (defquery! name (apply build-query name body)))
+  (defquery! name (dsl/build-query name body)))
 
 (sc/defn gen-beta-network :- [sc/Any] ; Returns a sequence of compiled nodes.
   "Generates the beta network from the beta tree. "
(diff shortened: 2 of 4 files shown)


sparkofreason:
@WilliamParker thanks, no problem. One thing worth noting is that the current solution only applies to CLJS, and I think parity in CLJ should probably also be a goal.

WilliamParker:
The idea here looks sound to me @sparkofreason .  As you suggest though it would probably be better to have parity between Clojure and ClojureScript here.  My suggestion would be to add the build-rule and build-query functions to the dsl namespace, and then call them in both clara.rules.macros (for ClojureScript) and clara.rules (for Clojure).  The signatures would need to change slightly to account for the (meta &form) argument that are being passed in to parse-rule* and parse-query* in Clojure.  If ClojureScript has the same available it could do likewise; the usage appears to be in giving more informative exceptions from [throw-dsl-ex](https://github.com/cerner/clara-rules/blob/0.16.1/src/main/clojure/clara/rules/dsl.clj#L29).  

Fundamentally, the build-rule and build-query functions seem like DSL logic to me, and putting them there would allow sharing between clara.rules, clara.rules.macros, and other users in both Clojure and ClojureScript macros.  It would also reduce our codebase's size a bit and enforce more similarity between Clojure and ClojureScript.

Thoughts/objections/alternative suggestions?

sparkofreason:
I'll dig into this tomorrow, just got back from holiday travels. That all sounds fine, just note that once we move those functions to a different namespace we won't be able to use the CLJS `env/*compiler*` any more for storing productions during CLJS compilation. I tested on my local copy with an atom defined in the DSL namespace, and everything seems to work fine. I can't think of any reason why `env/*compiler*` is required for this case. It's just an atom, with the added feature that the CLJS compiler resets it for each namespace being compiled, but that isn't being used here AFAICT.

sparkofreason:
Also, I think I mentioned this in the discussion of #362: as part of improving support for 3rd-party DSLs, I'd like to support keyword-named rules and queries. Let me know if that's okay to include here, or if I should make a separate issue/PR.

WilliamParker:
@sparkofreason I can't speak for anyone else, but I was thinking of having build-rule be in the dsl namespace, not the defrule! function.  I think two different implementations of the latter in the macros and rule namespaces for clj and cljs respectively seems reasonable.  Even if they were the same function you'd end up with two different implementations controlled by a switch between clj and cljs, just in the same fn body.  I don't think build-rule has the same concerns though unless I'm missing something?

Regarding keyword names, so you're proposing something like 

````
(defrule! :your-name-here rule-body-here)
````

That seems like something we could discuss, but I'd suggest logging another issue for it so that it can be discussed in a more visible way than deep in the comments of another issue.

**Comment by sparkofreason** [comment]
Re: `defrule!` and `build-rule` namespaces, seems reasonable.

For keyword support, I was thinking of just adding the support internally so other DSLs could use it, rather than modifying the default clara DSL. The former is a pretty minor change change, just checking if the name is a keyword and transforming appropriately (see https://github.com/Provisdom/clara-rules/commit/6dd773b43b4ef14ee4fa2d774f0235d33e171410). The latter seems like a bigger deal with no obvious benefit, and seems like it breaks the var/namespace abstraction in the current DSL.

**Comment by sparkofreason** [comment]
See #371.

File: src/main/clojure/clara/rules/dsl.clj [hunk 1]
```diff
@@ -294,7 +294,35 @@
...
+  "Function used to parse and build a query using the DSL syntax."
+  ([name body] (build-query name body {}))
+  ([name body form-meta]
+   (let [doc (if (string? (first body)) (first body) nil)
+         binding (if doc (second body) (first body))
+         definition (if doc (drop 2 body) (rest body))]
+     (cond-> (parse-query* binding definition {} form-meta)
+             name (assoc :name (if (com/compiling-cljs?)
+                                 (str (clojure.core/name (com/cljs-ns)) "/" (clojure.core/name name))
+                                 (str (clojure.core/name (ns-name *ns*)) "/" (clojure.core/name name))))
```
WilliamParker: Minor: it might be a little easier to read if we reduced the duplication in forming the name by factoring out the ns construction logic into a let binding and sharing the rest.  Something along the lines of:

````
name (assoc :name (let [query-ns (clojure.core/name (if (com/compiling-cljs?) (com/cljs-ns) (ns-name *ns*)))]
                               (str query-ns "/" (clojure.core/name name)))
````

File: src/main/clojure/clara/rules/dsl.clj [hunk 2]
```diff
@@ -294,7 +294,35 @@
...
+  "Function used to parse and build a rule using the DSL syntax."
+  ([name body] (build-rule name body {}))
+  ([name body form-meta]
+   (let [doc (if (string? (first body)) (first body) nil)
+         body (if doc (rest body) body)
+         properties (if (map? (first body)) (first body) nil)
+         definition (if properties (rest body) body)
+         {:keys [lhs rhs]} (split-lhs-rhs definition)]
+     (cond-> (parse-rule* lhs rhs properties {} form-meta)
+             name (assoc :name (if (com/compiling-cljs?)
```
WilliamParker: Same as the comment on building the name in build-query

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@ This is a history of changes to clara-rules.
 * Fix issue with incorrect namespace qualification of rule and query code in ClojureScript. See [issue 359](https://github.com/cerner/clara-rules/issues/359).
 * Add clear-ns-productions! functionality to support clean reloading of rule and query definitions. See [issue 316](https://github.com/cerner/clara-rules/issues/316) for details.
 * Support session inspection and fact-graph in ClojureScript. See [issue 307](https://github.com/cerner/clara-rules/issues/307) for details.
+* Fix issue when compiling CLJS and session, rules, and fact-types are not in the same namespace or explicitly referred. See [issue 359](https://github.com/cerner/clara-rules/issues/359).
```
WilliamParker: Issue 359 is in the changelog already on line 5.  I believe I changed the changelog message when I did the release on that particular entry; basically I just wanted to specifically indicate what the issue was rather than just indicating that an issue existed.  If you think it could be more clear I'm open to making changes, we'll want to have a single bullet point in the changelog though.

src/main/clojure/clara/macros.clj
```diff
@@ -43,33 +43,25 @@
    (coll? source) (seq source)
    :else (throw (IllegalArgumentException. "Unknown source value type passed to defsession"))))
```
WilliamParker: Looking back at this I'm not sure why we don't add a docstring to the var def the same way as in Clojure.  That's the status quo though, we don't need to make changes in this PR.

CHANGELOG.md
sparkofreason: I just messed up the merge from upstream. Removing.

src/main/clojure/clara/rules/dsl.clj
sparkofreason: Done.

alex-dixon:
LGTM.

Maybe we should add an issue to document how it can/should be used?

Didn't notice this before but looks like `parse-rule*` depends on the binding of *ns*. `build-rule` does also (because it calls `parse-rule*`) but also here: https://github.com/Provisdom/clara-rules/blob/d1c903dde46b9b3f364af4c66ca2116936d2b3c9/src/main/clojure/clara/rules/dsl.clj#L307 

I don't know any ramifications to any of that...maybe we could consider mentioning it in the docstring like we have for `parse-rule*`? 

Should there be an optional `ns` arg that defaults to `*ns*`? Not strictly necessary in CLJ since you can always call with `(binding [*ns* my-ns] ...)`, but in CLJS doesn't seem like we give the user the option over what namespace the rule is defined in.

sparkofreason:
I'm working on a project (https://github.com/Provisdom/maali) which uses a custom DSL, and would be happy to contribute some documentation  for DSL-builders based on that.

WilliamParker:
@alex-dixon actually I think that the ns can be controlled in ClojureScript; see [this code qualifying symbols](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/macros.clj#L190).  Basically it is just another dynamic variable to bind, namely \*cljs-ns\*.  I'd be open to discussing allowing the user to pass in an explicit argument, but IMO we can do that in another issue, noting that we haven't released these changes yet even if merged.  I do think we'd want to have the ns of the rule name and the logical namespace of the rule (i.e. the ns where the forms are evaluated) to be the same if the name corresponds to a var in a namespace.  Technically this doesn't need to be the case, but it seems confusing if it isn't.  This could related to https://github.com/cerner/clara-rules/issues/371 as well since that would change the expectations around production names.

@alex-dixon @sparkofreason Adding some documentation for this sounds like a fine idea. :)  Perhaps an additional section under "Advanced Topics" at clara-rules.org?  The source for the site is at https://github.com/cerner/clara-site in case you didn't know.  Feel free to suggest improvements, submit PRs, etc. to that project as well, for this or other topics.

I'm going to merge this since it has been in its current state for a while without any objections to doing so.  The further issues on this seem like things we can iterate on after a merge.

Issue #377, opened 2018-01-22
Enhance performance while adding conjunctions in beta graph construction


EthanEChristian:
Profiling locally indicates that there is a bit of a hotspot in beta graph construction, specifically [add-conjunctions](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/compiler.clj#L1052).
There are probably other hot spots but this one stood out as a potentially easy one to solve, this issue lies with the `sorted-set` and the `first` call on the for loop seen [here](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/compiler.clj#L1081-L1098).

The sorted nature of the set of forward edges must be maintained to provide consistent session creation, however we could potentially save some time if we had a different approach.

EthanEChristian:
I have a working branch, and i plan on posting a Pull request soon

WilliamParker:
I've merged the PR.  Enhancements to minimize calls to try-eval can be further discussed on https://github.com/cerner/clara-rules/issues/381.  Closing the issue, thanks @EthanEChristian

PR #378, opened 2018-01-22
ISSUE-377: Enhance performance while adding conjunctions in beta grap…

EthanEChristian:
…h construction

#377

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1077,28 +1077,31 @@
                            result-binding (conj result-binding)
                            fact-binding (conj fact-binding))
 
-            ;; Find children that all parent nodes have.
-            common-children-ids (into
-                                 ;; Since we require that id-to-condition-nodes have an equal value to "node" under the ID
-                                 ;; for this to be used, the order of this set probably doesn't matter in most cases.  However, in any possible edge
-                                 ;; cases where there are equal nodes under different IDs, sorting the set will add determinism.
-                                 ;; Having different nodes under the same
-                                 ;; ID would be a bug, but having an equivalent node under multiple IDs wouldn't necessarily be one.
-                                 (sorted-set)
-                                 (apply s/union (-> (:forward-edges beta-graph)
-                                                    (select-keys parent-ids)
-                                                    (vals))))
-
+            ; Find children that all parent nodes have.
+            forward-edges (apply s/union (-> (:forward-edges beta-graph)
+                                                (select-keys parent-ids)
+                                                (vals)))
 
             id-to-condition-nodes (:id-to-condition-node beta-graph)
 
-            ;; Find an id for an equivalent node.
-            existing-id (first (for [child-id common-children-ids
-                                     :when (= node (get id-to-condition-nodes child-id))]
-                                 child-id))
+            ;; Since we require that id-to-condition-nodes have an equal value to "node" under the ID
+            ;; for this to be used, the order of this set probably doesn't matter in most cases.  However, in any possible edge
+            ;; cases where there are equal nodes under different IDs, sorting the set will add determinism.
+            ;; Having different nodes under the same
+            ;; ID would be a bug, but having an equivalent node under multiple IDs wouldn't necessarily be one.
+            update-node->id (fn [m id]
+                              (let [node (get id-to-condition-nodes id)
+                                    prev (get m node)]
+                                (if prev
+                                  (assoc! m node (min prev id))
+                                  (assoc! m node id))))
 
-            ;; Use the existing id or create a new one.
-            node-id (or existing-id
+            node->id (reduce update-node->id
+                             (transient {})
+                             forward-edges)
+
+            ; Use the existing id or create a new one.
+            node-id (or (get node->id node)
                         (create-id-fn))
 
             graph-with-node (add-node beta-graph parent-ids node-id node)]
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1080,20 +1080,30 @@
             ;; Find children that all parent nodes have.
             forward-edges (if (= 1 (count parent-ids))
                             ;; If there is only one parent then there is no need to reconstruct
-                            ;; the set of forward edges
+                            ;; the set of forward edges. This is an intentional performance optimization for large
+                            ;; beta graphs that have many nodes branching from a common node, typically the root node.
                             (-> (:forward-edges beta-graph)
                                 (get (first parent-ids)))
                             (->> (select-keys parent-ids (:forward-edges beta-graph))
                                  vals
-                                 (into [] (comp cat (distinct)))))
+                                 ;; Order doesn't matter here as we will effectively sort it using update-node->id later,
+                                 ;; thus adding determinism.
+                                 (into #{} cat)))
 
             id-to-condition-nodes (:id-to-condition-node beta-graph)
 
             ;; Since we require that id-to-condition-nodes have an equal value to "node" under the ID
-            ;; for this to be used, the order of this set probably doesn't matter in most cases.  However, in any possible edge
-            ;; cases where there are equal nodes under different IDs, sorting the set will add determinism.
-            ;; Having different nodes under the same
-            ;; ID would be a bug, but having an equivalent node under multiple IDs wouldn't necessarily be one.
+            ;; for this to be used. In any possible edge cases where there are equal nodes under different IDs,
+            ;; maintaining the lowest node id will add determinism.
+            ;; Having different nodes under the same ID would be a bug,
+            ;; but having an equivalent node under multiple IDs wouldn't necessarily be one.
+            ;;
+            ;; Using maps(nodes) as keys acts as a performance optimization here, we are relying on the fact that maps cache
+            ;; their hash codes. This saves us time in the worst case scenarios when there are large amounts of forward edges
+            ;; that will be compared repeatedly. For example, when adding new children to the root node we must first compare
+            ;; all of the old children for a prior match, for each additional child added we incur the assoc but the hash code
+            ;; has already been cached by prior iterations of add-conjunctions. In these cases we have seen that the hashing
+            ;; will out perform equivalence checks.
             update-node->id (fn [m id]
                               (let [node (get id-to-condition-nodes id)
                                     prev (get m node)]
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1077,10 +1077,15 @@
                            result-binding (conj result-binding)
                            fact-binding (conj fact-binding))
 
-            ; Find children that all parent nodes have.
-            forward-edges (apply s/union (-> (:forward-edges beta-graph)
-                                                (select-keys parent-ids)
-                                                (vals)))
+            ;; Find children that all parent nodes have.
+            forward-edges (if (= 1 (count parent-ids))
+                            ;; If there is only one parent then there is no need to reconstruct
+                            ;; the set of forward edges
+                            (-> (:forward-edges beta-graph)
+                                (get (first parent-ids)))
+                            (->> (select-keys parent-ids (:forward-edges beta-graph))
+                                 vals
+                                 (into [] (comp cat (distinct)))))
 
             id-to-condition-nodes (:id-to-condition-node beta-graph)
 
@@ -1096,11 +1101,12 @@
                                   (assoc! m node (min prev id))
                                   (assoc! m node id))))
 
-            node->id (reduce update-node->id
-                             (transient {})
-                             forward-edges)
+            node->id (-> (reduce update-node->id
+                                 (transient {})
+                                 forward-edges)
+                         persistent!)
 
-            ; Use the existing id or create a new one.
+            ;; Use the existing id or create a new one.
             node-id (or (get node->id node)
                         (create-id-fn))
 


EthanEChristian:
Generating some rules to judge the performance:
```
(ns clara.test-gen-rules
  (:require [clojure.test :refer :all]
            [clara.rules.compiler :as com]
            [clojure.template :as temp]
            [clara.rules :as r]))

(defrecord CommonRecord [id])

(defmacro gen-rules
  [num-rules]
  (let [gen-facts (mapv (comp keyword str) (range num-rules))
        names (mapv (comp gensym #(str "rule-" %)) (range num-rules))
        body (interleave gen-facts names)]
    `(temp/do-template
       [~'fact
        ~'name]
       (r/defrule ~'name
                  [~'?a ~'<- ~'CommonRecord]
                  [~'?b ~'<- ~'fact (= (:id ~'?a) ~'this)]
                  ~'=>
                  nil)
       ~@body)))

(gen-rules 2000)
```

and 

```
(ns clara.test-performance
  (:require [clojure.test :refer :all]
            [clara.rules :as r]
            [clojure.string :as cs]
            [clara.test-gen-rules]))

(let [iterations 15
      sum (reduce
            (fn [prev iter]
              (let [_ (println "Iteration: " iter)
                    time-str (with-out-str (time (r/mk-session 'clara.test-gen-rules :cache false)))]
                (-> time-str
                    (cs/replace #"\"Elapsed time: " "")
                    (cs/replace #" msecs\"" "")
                    Double/valueOf
                    (+ prev))))
            0
            (range iterations))]
  (println (str "Average: " (/ sum iterations) " ms")))
```

EthanEChristian:
```
BEFORE:
Iteration:  0
Iteration:  1
Iteration:  2
Iteration:  3
Iteration:  4
Iteration:  5
Iteration:  6
Iteration:  7
Iteration:  8
Iteration:  9
Iteration:  10
Iteration:  11
Iteration:  12
Iteration:  13
Iteration:  14
Average: 27518.9137552 ms
```


```
AFTER:
Iteration:  0
Iteration:  1
Iteration:  2
Iteration:  3
Iteration:  4
Iteration:  5
Iteration:  6
Iteration:  7
Iteration:  8
Iteration:  9
Iteration:  10
Iteration:  11
Iteration:  12
Iteration:  13
Iteration:  14
Average: 24443.8885162 ms
```

EthanEChristian:
While 3 secs isn't a lot, this only 2000 productions.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1077,28 +1077,31 @@
...
-                                 ;; ID would be a bug, but having an equivalent node under multiple IDs wouldn't necessarily be one.
-                                 (sorted-set)
-                                 (apply s/union (-> (:forward-edges beta-graph)
-                                                    (select-keys parent-ids)
-                                                    (vals))))
-
+            ; Find children that all parent nodes have.
+            forward-edges (apply s/union (-> (:forward-edges beta-graph)
+                                                (select-keys parent-ids)
+                                                (vals)))
```
mrrodriguez: If I'm reading this logic correctly, isn't an alternative implementation something like:
```clj

node-id (if (seq forward-edges)
          (apply min forward-edges)
          (create-id-fn))
```

Something like that seems to have less building involved than the `sorted-set` or this new `transient {}` impl. I may be missing something about what else is happening here though.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1077,28 +1077,31 @@
...
+            ;; Having different nodes under the same
+            ;; ID would be a bug, but having an equivalent node under multiple IDs wouldn't necessarily be one.
+            update-node->id (fn [m id]
+                              (let [node (get id-to-condition-nodes id)
+                                    prev (get m node)]
+                                (if prev
+                                  (assoc! m node (min prev id))
+                                  (assoc! m node id))))
+
+            node->id (reduce update-node->id
```
mrrodriguez: The map returned here is a transient. Although, `get` does work on that, I think it is something that'd need to be handled with care. I think this is only a 1-time use for now though, so probably ok. I think the cost of going to `persistent!` is fast, but I haven't done a comparison.

Also, as I mentioned above, I'm questioning the need to build this intermediate construct.

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: only if the "forward edge"'s node is equivalent to that of the node that was constructed

mrrodriguez: Good point. So revising my question, what about:
```clj

node-id (or (some->> forward-edges
                     seq
                     (filter #(= node (get id-to-condition-nodes %)))
                     (apply min))
            (create-id-fn))
```

I'm basically just trying to see if there isn't some alternative that doesn't require such heavy construction of intermediate structures. I guess the `sorted-set` was somewhat expensive. I wonder why? 
Also, `clojure.core/min` is pretty slow I believe. And having to use `apply` since it is variatic is even worse. Could write a faster one (that takes a collection) for this usage and it may make something like my example here even more performant to compare with.

EthanEChristian: From VM snapshots, sorted-set was expensive due to `cons` on the underlying data struct.

The reason i avoided something like the `filter`, was because the `equiv` gets hot, hotter than a simple hash while using `assoc!`.

EthanEChristian:
We have examples, with ~40k productions and the savings is ~44 secs

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: I see, so trying to offload `equiv` by using more hashing. Interesting.
I guess the transient map thing seems reasonable. They do build pretty fast compared to other persistent colls.

The only remaining question I'd have then, is why not just use a sorted collection via `sort` instead of a `sorted-set` to solve the problem.

It might be a bit better to build the `forward-edges` set inline during the `reduce` via `transduce`. I think you'd still want to use `distinct` though, so you don't get to get fully away from building a set.
```clj
node->id (transduce (comp cat (distinct))
                    (completing update-node->id)
                    (-> (:forward-edges beta-graph)
                        (select-keys parent-ids)
                        vals))
```
Maybe not a big deal though if you don't see any reason to question the perf of the `apply s/union` happening here. If `sorted-set` was expensive though, this must be a fairly hot spot. I'm not sure `apply s/union` would be that cheap depending on how many collections were involved typically.

WilliamParker: As I understand this if we set aside the effort spent on constructing the sorted-set, we have the following sources of algorithmic complexity in a big-O sense:


Ethan's proposal:
````
update-node->id (fn [m id]
                  (let [node (get id-to-condition-nodes id)
                        prev (get m node)]
                    (if prev
                      (assoc! m node (min prev id))
                      (assoc! m node id))))

node->id (reduce update-node->id
                 (transient {})
                 forward-edges)
````
For each forward edge, we'll have a get on an integer key, a get on a node key, a map update and potentially a min/max comparison.  The map gets will be O(n) in the number of elements in forward-edges.

Current version:
````
;; Find an id for an equivalent node.
existing-id (first (for [child-id forward-edges
                         :when (= node (get id-to-condition-nodes child-id))]
                     child-id))
````
Note that I replaced "common-children-ids" with "forward-edges" here for clarity since they have the same information in both versions, albeit with different names.  We basically have O(n) equality comparisons and gets on integer keys in the number of elements in forward-edges.

Between these, it seems like equality could be faster due to potentially quick returns of false and the avoidance of calculating hashcodes on larger objects like nodes.. but I could also mistaken.  The real problem here looks like it is probably the sorted-set, and VisualVM snapshots show lots of time spent in clojure.lang.PersistentTreeSet.cons.  Sorted sets (and sorted maps) don't support transients per [CLJ-955](https://dev.clojure.org/jira/browse/CLJ-995), and I verified this in a REPL.

````
my-ns.core> (transient (sorted-set))
ClassCastException clojure.lang.PersistentTreeSet cannot be cast to clojure.lang.IEditableCollection  clojure.core/transient (core.clj:3233)
````

As a result Clojure will have to resort to building the sorted-set one element at a time in "into"; see [here](https://github.com/clojure/clojure/blob/clojure-1.9.0/src/clj/clojure/core.clj#L6821).  This will be slow:

````
(into (sorted-set) stuff-here)
````

Was something like the following considered?

````
(sort
  (apply s/union (-> (:forward-edges beta-graph)
                     (select-keys parent-ids)
                     (vals))))
````

The reason I wonder if this might perform better is that the [sort](https://github.com/clojure/clojure/blob/clojure-1.9.0/src/clj/clojure/core.clj#L3107) implementation is ultimately implemented by a sort of a mutable Java array, which is probably pretty performant, even if it isn't of linear complexity.

Regarding the time saved, my understanding is that the actual case motivating this is significantly larger than the example given in the issue; please correct me if I'm mistaken @EthanEChristian.

File: src/main/clojure/clara/rules/compiler.clj (hunk 2 above)
Comment (WilliamParker) [reply src/main/clojure/clara/rules/compiler.clj]: The [doPersistent](https://github.com/clojure/clojure/blob/clojure-1.9.0/src/jvm/clojure/lang/PersistentHashMap.java#L350) call appears to just build a new persistent map that shares the underlying structure; see the [constructor](https://github.com/clojure/clojure/blob/clojure-1.9.0/src/jvm/clojure/lang/PersistentHashMap.java#L103).  Unless I'm misunderstanding something here the cost of persistent! seems trivial so I'd be inclined to use it if we keep this construct.

File: src/main/clojure/clara/rules/compiler.clj (hunk 1 above)
Comment (mrrodriguez) [reply src/main/clojure/clara/rules/compiler.clj]: @WilliamParker 
I also considered `sort`. Probably without `s/union`. I don't think `apply s/union` is very fast at all. In particular if it goes into the > 2 arity cases (which I think happens here fairly easily).

Something like
```clj
common-children-ids (->> (select-keys (:forward-edges beta-graph))
                         vals
                         (into [] (comp cat (distinct)))
                         sort)
```

Also,
> Between these, it seems like equality could be faster due to potentially quick returns of false and the avoidance of calculating hashcodes on larger objects like nodes..

Equality may not be quick to return false. Perhaps nodes are always similar in the first keys looked at. However, hash also has its own penalty, as you've mentioned. The hash code may be cached though and perhaps the same nodes are repeatedly hashed. Then again, if they are records, they don't cache hash codes at the top-level at least, until clj v1.9. Even if the top-level is cached though, larger internal structures may be (if they are clj persistent structures for example).

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: Yeah `persistent!` has in docs that it is constant time. So it definitely can't be rebuilding a full collection over again. I just wasn't positive what other overhead there was involved. Given the links you posted, looks like very little overhead.

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: Correct, the issue i was seeing has at max ~10k forward edges, however there are multiple calls where the forward edges approach 10k.

These are pretty big sessions, 40k productions.

@mrrodriguez,
Testing the `transduce` suggested seems to be worse performing, do to what seems to be a performance enhancement in the `clojure.set` library when [unioning]( https://github.com/clojure/clojure/blob/master/src/clj/clojure/set.clj#L13-L31) several sets.

mrrodriguez:
> We have examples, with ~40k productions and the savings is ~44 secs.

That is a sizeable difference. 44 seconds is already a long time to be spending building the network, yikes! I hope after the 44 sec reduction there were many seconds left... :P

EthanEChristian:
@mrrodriguez 
Sadly no, the next hotspots seem to be in the `eval`ing of the nodes.

Its down to ~10 mins... not the best but ~44 secs is better than nothing

mrrodriguez:
@EthanEChristian 

> Sadly no, the next hotspots seem to be in the evaling of the nodes.

Sounds like there is a lot of potential improvements to be had in the compiler then. I've been unhappy with the amount of `eval` before too. We just need to get a bit more clever there I think.

There is some relationship of this topic to Drools. In Drools it has an "interpreted" mode for most/all conditions at first and only JITs them to a specialized class when they are used more than some threshold times (like 10). However, that only works for built-in operations that the interpreter is aware of. You implement ad hoc stuff via function calls. 

Clara doesn't specialize on anything or have an interpreter to do something like that. The network could potential delay `eval` until the expression is actually needed. However, if they all end up being needing it, I don't think it'd do much more than just give a bunch of slow "warm-up" time.

Perhaps there could be more sharing of common s-expr done though to avoid the need for so much `eval`. The tricky thing to look out for there is how to compare expressions as `=`. The thing that stands out to me there is that metadata can be important. However, I think there are libs out there that have done "equal including metadata" things before (`potemkin` I believe does in some cases).

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: @EthanEChristian I saw that optimization and did not believe it could actually be very fast. Looks pretty slow. It just picks the biggest set to put in front of the merge. However, the steps to do that seem costly. Also, I didn't think our set sizes should be drastically different in this union op here. However, I don't know that. I'll take your perf measurements word for it. Thanks for trying it out though.

EthanEChristian: @mrrodriguez @WilliamParker , digging deeper into the example i have, it would seem the time savings is not due to the logic implemented in `clojure.set`. 

The actual reason its faster is because the nodes that are being added when the number of forward forward edges is the highest, are children of the root node. Therefore the sets that are being unioned, is in actuality a single set and then goes down the single arg path of union, basically identity.

EthanEChristian: I think that i might change to the idea that @mrrodriguez  purposed above, but special case the scenario where there is only one parent so that we are not rebuilding the set of forward-edges. 

I will try and post an update shortly.

EthanEChristian: switching to use:
```
            forward-edges (if (= 1 (count parent-ids))
                            ;; If there is only one parent then there is no need to reconstruct
                            ;; the set of forward edges
                            (-> (:forward-edges beta-graph)
                                (get (first parent-ids)))
                            (->> (select-keys parent-ids (:forward-edges beta-graph))
                                 vals
                                 (into [] (comp cat (distinct)))))

            id-to-condition-nodes (:id-to-condition-node beta-graph)

            existing-id (some #(when (= node (get id-to-condition-nodes %))
                                 %)
                              (sort forward-edges))
```

Though this is still slightly slower than what i have currently, ~3 secs in the add-conjunctions function.

EthanEChristian:
@mrrodriguez @WilliamParker 

I have modified the PR to include some of the set creation pattern that you both mentioned above, while avoiding recreating it when its unnecessary. This saved ~2 secs over what it was before, but i still retained the hashmap implementation that i had before as the equiv checks still seem more costly than the hashcodes.

I also addressed the other comments made, around persistent! and the semicolon on the comment.

WilliamParker:
This looks good to me @EthanEChristian although I'd like to take another final look before merging, particularly since this is a high-risk part of the codebase.  If he has time perhaps @rbrush could take a glance as well since he did a lot of this node sharing functionality originally IIRC.

IMO we should also look at having a source-controlled set of benchmarks to which the ones you created can be added, even if we don't add a nice user interface to run them immediately although we should do that too.  At one point the [clara-benchmark](https://github.com/rbrush/clara-benchmark) project was created to compare Drools and Clara, but that has some drawbacks:

- The benchmarks are mostly very simple cases.
- The complication of comparing Drools and Clara isn't useful anymore IMO since whether Clara could be adopted for production use is a ship that has sailed at this point (Cerner has built major systems on it for example), and even if in a particular case Clara performs more poorly than Drools the key is that Clara not regress relative to its previous versions and preferably improve.

rbrush:
+1. Changes look good to me.

As for the eval performance questions, I agree with @mrrodriguez that there are optimizations to be made there as well. Since large chunks of Clojure code can be compiled much faster than that, I wonder if we're running into some base overhead for each eval call that could be eliminated by building a common, large data structure of functions to compile --  and eval-ing it all at once. (Although this might best be a topic for another issue.)

mrrodriguez:
@rbrush 

> Since large chunks of Clojure code can be compiled much faster than that, I wonder if we're running into some base overhead for each eval call that could be eliminated by building a common, large data structure of functions to compile -- and eval-ing it all at once. 

This is really similar to what I was recently thinking about the subject.

I've logged #381 for this and included these thoughts and a few more.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1078,27 +1078,36 @@
...
 
-            ;; Find an id for an equivalent node.
-            existing-id (first (for [child-id common-children-ids
-                                     :when (= node (get id-to-condition-nodes child-id))]
-                                 child-id))
+            ;; Since we require that id-to-condition-nodes have an equal value to "node" under the ID
+            ;; for this to be used, the order of this set probably doesn't matter in most cases.  However, in any possible edge
+            ;; cases where there are equal nodes under different IDs, sorting the set will add determinism.
+            ;; Having different nodes under the same
+            ;; ID would be a bug, but having an equivalent node under multiple IDs wouldn't necessarily be one.
```
WilliamParker: This comment is out of date now since this isn't a set after these changes - could you remove this and add appropriate comments explaining that we're maintaining deterministic ordering (along with how and why) along with an explanation of the optimizations?

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1078,27 +1078,36 @@
...
+            update-node->id (fn [m id]
+                              (let [node (get id-to-condition-nodes id)
```
WilliamParker: Since the node here is part of the id-to-condition-nodes, part of the beta-graph that is built up iteratively, I believe this will the same object if used in multiple iterations of this.  As a result the hashcode will be cached, minimizing the cost.  In the [current version on the other hand](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/compiler.clj#L1072) the node that we [do an equality check on later](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/compiler.clj#L1097) is a new object, which limits opportunities to reduce the work.  I suggest mentioning this facilitating of hashcode caching in a comment.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1078,27 +1078,36 @@
...
-                                                    (vals))))
-
+            forward-edges (if (= 1 (count parent-ids))
+                            ;; If there is only one parent then there is no need to reconstruct
+                            ;; the set of forward edges
+                            (-> (:forward-edges beta-graph)
+                                (get (first parent-ids)))
+                            (->> (select-keys parent-ids (:forward-edges beta-graph))
+                                 vals
+                                 (into [] (comp cat (distinct)))))
```
WilliamParker: Let's discuss why this doesn't have to be sorted in comment(s) - namely, that update-node->id gets the minimum id for a node, which is the only information from the current sorting that we actually need.

WilliamParker: We could probably improve on Clojure'd distinct transducer here for the case of integers, since it just [uses a persistent set](https://github.com/clojure/clojure/blob/clojure-1.9.0/src/clj/clojure/core.clj#L4976), and we could use either a transient set or a logical set backed by an array with a length set to prevent overflow.  IMO that would be premature evidence until/unless we have evidence that this is a perf problem.

Overall I expect removing the sorted-set here to be a significant benefit.  From profiling it seems it performs really poorly, which is not entirely surprising since its implementation seems to build iteratively rather than using some kind of transient strategy, presorting when provided all elements at once, etc.

WilliamParker:
@EthanEChristian basically it looks good, I think we just need to update some code comments that are now incorrect with these changes and probably add some others.  I added comments explaining what I'd like to see clarified.  After we make appropriate changes I'll plan to merge this unless anyone objects.  

I'd still like to add a benchmarks collections into source control in Clara in the near future, but since we do have the benchmarks we'd add there for this I don't think we need to immediately block this on that.

EthanEChristian:
@WilliamParker,

I have made some changes to the comments that you recommended, it think that they are a little more explicit on the intent of the changes made.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1078,27 +1078,46 @@
...
+                            ;; If there is only one parent then there is no need to reconstruct
+                            ;; the set of forward edges. This is an intentional performance optimization for large
+                            ;; beta graphs that have many nodes branching from a common node, typically the root node.
+                            (-> (:forward-edges beta-graph)
+                                (get (first parent-ids)))
+                            (->> (select-keys parent-ids (:forward-edges beta-graph))
+                                 vals
+                                 ;; Order doesn't matter here as we will effectively sort it using update-node->id later,
+                                 ;; thus adding determinism.
+                                 (into #{} cat)))
```
WilliamParker: This is probably marginally faster than your previous version:

````
(into [] (comp cat (distinct)))
````
since that one conj'ed on every distinct element to a persistent set (thus creating lots of sets that will have to be GC'ed) as well as building a vector via transients, while this new version will just build a set via transients.  See [this optimization in into](https://github.com/clojure/clojure/blob/clojure-1.9.0/src/clj/clojure/core.clj#L6820).

Nevertheless, since there was a non-comment change since the last iteration, I'd like to make sure we've benchmarked this final version against the test we were using for the previous version - did you do so @EthanEChristian ?  The main benefit here might not show up in a benchmark easily, since it would be to decrease the number of objects created that might not be GC'ed until later, but I'd like to make sure we didn't inadvertently cause a regression.

EthanEChristian: For the most part in the real world Sessions that i have been benchmarking this against, this change has decreased the time spent in add-conjunction slightly.

I took some time to profile the generative rules mentioned [here](https://github.com/cerner/clara-rules/pull/378#issuecomment-359600045), this time with 20000 productions.

The file below contains VM snapshots gathered using the current state of clara, versus the current state of this branch.
[ISSUE_377_NON_IP.zip](https://github.com/cerner/clara-rules/files/1755287/ISSUE_377_NON_IP.zip)

WilliamParker:
Merged as performance trials suggest that the minor change discussed at https://github.com/cerner/clara-rules/pull/378#discussion_r169125127 was mildly positive (i.e. it decreased the runtime), the most important thing being that it wasn't negative.

Issue #379, opened 2018-01-30
Fix error thrown when a common condition throws an exception in the LHS


EthanEChristian:
Given:
```
(defrule rule-1
  [Fact1]
  [Fact2 (< some-field 5)]
  =>
  (do-something))

(defrule rule-2
  [Fact3]
  [Fact2 (< some-field 5)]
  =>
  (do-something-else))
```

While inserting `(->Fact2 nil)` i would expect this to throw an error in the evaluation of the conditions in the LHS of both rules, and report the fact, conditions and rules that contain the conditions. This functionality was enhanced as part of #255.

However the actual result is another error being thrown:
```
java.lang.ClassCastException: clojure.lang.PersistentTreeSet cannot be cast to java.lang.Comparable
```

This is due to mishandling the scenario when a condition is common to multiple rules, see [get-conditions-and-rule-names](https://github.com/cerner/clara-rules/blob/0.16.1/src/main/clojure/clara/rules/engine.cljc#L218). Rather than `conj`, i think this should be `into`.

EthanEChristian:
#380

WilliamParker:
This looks like a pretty straightforward fix so I went ahead and merged the PR; we can leave the issue open for another day or two in case anyone else has thoughts on this after which I'll plan to close it.

PR #380, opened 2018-01-30
ISSUE-379: Fix error thrown when a common condition throws an excepti…

EthanEChristian:
…on in the LHS

#379

### DIFF
--- a/src/test/clojure/clara/test_rules.clj
+++ b/src/test/clojure/clara/test_rules.clj
@@ -2371,6 +2371,31 @@
                         (fire-rules)
                         (query temperature-query1)))))
 
+;; See https://github.com/cerner/clara-rules/issues/379 for more info
+(deftest test-single-condition-multiple-rule-constraint-exception
+  (let [rule-one (dsl/parse-rule [[WindSpeed
+                                   (= ?location location)]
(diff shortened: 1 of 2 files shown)


src/test/clojure/clara/test_rules.clj
```diff
@@ -2371,6 +2371,31 @@
...
+        temperature-fact (->Temperature nil "MCI")]
+
+    (assert-ex-data {:bindings nil
+                     :fact temperature-fact
+                     :conditions-and-rules
+                     {[clara.rules.testfacts.Temperature '(> temperature 0)]
+                      (sorted-set [:production  "rule-1"] [:production "rule-2"])}}
+                    (-> (mk-session [temperature-rule1 temperature-rule2] :cache false)
+                        (insert temperature-fact)
+                        (fire-rules)))))
```
WilliamParker: This test also has the nice side benefit of ensuring that we actually are sharing the node for the 
````
[Temperature (> temperature 0)]
````
condition since otherwise we wouldn't get both rules in the exception data.  Clean ways to test node sharing are something I've struggled with before.. this pattern might be worth considering for such tests in the future.

Issue #371, opened 2018-01-04
Allow qualified keyword names for rules and queries


sparkofreason:
Third-part DSLs can benefit from having production names specified as qualified keywords rather than symbols, e.g. allows query names to be easily correlated with specs defining the query results. The current code assumes that a production name will take the namespace where it is defined, which won't generally be the case using qualified keywords. The changes are pretty minor, see example from my working fork: https://github.com/Provisdom/clara-rules/commit/6dd773b43b4ef14ee4fa2d774f0235d33e171410

If possible, I'd like to just include this change as part of PR #365.

WilliamParker:
This sounds reasonable to me.

## Necessary changes ##
- The RHS compilation uses the :ns-name, not the namespace of the :name, in both [Clojure](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/compiler.clj#L1400) and [ClojureScript](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/macros.clj#L190).  Note that as discussed in [this comment on Issue 178](https://github.com/cerner/clara-rules/issues/178#issuecomment-214278167) similar logic was not applied to the LHS.  Therefore this qualification logic shouldn't be impacted by the proposed change.
- [This logic for naming extracted rules](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/compiler.clj#L987) that was added to solve [issue 149](https://github.com/cerner/clara-rules/issues/149) would need to be updated to accept a keyword name and create some reasonably new name.
- Schemas for :name will need to be updated to be either a keyword or string instead of just a string.
- If a user defines rules using defrule and defquery the fact that Clojure replaces duplicate var definitions makes it less likely to have multiple productions with the same name, although still possible since the rules are just data.  Having qualified keywords would probably make it easier to have such duplicates.  I'd suggest adding enforcement of name uniqueness on a per-session level in the compiler if we make this proposed change.  An exception would be thrown if duplicate names were used.

## User experience ##

- End users who just use Clara's defrule shouldn't be impacted by this.  Clara's defrule and defmacro will continue to produce the same outputs, along with everything that consumes them.  We should make sure to document appropriately that the :name will only be a keyword if the user is doing something beyond Clara's out-of-the-box functionality so that users don't think they have to update any data processing code they have to accept both strings and keywords if their use-case won't create keyword names.
[...]

PR #382, opened 2018-02-06
ISSUE-371: Allow qualified keyword names for rules and queries

sparkofreason:
See #371.


sparkofreason:
Unfortunately, the changes for this branch and #362 were done in an overlapping manner, hence the merge conflicts. Sorry about that.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1721,6 +1732,7 @@
         ;;
         ;; Note that this ordering is not for correctness; we are just trying to increase consistency of rulebase compilation,
         ;; and hopefully thereby execution times, from run to run.
+        productions (names-unique productions)
```
WilliamParker: The output of names-unique isn't really useful since it just returns its input or throws an exception, but someone reading through this code would need to look at the implementation of names-unique to know that.  Perhaps just 

````
_ (names-unique productions)
````
This currently doesn't actually change the value of the "productions" binding later but it looks like it could, which is unnecessary obfuscation IMO.

File: src/main/clojure/clara/rules/compiler.clj [hunk 2]
```diff
@@ -1759,6 +1771,14 @@
                    :listeners (get options :listeners  [])
                    :get-alphas-fn get-alphas-fn})))
 
+(defn add-production-load-order
+  "Adds ::rule-load-order to metadata of productions. Custom DSL's may need to use this if
+   creating a session in Clojure without calling mk-session below."
+  [productions]
+  (map (fn [n production]
+         (vary-meta production assoc ::rule-load-order (or n 0)))
+       (range) productions))
```
WilliamParker: I'm hesitant to expand the surface area of functions that users would call directly in the compiler namespace, since most of these functions are very much internal implementation details of Clara and should really be considered "project private".  Perhaps we could add this in the DSL namespace or somewhere else instead?

Thoughts anyone?

src/test/clojure/clara/test_keyword_names.clj
```diff
@@ -0,0 +1,48 @@
+(ns clara.test-keyword-names
```
WilliamParker: I think we can test this in ClojureScript using a strategy similar to [this](https://github.com/cerner/clara-rules/blob/0.17.0/src/test/clojurescript/clara/test_rules.cljs#L97) which pulls its rules data from [here](https://github.com/cerner/clara-rules/blob/0.17.0/src/test/clojurescript/clara/test_rules_data.clj#L5).  Basically raw Clojure data can be passed to the ClojureScript compiler path to create sessions.  We might not be able to test the case of failing due to duplicate production names so easily in ClojureScript though since we'd have to validate that ClojureScript compilation fails.  Every way I can think of to do that right now would be hard enough that it likely wouldn't be worthwhile.

src/test/clojure/clara/test_keyword_names.clj
```diff
@@ -0,0 +1,48 @@
...
+;;; and run a query with keyword-named productions.
+(deftest test-simple-insert
+  (let [session (-> (mk-session [is-cold-and-windy find-cold-and-windy])
+                    (insert (facts/->Temperature 15 "MCI"))
+                    (insert (facts/->WindSpeed 45 "MCI"))
+                    (fire-rules))]
+
+    (is (= #{{:?fact (facts/->ColdAndWindy 15 45)}}
+           (set
+             (query session ::find-cold-and-windy))))))
```
WilliamParker: Minor: Since there is only one result expected you can test for equality with a sequence, correct?

When there is more than one expected output, I like to use the [frequencies function](http://clojuredocs.org/clojure.core/frequencies) to assert on the count of things in sequences as well as what items are present.  For example

````
(is (= (frequencies [1 2 2]) (frequencies output-tested)))
````

File: src/main/clojure/clara/rules/dsl.clj [hunk 5]
```diff
@@ -294,7 +294,46 @@
   ([lhs rhs properties]
      (parse-rule* lhs rhs properties &env)))
 
+;;; added to clojure.core in 1.9
+(defn- qualified-keyword?
+  "Return true if x is a keyword with a namespace"
+  [x] (and (keyword? x) (namespace x) true))
+
```
WilliamParker: Don't we need schema changes for [Rule](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/schema.cljc#L66) and [Query](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/schema.cljc#L66)?  It is interesting that we didn't fail on them - we probably aren't making appropriate assertions, which may be legitimate since we've seen schema checks be a perf problem in Clara before - but I'd rather have the schemas be correct in any case since they are also a form of documentation.

src/main/clojure/clara/macros.clj
```diff
@@ -213,6 +205,10 @@
 
 (defn productions->session-assembly-form
   [productions options]
+  ;;; Calling com/names-unique here rather than in sources-and-options->session-assembly-form
+  ;;; as a ClojureScript DSL may call productions->session-assembly-form if that DSL
+  ;;; has its own mechanism for grouping rules which is different than the clara DSL.
```
WilliamParker: +1 comments like this are helpful

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: I think it is fairly idiomatic to have functions named like `validate-x` that just returned `x` unchanged or threw an exception if invalid.

It may still make sense to bind to underscore (ie ignored) if that helps readability though.

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: It seems like perhaps there'd be benefit of having a `clara.internal.*` ns prefix if the number of these sorts of things grew. That'd at least make it more clear where the boundaries are.

In Leiningen, there is an `:internal` metadata key added to indicate these sorts of situations. That seems reasonable to me, but would be something to document since there are no enforcement around that sort of thing. It's just a convention.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ This is a history of changes to clara-rules.
 * Fix issue with incorrect namespace qualification of rule and query code in ClojureScript. See [issue 359](https://github.com/cerner/clara-rules/issues/359).
 * Add clear-ns-productions! functionality to support clean reloading of rule and query definitions. See [issue 316](https://github.com/cerner/clara-rules/issues/316) for details.
 * Support session inspection and fact-graph in ClojureScript. See [issue 307](https://github.com/cerner/clara-rules/issues/307) for details.
+* Refactor defrule and defquery to better support customization. See [issue 362](https://github.com/cerner/clara-rules/issues/362) for details.
```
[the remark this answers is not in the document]
sparkofreason: I think that line is correct. The entry for this PR is under 0.18.0-SNAPSHOT, though did find a minor error there which I will fix.

src/main/clojure/clara/rules/dsl.clj
sparkofreason: Did you want the changes in this PR, or is this a separate issue?

src/main/clojure/clara/rules/compiler.clj
sparkofreason: Done.

src/test/clojure/clara/test_keyword_names.clj
sparkofreason: Done.

src/test/clojure/clara/test_keyword_names.clj
sparkofreason: Done.

sparkofreason:
Also I somehow made a mess of the history, though I think the resulting code is fine. Not exactly sure what happened there. IntelliJ is increasingly making a mess of my git settings, so it might have switched a default behind my back or something when I did an upstream fetch.

src/main/clojure/clara/rules/dsl.clj
WilliamParker: In this PR please since it is basically changing the documentation to reflect the changes already in the PR.  To be clear, I'm not saying that we need to increase schema validation etc. in this PR.  Just changing the schemas in the appropriate places from accepting a String to accepting either a String or keyword is sufficient IMO.

src/test/clojure/clara/test_keyword_names.clj
WilliamParker: There seems to be some duplication of rule and query definitions that isn't necessary.  The Clojure version should be able to use the same data definitions of the rules, so we shouldn't need to have more than one definition of each rule/query, and I don't see any reason to need more than two (clj and cljs).  You mentioned having some oddities with Git, perhaps multiple iterations on this were combined in your branch?

The way record types are referenced in Clojure and ClojureScript code forms is different, but we can get around that by using a custom :fact-type-fn that, say, looks at a specific field on a map that is expected to be a keyword.  I don't see immediately see anything else that would prevent sharing the rule/query structures between the Clojure and ClojureScript tests.

src/test/clojure/clara/test_keyword_names.clj
```diff
@@ -0,0 +1,47 @@
...
+                    (insert (facts/->WindSpeed 45 "MCI"))
+                    (fire-rules))]
+
+    (is (= [{:?fact (facts/->ColdAndWindy 15 45)}]
+           (query session ::find-cold-and-windy)))))
+
+;;; Verify that an exception is thrown when a duplicate name is encountered.
+;;; Note we create the session with com/mk-session*, as com/mk-session allows
+;;; duplicate names and will take what it considers to be the most recent
+;;; definition of a production.
```
WilliamParker: This can just be inside a clj-only conditional if added to a shared cljc test file; see my comment about increasing test sharing above.

src/test/clojure/clara/test_keyword_names.cljc
```diff
@@ -0,0 +1,58 @@
+(ns clara.test-keyword-names
```
WilliamParker: It is possible that I'm missing something, but I don't think this file will actually be used in ClojureScript since it is in the src/test/clojure directory and it [looks from the project.clj](https://github.com/cerner/clara-rules/blob/0.17.0/project.clj#L45) like only src/test/clojurescript and src/test/common will be used in ClojureScript tests.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Hmm.. so on reflection having the compiler require the DSL namespace seems odd to me.  We can probably just leave this in the compiler for the moment.  It is certainly not the only place where the boundaries between public APIs and implementation details are somewhat ambiguous, particularly since it isn't necessarily just a case of labeling functions public vs private.  A major discussion around that is out of scope here IMO so we can just set it aside for the time being.  We might want to move this function later, but if the change would just be to require the same function from another namespace that probably wouldn't be too disruptive to the probably-limited pool of users that would call this function.

CHANGELOG.md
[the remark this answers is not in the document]
WilliamParker: You're correct, never mind.

src/test/clojure/clara/test_keyword_names.cljc
sparkofreason: I think that was left over from when I was trying to get the test to run in both CLJ and CLJS. I'll remove it.

src/main/clojure/clara/rules/dsl.clj
sparkofreason: Done. While I was in there I updated the use of `s/either` to `s/cond-pre` to get rid of the deprecation warning.

sparkofreason:
Refactored the rule definitions so both the CLJ and CLJS tests around keyword names share test_rules_data.clj. I also updated to just transform the name so we didn't have the other 95% of the definitions repeated. I had to move test_rules_data.clj to live under the `clara.rules` hierarchy so it was available to all of the tests, similar to testfacts.cljc, which affected at least one other test (test_salience.cljs). Got rid of the dedicated test_keyword_names.* files, tests now live in test_rules.clj and test_rules.cljs. It may be possible to consolidate those - it wasn't clear to me that we could make that work with CLJC, plus they are slightly different (`defsession` used in CLJS, `mk-session` in CLJ).

src/test/clojure/clara/test_rules.clj
```diff
@@ -2675,3 +2701,25 @@
...
+                    (fire-rules))]
+    (is (= [{:?fact (->ColdAndWindy 15 45)}]
+           (query session ::rules-data/find-cold-and-windy-data)))))
+
+;;; Verify that an exception is thrown when a duplicate name is encountered.
+;;; Note we create the session with com/mk-session*, as com/mk-session allows
+;;; duplicate names and will take what it considers to be the most recent
+;;; definition of a production.
+(deftest test-duplicate-name
+  (try
```
WilliamParker: Let's use assert-ex-data instead.  That has the benefit of, among other things being less brittle if another layer of try-catch is added.  I think I just didn't notice this before, sorry for the late changes.

src/main/clojure/clara/rules/test_rules_data.clj
```diff
@@ -0,0 +1,31 @@
+(ns clara.rules.test-rules-data
```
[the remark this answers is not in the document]
sparkofreason: When I put it in src/test/common, I got errors that it couldn't be found. I found that puzzling, maybe I just had a typo or something, will give it another try.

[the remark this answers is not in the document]
WilliamParker: If you still run into errors that you can't find a reason for with this in src/test/common please post them, I'll take a look and see if I have some insight into their cause(s).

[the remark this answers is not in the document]
sparkofreason: clojure.lang.ExceptionInfo: failed compiling file:src/test/clojurescript/clara/test_rules.cljs {:file #object[java.io.File 0x1a314f8b "src/test/clojurescript/clara/test_rules.cljs"]}
...
Caused by: clojure.lang.ExceptionInfo:  at line 97 src/test/clojurescript/clara/test_rules.cljs {:file "src/test/clojurescript/clara/test_rules.cljs", :line 97, :column 1, :tag :cljs/analysis-error}
...
Caused by: java.lang.NullPointerException

I assume this is because when `defsession` tries to eval `(clara.test-rules-data/weather-rules)` it's not found, because the common folder doesn't get AOT'ed while the main source folder is all compiled before the tests are run. So I tried moving it back into the clojurescript folder, where it started. Still doesn't work.

sparkofreason:
I didn't have any luck getting the tests to work when the test_rules_data.clj file was moved out of the main source tree. If you have any suggestions there, I'll give them a try. I pushed the other fixes, along with a couple of minor cleanup things I found.

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: >  so on reflection having the compiler require the DSL namespace seems odd to me

Yeah, I think that sounds wrong.

src/main/clojure/clara/rules/dsl.clj
```diff
@@ -294,7 +294,43 @@
...
+(defn- production-name
+  [name]
```
mrrodriguez: In functions like this (especially), it just seems too risky to shadow `clojure.core/name`. I know it is fully-qualified throughout the body, but just error-prone. 
We could just call the arg `prod-name`

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -215,7 +215,7 @@
      (->> node
          :children
          (map get-conditions-and-rule-names)
-         (reduce (partial merge-with conj) {})))))
```
mrrodriguez: So this was wrong previously?

src/main/clojure/clara/rules/test_rules_data.clj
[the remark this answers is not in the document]
mrrodriguez: I'm not convinced that `defsession`s use of `eval` within the macro is a good practice. I've had similarly difficult issues with upgrading the Clara project's cljs to use a more modern Figwheel setup. I plan to work on that some again soon.

In this particular case, it doesn't make sense to me that the `clara.test-rules-data` ns is being brought in via `:require-macros`. It isn't being used for macros. It is being used to call a fn.

I think that the namespace should be moved back to "src/test/common/clara/test_rules_data.cljc". So use a cljc file and then just do a `:require` on it from the cljs side. If there is any clj specifics in it try using read-conditionals.

src/main/clojure/clara/rules/engine.cljc
sparkofreason: This is a result of my merge snafu. This change as actually made for #379, in commit https://github.com/Provisdom/clara-rules/commit/3c9fc4e5a6bf1d9e2fd3a3c65d6fa78c1ddf0a29#diff-00da05e66726186c05209662d990b97c.

src/main/clojure/clara/rules/dsl.clj
sparkofreason: Agreed, will fix.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: Thanks for the reminder. I see now.

src/main/clojure/clara/rules/test_rules_data.clj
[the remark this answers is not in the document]
sparkofreason: I absolutely agree, and explored multiple avenues, but could not get it to work in any other but the current form. I think there may be two separate issues. One is that for reasons unknown, even leaving everything exactly as is and putting test_rules_data.clj in common causes compile errors, which surprises me, but I assume is related to the same reason clara.rules.testfacts is not included in the test tree. The other reason I presume has to do with the `defsession` operation as you describe. The data used to create the session has to be available to the `defsession` macro at compile-time, which means it can't just be a `def` in CLJS or CLJC. I assume the use of `require-macros` on a CLJ file is a hack to get that data into scope at compile time. Even directly referencing the symbols like `the-rules` from `defsession` doesn't work. It has to be the form seen here, a call to a function which just returns the value referenced by the symbol. I assume that has something to do with how `eval` operates.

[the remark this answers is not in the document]
mrrodriguez: I think the `clara.rules.testfacts` thing is unrelated. That was some older issue with how Clara was resolving the `defrecord` types, which I think is no longer an issue and those should have been moved. I'm pretty sure it wasn't an issue (at least not at the time) with cljs compilation.

I'm glad you pointed out that data for `defsession` would have to be available at compilation time and that `require-macros` was probably trying to workaround that. That definition sounds like an abuse of `require-macros` to me. I'm not sure this pattern of bringing in the common rules to `defsession` is a sustainable one. If anything, the rule data structure needs to be exposed via something that `defsession` is meant to consume, which I believe would be a collection of symbols referring to production data structures, or a symbol referring to a namespace (presumably with productions in it).
`defsession` is also odd in it's use of quotes on symbols in some cases, but not others. It sounds like that macro could use some work.

I think this way of passing the productions for testing to it needs work too though.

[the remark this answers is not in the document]
mrrodriguez: For now, I wonder if the `clara.test-rules-data` could be clj or cljc and written like
```clj
(ns clara.test-rules-data
  (:require [clara.rules]
            [clara.rules.testfacts]))

(def the-rules
  [{:doc "Rule to determine whether it is indeed cold and windy."
    :name "clara.test-rules-data/is-cold-and-windy-data"
    :lhs '[{:type clara.rules.testfacts/Temperature
            :constraints [(< temperature 20)
                          (== ?t temperature)]}
           {:type clara.rules.testfacts/WindSpeed
            :constraints [(> windspeed 30)
                          (== ?w windspeed)]}]
    :rhs '(clara.rules/insert! (clara.rules.testfacts/->ColdAndWindy ?t ?w))}

   {:name "clara.test-rules-data/find-cold-and-windy-data"
    :lhs '[{:fact-binding :?fact
            :type clara.rules.testfacts/ColdAndWindy
            :constraints []}]
    :params #{}}])

(defmacro weather-rules
  "Return some weather rules"
  []
  `'~the-rules)
```

Then in an example, like `clara.test-rules` on the cljs side
Add `:require` on `[clara.test-rules-data :as trd :include-macros true]`

then use like
```clj

(defsession my-session-data (trd/weather-rules))
```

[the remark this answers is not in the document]
sparkofreason: I tried something like that and couldn't get it to work, though to be fair there were potentially some other issues lurking in there, like getting the logic straight for handling the difference between how CLJ and CLJS reference records.

Personally, I'd rather concentrate effort on changing the underlying methods for CLJS compilation so this sort of thing isn't necessary.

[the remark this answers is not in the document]
mrrodriguez: > Personally, I'd rather concentrate effort on changing the underlying methods for CLJS compilation so this sort of thing isn't necessary.

Sure. That sounds like it could be a separate PR effort too. Just not sure what a good interim solution is rather than including this file in the main jar. Perhaps that's acceptable, or just copy paste this rule definition into the namespaces where it is needed, etc.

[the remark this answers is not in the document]
sparkofreason: Just to clarify, I think there are two separate issues here. One has to do with the location of the file, which I suspect is probably solvable with the right project settings or something. I just don't know where the problem lies. The other issue is a deeper one around CLJS compilation and hacks like the use of `require-macros`, where the "solution" is really larger in scope.

I agree that including test_rules_data.clj in the main build is gross, and definitely open to suggestions there.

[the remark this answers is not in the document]
mrrodriguez: Yes, 

> One has to do with the location of the file, which I suspect is probably solvable with the right project settings or something. I just don't know where the problem lies.

This is something I want to investigate in  #388 . I want to get a better handle on the project structure with regards to cljs builds.

> The other issue is a deeper one around CLJS compilation and hacks like the use of require-macros, where the "solution" is really larger in scope.

I agree. I'm hoping if I can work through #388 I may get a bit more insight since I want to look at these errors we are seeing so far a little closer. 

> I agree that including test_rules_data.clj in the main build is gross, and definitely open to suggestions there.

Not sure, perhaps @WilliamParker has more thoughts.
If we still have `clara.rules.testfacts` in the main jar, I guess this wouldn't be that different. I'd probably at least put a comment on it about it needing to be moved later and only there temporarily for testing reasons.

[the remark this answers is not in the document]
WilliamParker: Full stack trace: [Gist](https://gist.github.com/WilliamParker/077b50fbbbca75ccc89a545afee35df0)

I think the key lines here are: 

```

	at clara.rules.compiler$get_cljs_accessors.invokeStatic(compiler.clj:144)
	at clara.rules.compiler$get_cljs_accessors.invoke(compiler.clj:139)
	at clara.rules.compiler$get_fields.invokeStatic(compiler.clj:196)
	at clara.rules.compiler$get_fields.invoke(compiler.clj:189)
	at clara.rules.compiler$field_name__GT_accessors_used.invokeStatic(compiler.clj:325)
	at clara.rules.compiler$field_name__GT_accessors_used.invoke(compiler.clj:321)
```

Basically, we are 

- Passing ClojureScript code to a Clojure macro
- Evaluating that code in a Clojure environment.
- The return value from the previous step contains paths to ClojureScript objects, which we use information from the ClojureScript compiler to reason about.

This analysis is in a way outside the dependency DAG between ClojureScript namespaces, so it wouldn't surprise me if some assumption(s) made by the ClojureScript compiler are being broken.  It also makes me wonder if there is some way to expose this scenario without the main vs test code dichotomy here, which would be a bug.

For an immediate solution, I suspect using rules that don't operate on record types, but rather use primitive data structures, e.g. maps, would work.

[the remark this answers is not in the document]
mrrodriguez: Just FYI, I think solving this issue is probably best a separate issue. I plan to look over some of the cljs testing stuff with #388 
I've already did some investigation of the problems we are seeing here. I can also cause problems in `master` with compiling these cljs test namespaces with no further changes.

The kicker is just that `lein-cljsbuild` seems to be able to successfully compile these test cljs namespaces, when the `cljs.build.api` and also figwheel cannot. I think something sneaky is just masking the problem in `lein-cljsbuild`. I plan to work more on this and have some more details on #388 within the next few days.

alex-dixon:
FWIW, in Precept, changes with namespace resolution in CLJS in 0.17 prevented some rules from being loaded at session creation time. The fix requires them in the CLJS namespace context. 
https://github.com/CoNarrative/precept/pull/114/files

Previously these were able to be loaded via a require in com/compiling-cljs?

WilliamParker:
I created a new PR with the merge conflicts addressed at https://github.com/cerner/clara-rules/pull/389

If there aren't any objections and I don't seen any problems later I'll plan to merge that one.

WilliamParker:
I merged https://github.com/cerner/clara-rules/pull/389 so I don't think we need this PR anymore as https://github.com/cerner/clara-rules/pull/389 contains the same changes with the merge conflicts addressed.  Closing.

Issue #383, opened 2018-02-12
Remove unused let-bindings in compiled RHS action functions


WilliamParker:
Take the following rule:

````
(defrule cold-rule
        [ColdAndWindy (= ?temperature temperature) (= ?windspeed windspeed)]
        =>
        (insert! (->Cold ?temperature)))
````

When we create a session with this rule and inspect the compiled RHS function, noting that the the unevaluated expression is [added as metadata to the ProductionNode](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/compiler.clj#L1403) for [later use in durability](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/durability.clj#L57)  :
````
(-> (mk-session [cold-rule]) .rulebase :alpha-roots (get ColdAndWindy) first :children first :children first meta)
````

we retrieve the following:

````
{:action-expr
 (clojure.core/fn
  [?__token__ ?__env__]
  (clojure.core/let
   [?temperature
    (get-in ?__token__ [:bindings :?temperature])
    ?windspeed
    (get-in ?__token__ [:bindings :?windspeed])]
   (do (insert! (->Cold ?temperature)))))}
````

Note that a let-binding for ?windspeed is created even though it is not used in the user's RHS code.  It does not appear that [the code which creates these RHS let-binding](https://github.com/cerner/clara-rules/blob/0.17.0/src/main/clojure/clara/rules/compiler.clj#L389) makes any attempt to determine what bindings are used in the RHS.  As a result unnecessary bindings are created, consuming time to create at runtime and increasing the amount of Clojure code evaluated at session creation time, thus likely increasing total session compilation time.

In order to fix this we can apply a similar fix to what was done in https://github.com/cerner/clara-rules/pull/180 for let bindings of fields that aren't actually used in constraints.  We'd basically just walk the user's RHS code and detect what bindings are present in it.  Since the bindings have to be literals in the code (rather than in outside user functions called by the RHS, where they aren't resolvable by Clojure's compiler) this can be done safely.

WilliamParker:
I've created a preliminary PR for this at https://github.com/cerner/clara-rules/pull/384.  Please don't merge it yet though.

Further items to address:

[...]

PR #384, opened 2018-02-27
Issue 383: Remove unused let-bindings from the generated RHS activati…

WilliamParker:
…on function

Draft/proposed changes for [issue 383](https://github.com/cerner/clara-rules/issues/383).

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -73,6 +73,12 @@
                         ;; Function that takes facts and determines what alpha nodes they match.
                         get-alphas-fn])
 
+(defn- is-variable?
+  "Returns true if the given expression is a variable (a symbol prefixed by ?)"
+  [expr]
+  (and (symbol? expr)
+       (.startsWith (name expr) "?")))
+
 (def ^:private reflector
   "For some reason (bug?) the default reflector doesn't use the
    Clojure dynamic class loader, which prevents reflecting on
@@ -309,8 +315,7 @@
   "Returns a set of the symbols in the given s-expression that start with '?' as keywords"
   [expression]
   (into #{} (for [item (flatten-expression expression)
-                  :when (and (symbol? item)
-                             (= \? (first (name item))))]
+                  :when (is-variable? item)]
               (keyword  item))))
 
 (defn field-name->accessors-used
@@ -386,7 +391,19 @@
 (defn compile-action
   "Compile the right-hand-side action of a rule, returning a function to execute it."
   [binding-keys rhs env]
-  (let [assignments (mapcat #(list (symbol (name %)) (list 'get-in '?__token__ [:bindings %])) binding-keys)
+  (let [;; Avoid creating let bindings in the compile code that aren't actually used in the body.
+        ;; The bindings only exist in the scope of the RHS body, not in any code called by it,
+        ;; so this scanning strategy will detect all possible uses of binding variables in the RHS.
+        ;; Note that some strategies with macros could introduce bindings, but these aren't something
+        ;; we're trying to support.  If necessary a user could macroexpand their RHS code manually before
+        ;; providing it to Clara.
+        rhs-bindings-used (variables-as-keywords rhs)
+
+        assignments (sequence
+                     (comp
+                      (filter rhs-bindings-used)
+                      (mapcat #(list (symbol (name %)) (list 'get-in '?__token__ [:bindings %]))))
+                     binding-keys)
 
         ;; The destructured environment, if any.
         destructured-env (if (> (count env) 0)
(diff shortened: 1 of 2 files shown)


src/main/clojure/clara/rules/compiler.clj
```diff
@@ -386,7 +391,19 @@
...
+        ;; so this scanning strategy will detect all possible uses of binding variables in the RHS.
+        ;; Note that some strategies with macros could introduce bindings, but these aren't something
+        ;; we're trying to support.  If necessary a user could macroexpand their RHS code manually before
+        ;; providing it to Clara.
+        rhs-bindings-used (variables-as-keywords rhs)
+
+        assignments (sequence
+                     (comp
+                      (filter rhs-bindings-used)
+                      (mapcat #(list (symbol (name %)) (list 'get-in '?__token__ [:bindings %]))))
```
mrrodriguez: It isn't really part of the change you are making here, but 2 things:

`get-in` shouldn't be left unqualified. There is no good reason to not ensure it is `clojure.core/get-in` instead of perhaps an overridden var with that same name in the compilation ns.

That said, I think using `get-in` here at all is unnecessarily slow. This is a 2 keyword lookup and should be done in a more direct manner.
```clj
(mapcat #(list (symbol (name %)) (-> '?__token__ :bindings %)))
```

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -309,8 +315,7 @@
   "Returns a set of the symbols in the given s-expression that start with '?' as keywords"
   [expression]
   (into #{} (for [item (flatten-expression expression)
-                  :when (and (symbol? item)
```
mrrodriguez: Nice

src/test/common/clara/test_performance_optimizations.cljc
```diff
@@ -0,0 +1,68 @@
+;; These tests validate that operations that the rules engine should optimize
+;; away are in fact optimized away.  The target here is not the actual execution time,
+;; which will vary per system, but verification that the action operations in question are not performed.
+#?(:clj
+   (ns clara.test-performance-optimizations
```
mrrodriguez: I think it's pretty heavy-weight to fully repeat the whole `ns` form across the 2 read conditionals.
Ideally we'd be minimizing the need to even specialize most parts of it for clj vs cljs.

For example, `:require-macros` should be unnecessary in most cases nowadays.

Another good simplification is to be pretty aggressive in the use of ns aliases instead of `:refer`. That can make it so that the need for any "<something>-macros" cljs declarations are unnecessary too.

This may not be that important for the sake of this review. I just think this will become an annoying pattern to maintain/work with.

src/test/common/clara/test_performance_optimizations.cljc
```diff
@@ -0,0 +1,68 @@
...
+(use-fixtures :once st/validate-schemas #?(:clj tu/opts-fixture))
+(use-fixtures :each tu/side-effect-holder-fixture)
+
+;; Note that if the binding is present, this will show up as an error (since an exception is thrown)
+;; rather than a failure.  This is not ideal but since this will happen at compile time in ClojureScript,
+;; rather than test execution time, we can't just register a test failure via the clojure.test/is macro.
+#?(:clj
+   (defmacro fail-if-binding-present
+     []
+     (when (contains? &env '?unused-binding)
```
mrrodriguez: Couldn't the macro return a form (just a boolean even) that indicated whether it was a success or failure? The RHS could then use that to conditional do something (eg insert `Cold` one way vs another) that is testable.

```clj

(defmacro unused-binding-present? []
  (contains? &env '?unused-binding))
```

src/main/clojure/clara/rules/compiler.clj
WilliamParker: +1 to both points.  I will make the change.

src/test/common/clara/test_performance_optimizations.cljc
WilliamParker: +1 that should work nicely.  I'll make the change.

src/test/common/clara/test_performance_optimizations.cljc
WilliamParker: I can take a look at this, either as part of this PR or otherwise.  When I initially started creating these test namespaces I wasn't able to find a way to share enough to make it seem worthwhile - IIRC around 80% of the code in the header was in conditionals when I tried to make it a single header.  However I could have been missing some options in how to create CLJS ns headers.

```
Another good simplification is to be pretty aggressive in the use of ns aliases instead of :refer. That can make it so that the need for any "-macros" cljs declarations are unnecessary too.
```
How does the usage of aliases allow removal of require-macros declarations?

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -73,6 +73,12 @@
                         ;; Function that takes facts and determines what alpha nodes they match.
                         get-alphas-fn])
 
+(defn- is-variable?
```
Zylox: Dont we call these "unification variables/bindings" elsewhere? variable seems...vaguer than its actual intent.

Zylox:
👍

EthanEChristian:
👍

PR #387, opened 2018-03-11
Issue 386: Remove unneeded memory operations from ProductionNode

WilliamParker:
This is an incomplete draft but demonstrates the design of the changes.

TODOs:

- Potentially rename new functions, keywords, etc.  I didn't put much thought into the names at this stage.
- Write more docstrings and code comments as appropriate.
- Fix the tracing tests.
- Optional (could be a later PR): Add additional session inspection keys using the new listener e.g. explanations of unconditionally inserted facts and/or retractions.


src/main/clojure/clara/rules/engine.cljc
```diff
@@ -337,17 +337,17 @@
   ILeftActivate
   (left-activate [node join-bindings tokens memory transport listener]
 
+    ;; Provide listeners information on all left-retract calls for passivity,
```
mrrodriguez: Just a minor thing, but I think it is more important than just preserving old behavior. I think it is really useful behavior to be able to listen for this activity in the ProductionNode still.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1766,6 +1767,11 @@
                     ;; Therefore, the reordering of retractions and insertions should have no impact
                     ;; assuming that the evaluation of rule conditions is pure, which is a general expectation
                     ;; of the rules engine.
+                    (l/fire-activation! listener
+                                        activation
```
mrrodriguez: These extra `deref` calls are going to incur runtime cost. That's sort of an unfortunate part of the listener calls in the engine all being functions (as opposed to macros that could try to mitigate the cost when they are "off").

In this case, I think You could just bind to the `deref`ed values here and use them in the places you need them. They are needed outside of the `l/fire-activation!` call anyways.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -2002,6 +2008,29 @@
...
+  (let [{:keys [listeners] :as components} (components session)]
+    (assemble (assoc components
+                     :listeners
+                     (conj listeners
+                           listener)))))
+
+(defn remove-listeners
+  [session pred]
+  (let [{:keys [listeners] :as components} (components session)
+        matching-listeners (filterv pred listeners)]
```
mrrodriguez: I think using `some` here would make more sense. No reason to build a vector of all the matches that aren't used later.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -2002,6 +2008,29 @@
...
+    (if (seq matching-listeners)
+      (assemble (assoc components
+                       :listeners
+                       (remove pred listeners)))
```
mrrodriguez: This seems like something that'd be better off not lazy. It just seems questionable when it will be realized later.
`(into [] (remove pred) listeners)`

src/main/clojure/clara/tools/internal/inspect.cljc
```diff
@@ -0,0 +1,76 @@
+(ns clara.tools.internal.inspect
+  (:require [clara.rules.listener :as l]
+            [clara.rules.engine :as eng]))
+
+(declare to-persistent-listener)
+
+(deftype TransientRuleActivationListener [activations]
+  l/ITransientEventListener
+  (fire-activation! [listener activation resulting-operations]
+    (reset! (.-activations listener) (conj @(.-activations listener) {:activation activation
```
mrrodriguez: This is a slower version of `swap!` I believe
```clj

(swap! (.-activations listener) conj {:activation activation :resulting-operations resulting-operations})

```

src/main/clojure/clara/tools/internal/inspect.cljc
```diff
@@ -0,0 +1,76 @@
...
+  (remove-accum-reduced! [listener node join-bindings fact-bindings]
+    listener)
+  (add-activations! [listener node activations]
+    listener)
+  (remove-activations! [listener node activations]
+    listener)
+  (fire-rules! [listener node]
+    listener))
+
+(deftype PersistentRuleActivationListener [activations]
```
mrrodriguez: I wonder if "rule" is needed in these names. Could it all just be "activation listener" instead of "rule activation listener"?

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -2002,6 +2008,29 @@
...
+      session)))
+
+(defn listeners-matching-pred
```
mrrodriguez: Perhaps`find-listeners`?

rbrush:
+1 to this change. (And thanks to Will for giving me a heads up here.)

PR #392, opened 2018-04-05
Don't override Clojure 1.9's `qualified-keyword?`

bfontaine:
When using Clojure 1.9 and Clara Rules 0.18.0, requiring `clara.rules.dsl` or any namespace that requires it results in the following warning:

    WARNING: qualified-keyword? already refers to: #'clojure.core/qualified-keyword? in namespace: clara.rules.dsl, being replaced by: #'clara.rules.dsl/qualified-keyword?

This change removes the inclusion of Clojure 1.9’s `qualified-keyword?` from this namespace, and thus the warning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 This is a history of changes to clara-rules.
 
+### 0.19.0-SNAPSHOT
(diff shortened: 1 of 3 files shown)


WilliamParker:
Thanks @bfontaine sorry about the oversight.

Issue #381, opened 2018-02-06
[performance] Reduce number of calls to eval by the compiler


mrrodriguez:
In #378 there was discussion about `eval` becoming a large bottleneck in the Clara compiler when it comes to large rulesets.
Some relevant comments are:
* https://github.com/cerner/clara-rules/pull/378#issuecomment-360597728
* https://github.com/cerner/clara-rules/pull/378#issuecomment-360601737
* https://github.com/cerner/clara-rules/pull/378#issuecomment-363284707

(1) As Ryan mentioned there, I think that one avenue we could explore would be to "merge" all of the `eval` code into one unified structure prior to calling `eval` so that we ended up with only on batch call. 

(2) I also have noticed that accumulator forms are always `eval`ed with no attempt to avoid duplicate forms being `eval`ed independently. If a large ruleset used the same accumulator form often, this may add up to a lot of unnecessary `eval`s.
A duplicate accumulator could be detected by analyzing the form along with any lexical environment being captured at the time.

(3) There may even be ways to generalize (2) in an attempt to get more sharing from join filter functions on other beta nodes. These forms are a bit more complex though, so care would have to be taken. One particular thing that would need to be handled carefully would be in the use of metadata on parts of the form, such as those for type hints. Type hints could cause 2 forms to not really be equal for the sake of `eval`.

It should be noted that any change to `eval` structures would like affect the Clara durability.

mrrodriguez:
@rbrush 

> For instance, what if we had something like an "extract-beta-functions-to-eval" that got everything we needed to eval from the beta graph, then eval them all at once. Then the existing compile-beta-graph would look up the corresponding function and use it, rather than calling eval each time. This would eliminate runtime overhead since it still calls the function directly...we're just eval'ing them all together into larger structure that is used at compile time.

Thanks for that clarification! I was not thinking about it correctly. I overlooked that these `eval`s could be made in batch, but still directly "linked" with their appropriate nodes during the compilation cycle (and not at runtime). I'll edit the description to not have that misleading note then.

WilliamParker:
[...]

PR #396, opened 2018-06-19
Issue #381 Reduce the number of Eval calls

EthanEChristian:
I have changed both the Durability and Compiler evaluation of functions that are embedded in the rulebase, as part of this I also gave AlphaNodes an Id. 

I have also added some performance tests that generate a simple rulebase and compile, serialize and deserialize the rulebase. Running performance locally (`lein test :performance` )
gave me a before of:
```
Running Performance tests for:
Generated Session Compilation
==========================================
Mean: 13412.32ms
Standard Deviation : 5924.659171429188ms

Running Performance tests for:
Generated Session serialization
==========================================
Mean: 205.94ms
Standard Deviation : 85.99474635115799ms

Running Performance tests for:
Generated Session deserialization
==========================================
Mean: 29966.56ms
Standard Deviation : 12047.334043945159ms
```

and an after of:
```
Running Performance tests for:
Generated Session Compilation
==========================================
Mean: 2682.92ms
Standard Deviation : 108.87604695248626ms


Running Performance tests for:
Generated Session serialization
==========================================
Mean: 223.9ms
Standard Deviation : 11.80211845390479ms


Running Performance tests for:
Generated Session deserialization
==========================================
Mean: 1683.98ms
Standard Deviation : 87.03987362123178ms
```

I chose to persist the functions on the rulebase, this allows for the functions to be easily pulled off for serialization and deserialization. Allowing for minimal traversal of the already constructed rulebase, memory usage seems comparable to what it was before and the performance numbers speak for themselves.

@WilliamParker @mrrodriguez @rbrush 
When you get some time, could you look over this?

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -68,8 +68,15 @@
   ;; perhaps because Leiningen is using this as uneval'ed code.
   ;; For now just duplicate the line.
   :test-selectors {:default (complement (fn [x]
-                                          (some->> x :ns ns-name str (re-matches #"^clara\.generative.*"))))
-                   :generative (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.generative.*")))}
+                                          (let [backlisted-packages #{"generative" "performance"}
+                                                patterns (into []
+                                                           (comp
+                                                             (map #(str "^clara\\." % ".*"))
+                                                             (interpose "|"))
+                                                           backlisted-packages)]
+                                            (some->> x :ns ns-name str (re-matches (re-pattern (apply str patterns)))))))
+                   :generative (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.generative.*")))
+                   :performance (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.performance.*")))}
   
   :scm {:name "git"
         :url "https://github.com/cerner/clara-rules"}
--- a/project.clj
+++ b/project.clj
@@ -68,12 +68,12 @@
   ;; perhaps because Leiningen is using this as uneval'ed code.
   ;; For now just duplicate the line.
   :test-selectors {:default (complement (fn [x]
-                                          (let [backlisted-packages #{"generative" "performance"}
+                                          (let [blacklisted-packages #{"generative" "performance"}
                                                 patterns (into []
                                                            (comp
                                                              (map #(str "^clara\\." % ".*"))
                                                              (interpose "|"))
-                                                           backlisted-packages)]
+                                                           blacklisted-packages)]
                                             (some->> x :ns ns-name str (re-matches (re-pattern (apply str patterns)))))))
                    :generative (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.generative.*")))
                    :performance (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.performance.*")))}
--- a/src/main/clojure/clara/macros.clj
+++ b/src/main/clojure/clara/macros.clj
@@ -196,10 +196,11 @@
 (sc/defn ^:always-validate compile-alpha-nodes
   [alpha-nodes :- [schema/AlphaNode]]
   (vec
-   (for [{:keys [condition beta-children env]} alpha-nodes
+   (for [{:keys [id condition beta-children env]} alpha-nodes
          :let [{:keys [type constraints fact-binding args]} condition]]
 
-     {:type (com/effective-type type)
+     {:id id
+      :type (com/effective-type type)
       :alpha-fn (com/compile-condition type (first args) constraints fact-binding env)
       :children (vec beta-children)})))
 
@@ -209,11 +210,14 @@
   ;;; as a ClojureScript DSL may call productions->session-assembly-form if that DSL
   ;;; has its own mechanism for grouping rules which is different than the clara DSL.
   (com/validate-names-unique productions)
-  (let [beta-graph (com/to-beta-graph productions)
+  (let [id-counter (atom 0)
+        create-id-fn (fn [] (swap! id-counter inc))
+
+        beta-graph (com/to-beta-graph productions create-id-fn)
         ;; Compile the children of the logical root condition.
         beta-network (gen-beta-network (get-in beta-graph [:forward-edges 0]) beta-graph #{})
 
-        alpha-graph (com/to-alpha-graph beta-graph)
+        alpha-graph (com/to-alpha-graph beta-graph create-id-fn)
         alpha-nodes (compile-alpha-nodes alpha-graph)]
     
     `(let [beta-network# ~beta-network
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1391,7 +1391,12 @@
                                                              :join-filter-expressions (:join-filter-expressions beta-node)
                                                              :env (:env beta-node)
                                                              :msg "compiling accumulate with join filter node"}}))
-            :query prev)))
+            :query prev
+
+            ;; This error should only be thrown if there are changes to the compilation of the beta-graph
+            ;; such as an addition of a node type.
+            (throw (ex-info "Invalid node type encountered while compiling rulebase."
+                            {:node beta-node})))))
       id->expr
       (:id-to-condition-node beta-graph))))
 
@@ -1448,9 +1453,10 @@
   "A helper function for retrieving a given key from the provided map. If the key doesn't exist within the map this
    function will throw an exception."
   [m k]
-  (let [v (get m k ::not-found)]
-    (if (= v ::not-found)
-      (throw (ex-info "Key not found with safe-get" {:map map :key k}))
+  (let [not-found ::not-found
+        v (get m k not-found)]
+    (if (identical? v not-found)
+      (throw (ex-info "Key not found with safe-get" {:map m :key k}))
       v)))
 
 (sc/defn ^:private compile-node
@@ -1875,34 +1881,34 @@
       (throw (ex-info (str "Non-unique production names: " non-unique) {:names non-unique})))))
 
 (def forms-per-eval-default
-  ;; The default max number of forms that will be evaluated together as a single batch.
-  ;; 5000 is chosen here due to the way that clojure will evaluate the vector of forms extracted from the nodes.
-  ;; The limiting factor here is the max java method size (64KiB), clojure will compile each form in the vector down into
-  ;; its own class file and generate another class file that will reference each of the other functions and wrap them in
-  ;; a vector inside a static method. For example,
-  ;;
-  ;; (eval [(fn one [_] ...) (fn two [_] ...)])
-  ;; would generate 3 classes.
-  ;;
-  ;; some_namespace$eval1234
-  ;; some_namespace$eval1234$one_1234
-  ;; some_namespace$eval1234$two_1235
-  ;;
-  ;; some_namespace$eval1234$one_1234 and some_namespace$eval1234$two_1235 contian the implementation of the functions,
-  ;; where some_namespace$eval1234 will contain two methods, invoke and invokeStatic.
-  ;; The invokeStatic method breaks down into something similar to a single create array call followed by 2 array set calls
-  ;; with new invocations on the 2 classes the method then returns a new vector created from the array.
-  ;;
-  ;; 5000 is lower than the absolute max to allow for modifications to how clojure compiles without needing to modify this.
-  ;; The current limit should be 5471, this is derived from the following opcode investigation:
-  ;;
-  ;; Array creation:                                                    5B
-  ;; Creating and populating the first 6 elements of the array:        60B
-  ;; Creating and populating the next 122 elements of the array:    1,342B
-  ;; Creating and populating the next 5343 elements of the array:  64,116B
-  ;; Creating the vector and the return statement:                      4B
-  ;;
-  ;; This sums to 65,527B just shy of the 65,536B method size limit.
+  "The default max number of forms that will be evaluated together as a single batch.
+   5000 is chosen here due to the way that clojure will evaluate the vector of forms extracted from the nodes.
+   The limiting factor here is the max java method size (64KiB), clojure will compile each form in the vector down into
+   its own class file and generate another class file that will reference each of the other functions and wrap them in
+   a vector inside a static method. For example,
+
+   (eval [(fn one [_] ...) (fn two [_] ...)])
+   would generate 3 classes.
+
+   some_namespace$eval1234
+   some_namespace$eval1234$one_1234
+   some_namespace$eval1234$two_1235
+
+   some_namespace$eval1234$one_1234 and some_namespace$eval1234$two_1235 contian the implementation of the functions,
+   where some_namespace$eval1234 will contain two methods, invoke and invokeStatic.
+   The invokeStatic method breaks down into something similar to a single create array call followed by 2 array set calls
+   with new invocations on the 2 classes the method then returns a new vector created from the array.
+
+   5000 is lower than the absolute max to allow for modifications to how clojure compiles without needing to modify this.
+   The current limit should be 5471, this is derived from the following opcode investigation:
+
+   Array creation:                                                    5B
+   Creating and populating the first 6 elements of the array:        60B
+   Creating and populating the next 122 elements of the array:    1,342B
+   Creating and populating the next 5343 elements of the array:  64,116B
+   Creating the vector and the return statement:                      4B
+
+   This sums to 65,527B just shy of the 65,536B method size limit."
   5000)
 
 (sc/defn mk-session*
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1276,12 +1276,10 @@
   [beta-graph :- schema/BetaGraph
    alpha-graph :- [schema/AlphaNode]]
   (let [backward-edges (:backward-edges beta-graph)
-        handle-expr (fn [id->expr s-expr id expr-key metadata]
+        handle-expr (fn [id->expr s-expr id expr-key compilation-ctx]
                       (assoc id->expr
-                        ;; Letting the key carry all of the metadata, this makes the batch evaluation of the expressions
-                        ;; easier because the metadata contains unevalable forms.
-                        (with-meta [id expr-key] (assoc metadata expr-key s-expr))
-                        s-expr))
+                        [id expr-key]
+                        [s-expr (assoc compilation-ctx expr-key s-expr)]))
 
         ;; If extract-exprs ever became a hot spot, this could be changed out to use more java interop.
         id->expr (reduce (fn [prev alpha-node]
@@ -1403,54 +1401,55 @@
    See #381 for more details."
   [key->expr :- schema/NodeExprLookup
    partition-size :- sc/Int]
-  (let [batching-try-eval (fn [node-keys exprs]
+  (let [batching-try-eval (fn [compilation-ctxs exprs]
                             ;; Try to evaluate all of the expressions as a batch. If they fail the batch eval then we
                             ;; try them one by one with their compilation context, this is slow but we were going to fail
                             ;; anyway.
                             (try
-                              (eval exprs)
+                              (mapv vector (eval exprs) compilation-ctxs)
                               (catch Exception e
                                 ;; Using mapv here rather than map to avoid laziness, otherwise compilation failure might
                                 ;; fall into the throw below for the wrong reason.
-                                (mapv (fn [expr node-keys]
-                                        (let [cmeta (meta node-keys)]
-                                          (with-bindings
-                                            {#'*compile-ctx* (:compile-ctx cmeta)
-                                             #'*file* (:file cmeta *file*)}
-                                            (try-eval expr))))
+                                (mapv (fn [expr compilation-ctx]
+                                        (with-bindings
+                                          {#'*compile-ctx* (:compile-ctx compilation-ctx)
+                                           #'*file* (:file compilation-ctx *file*)}
+                                          (try-eval expr)))
                                       exprs
-                                      node-keys)
+                                      compilation-ctxs)
                                 ;; If none of the rules are the issue, it is likely that the
                                 ;; size of the code trying to be evaluated has exceeded the limit
                                 ;; set by java.
                                 (throw (ex-info (str "There was a failure while batch evaling the node expressions, " \newline
                                                      "but wasn't present when evaling them individually. This likely indicates " \newline
                                                      "that the method size exceeded the maximum set by the jvm, see the cause for the actual error.")
-                                                {:node-keys node-keys}
+                                                {:compilation-ctxs compilation-ctxs}
                                                 e)))))]
     (into {}
           cat
           ;; Grouping by ns, most expressions will not have a defined ns, only expressions from production nodes.
           ;; These expressions must be evaluated in the ns where they were defined, as they will likely contain code
           ;; that is namespace dependent.
-          (for [[nspace group] (sort-by key (group-by (comp :ns meta key) key->expr))
+          (for [[nspace group] (sort-by key (group-by (comp :ns second val) key->expr))
                 ;; Partitioning the number of forms to be evaluated, Java has a limit to the size of methods if we were
                 ;; evaluate all expressions at once it would likely exceed this limit and throw an exception.
                 expr-batch (partition-all partition-size group)
-                :let [node-keys (map key expr-batch)]]
+                :let [node-expr-keys (mapv first expr-batch)
+                      compilation-ctxs (mapv (comp second val) expr-batch)
+                      exprs (mapv (comp first val) expr-batch)]]
             (mapv vector
-                  node-keys
+                  node-expr-keys
                   (with-bindings (if nspace
                                    {#'*ns* (the-ns nspace)}
                                    {})
-                    (batching-try-eval node-keys (mapv val expr-batch))))))))
+                    (batching-try-eval compilation-ctxs exprs)))))))
 
 (defn safe-get
   "A helper function for retrieving a given key from the provided map. If the key doesn't exist within the map this
    function will throw an exception."
   [m k]
   (let [v (get m k ::not-found)]
-    (if (identical? v ::not-found)
+    (if (= v ::not-found)
       (throw (ex-info "Key not found with safe-get" {:map map :key k}))
       v)))
 
@@ -1470,7 +1469,7 @@
                     (.loadClass (clojure.lang.RT/makeClassLoader) (name condition))
                     condition)
 
-        compiled-expr-fn (fn [id field] (safe-get expr-fn-lookup [id field]))]
+        compiled-expr-fn (fn [id field] (first (safe-get expr-fn-lookup [id field])))]
 
     (case (:node-type beta-node)
 
@@ -1684,7 +1683,7 @@
   (for [{:keys [id condition beta-children env] :as node} alpha-nodes]
     (cond-> {:id id
              :type (effective-type (:type condition))
-             :alpha-fn (safe-get expr-fn-lookup [id :alpha-expr])
+             :alpha-fn (first (safe-get expr-fn-lookup [id :alpha-expr]))
              :children beta-children}
             env (assoc :env env))))
 
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1308,7 +1308,7 @@
                                             ;; ProductionNode expressions can be arbitrary code, therefore we need the
                                             ;; ns where the production was define so that we can compile the expression
                                             ;; later.
-                                            {:ns (some-> (:ns-name production) str)
+                                            {:ns (:ns-name production)
                                              :file (-> production :rhs meta :file)
                                              :compile-ctx {:production production
                                                            :msg "compiling production node"}})))
@@ -1322,7 +1322,7 @@
                           condition)]
           (case (or (:node-type beta-node)
                     ;; If there is no :node-type then the node is the ::root-condition
-                    ;; however, creating a case for nil could potentially casue weird effects if something about the
+                    ;; however, creating a case for nil could potentially cause weird effects if something about the
                     ;; compilation of the beta graph changes. Therefore making an explicit case for ::root-condition
                     ;; and if there was anything that wasn't ::root-condition this case statement will fail rather than
                     ;; failing somewhere else.
@@ -1390,8 +1390,9 @@
       (:id-to-condition-node beta-graph))))
 
 (sc/defn compile-exprs :- schema/NodeFnLookup
-  "Takes a map in form produced by extract-exprs and evaluates the values(expressions) of the map in a batched manner.
-   This allows the eval calls to be more effecient, rather than evaluating each expression on its own."
+  "Takes a map in the form produced by extract-exprs and evaluates the values(expressions) of the map in a batched manner.
+   This allows the eval calls to be more effecient, rather than evaluating each expression on its own.
+   See #381 for more details."
   [key->expr :- schema/NodeExprLookup
    partition-size :- sc/Int]
   (let [batching-try-eval (fn [node-keys exprs]
@@ -1413,24 +1414,24 @@
                                 ;; size of the code trying to be evaluated has exceeded the limit
                                 ;; set by java.
                                 (throw (ex-info "The batching size of expressions has exceeded the limit set by java."
-                                                {:orginal-exception e})))))]
+                                                {:node-keys node-keys}
+                                                e)))))]
     (into {}
           cat
           ;; Grouping by ns, most expressions will not have a defined ns, only expressions from production nodes.
           ;; These expressions must be evaluated in the ns where they were defined, as they will likely contain code
           ;; that is namespace dependent.
-          (for [[ns pairs] (group-by (comp :ns meta key) key->expr)
+          (for [[ns group] (sort-by key (group-by (comp :ns meta key) key->expr))
                 ;; Partitioning the number of forms to be evaluated, Java has a limit to the size of methods if we were
                 ;; evaluate all expressions at once it would likely exceed this limit and throw an exception.
-                pairs (partition-all partition-size pairs)
-                :let [node-keys (map #(nth % 0) pairs)]]
-            (mapv (fn [node-key expr]
-                    [node-key expr])
+                expr-batch (partition-all partition-size group)
+                :let [node-keys (map key expr-batch)]]
+            (mapv vector
                   node-keys
                   (with-bindings (if ns
-                                   {#'*ns* (-> ns symbol the-ns)}
+                                   {#'*ns* (the-ns ns )}
                                    {})
-                    (batching-try-eval node-keys (mapv #(nth % 1) pairs))))))))
+                    (batching-try-eval node-keys (mapv val expr-batch))))))))
 
 (sc/defn ^:private compile-node
   "Compiles a given node description into a node usable in the network with the
@@ -1849,6 +1850,37 @@
       productions
       (throw (ex-info (str "Non-unique production names: " non-unique) {:names non-unique})))))
 
+(def forms-per-eval-default
+  ;; The default max number of forms that will be evaluated together as a single batch.
+  ;; 5000 is chosen here due to the way that clojure will evaluate the vector of forms extracted from the nodes.
+  ;; The limiting factor here is the max java method size (64KiB), clojure will compile each form in the vector down into
+  ;; its own class file and generate another class file that will reference each of the other functions and wrap them in
+  ;; a vector inside a static method. For example,
+  ;;
+  ;; (eval [(fn one [_] ...) (fn two [_] ...)])
+  ;; would generate 3 classes.
+  ;;
+  ;; some_namespace$eval1234
+  ;; some_namespace$eval1234$one_1234
+  ;; some_namespace$eval1234$two_1235
+  ;;
+  ;; some_namespace$eval1234$one_1234 and some_namespace$eval1234$two_1235 contian the implementation of the functions,
+  ;; where some_namespace$eval1234 will contain two methods, invoke and invokeStatic.
+  ;; The invokeStatic method breaks down into something similar to a single create array call followed by 2 array set calls
+  ;; with new invocations on the 2 classes the method then returns a new vector created from the array.
+  ;;
+  ;; 5000 is lower than the absolute max to allow for modifications to how clojure compiles without needing to modify this.
+  ;; The current limit should be 5471, this is derived from the following opcode investigation:
+  ;;
+  ;; Array creation:                                                    5B
+  ;; Creating and populating the first 6 elements of the array:        60B
+  ;; Creating and populating the next 122 elements of the array:    1,342B
+  ;; Creating and populating the next 5343 elements of the array:  64,116B
+  ;; Creating the vector and the return statement:                      4B
+  ;;
+  ;; This sums to 65,527B just shy of the 65,536B method size limit.
+  5000)
+
 (sc/defn mk-session*
   "Compile the rules into a rete network and return the given session."
   [productions :- #{schema/Production}
@@ -1872,10 +1904,7 @@
         id-counter (atom 0)
         create-id-fn (fn [] (swap! id-counter inc))
 
-        ;; 1250 is an arbitrary number, this could be lower or higher depending on the
-        ;; rulebase that is being compiled, with regards to the average size of the
-        ;; forms being evaluated.
-        forms-per-eval (:forms-per-eval options 1250)
+        forms-per-eval (:forms-per-eval options forms-per-eval-default)
 
         beta-graph (to-beta-graph productions create-id-fn)
         alpha-graph (to-alpha-graph beta-graph create-id-fn)
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -75,7 +75,7 @@
                         ;; Function that takes facts and determines what alpha nodes they match.
                         get-alphas-fn
                         ;; A map of [node-id field-name] to function.
-                        node-identifier-to-fn])
+                        node-expr-fn-lookup])
 
 (defn- is-variable?
   "Returns true if the given expression is a variable (a symbol prefixed by ?)"
@@ -1256,7 +1256,7 @@
             empty-beta-graph
             productions))
 
-(sc/defn extract-exprs :- {(schema/tuple sc/Int sc/Keyword) schema/SExpr}
+(sc/defn extract-exprs :- schema/NodeExprLookup
   "Walks the Alpha and Beta graphs and extracts the expressions that will be used in the construction of the final network.
    The extracted expressions are stored by their key, [<node-id> <field-key>], this allows for the function to be retrieved
    after it has been compiled.
@@ -1389,10 +1389,10 @@
       exprs
       (:id-to-condition-node beta-graph))))
 
-(sc/defn compile-exprs :- {(schema/tuple sc/Int sc/Keyword) IFn}
+(sc/defn compile-exprs :- schema/NodeFnLookup
   "Takes a map in form produced by extract-exprs and evaluates the values(expressions) of the map in a batched manner.
    This allows the eval calls to be more effecient, rather than evaluating each expression on its own."
-  [key->expr :- {(schema/tuple sc/Int sc/Keyword) schema/SExpr}
+  [key->expr :- schema/NodeExprLookup
    partition-size :- sc/Int]
   (let [batching-try-eval (fn [node-keys exprs]
                             ;; Try to evaluate all of the expressions as a batch. If they fail the batch eval then we
@@ -1439,7 +1439,7 @@
    id :- sc/Int
    is-root :- sc/Bool
    children :- [sc/Any]
-   id-key->expr :- {(schema/tuple sc/Int sc/Keyword) IFn}
+   expr-fn-lookup :- schema/NodeFnLookup
    new-bindings :- #{sc/Keyword}]
 
   (let [{:keys [condition production query join-bindings]} beta-node
@@ -1448,7 +1448,7 @@
                     (.loadClass (clojure.lang.RT/makeClassLoader) (name condition))
                     condition)
 
-        retrieve-fn (fn [id field] (get id-key->expr [id field]))]
+        compiled-expr-fn (fn [id field] (get expr-fn-lookup [id field]))]
 
     (case (:node-type beta-node)
 
@@ -1467,7 +1467,7 @@
           (eng/->ExpressionJoinNode
             id
             condition
-            (retrieve-fn id :join-filter-expr)
+            (compiled-expr-fn id :join-filter-expr)
             children
             join-bindings)
           (eng/->HashJoinNode
@@ -1484,7 +1484,7 @@
         (eng/->NegationWithJoinFilterNode
           id
           condition
-          (retrieve-fn id :join-filter-expr)
+          (compiled-expr-fn id :join-filter-expr)
           children
           join-bindings)
         (eng/->NegationNode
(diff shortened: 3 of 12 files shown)


src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1697,11 +1796,9 @@
                              entry))
 
         ;; type, alpha node tuples.
-        alpha-nodes (for [{:keys [type alpha-fn children env] :as alpha-map} alpha-fns
+        alpha-nodes (for [{:keys [id type alpha-fn children env] :as alpha-map} alpha-fns
                           :let [beta-children (map id-to-node children)]]
-                      [type (with-meta
-                              (eng/->AlphaNode env beta-children alpha-fn)
-                              {:alpha-expr (:alpha-expr (meta alpha-map))})])
+                      [type (eng/->AlphaNode id env beta-children alpha-fn)])
```
EthanEChristian: I was torn here, i was temped to add the AlphaNodes to the `id-to-node` since alpha nodes have ids now.

src/test/clojure/clara/performance/test_compilation.clj
```diff
@@ -0,0 +1,154 @@
...
+                      (map #(- % mean))
+                      (map #(Math/pow (double %) 2.0)))
+                    execution-times)
+              sum
+              (/ iterations)
+              Math/sqrt)]
+    {:std (double std)
+     :mean (double mean)}))
+
+(defn run-performance-test
```
EthanEChristian: I rolled my own here, rather than using a library like `criterium` because criterium makes assumptions based on the classes loaded and compiled. This lead to infinite looping.

EthanEChristian:
Issue #381

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+                                                             :msg "compiling accumulate with join filter node"}}))
+            :query prev)))
+      exprs
+      (:id-to-condition-node beta-graph))))
+
+(sc/defn compile-exprs :- {(schema/tuple sc/Int sc/Keyword) IFn}
+  "Takes a map in form produced by extract-exprs and evaluates the values(expressions) of the map in a batched manner.
+   This allows the eval calls to be more effecient, rather than evaluating each expression on its own."
+  [key->expr :- {(schema/tuple sc/Int sc/Keyword) schema/SExpr}
+   partiton-size :- sc/Int]
```
EthanEChristian: minor spelling mistake, i clean that up

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -584,19 +587,32 @@
...
-      (let [{:keys [rulebase-only? base-rulebase]} opts
+      (let [{:keys [rulebase-only? base-rulebase compilation-partition-size]} opts
             
             record-holder (ArrayList.)
             ;; The rulebase should either be given from the base-session or found in
             ;; the restored session-state.
             maybe-base-rulebase (when (and (not rulebase-only?) base-rulebase)
                                   base-rulebase)
+
+            compilation-partition-size (or compilation-partition-size 1250)
```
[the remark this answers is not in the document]
EthanEChristian: Its pretty arbitrary, originally i had 1k but on testing i was able to load a larger kb with 1.25k. 
Honestly, it could probably be a bit higher, it would just come down to the size of the functions being compiled.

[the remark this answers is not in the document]
curtman16: might be good to say what 1250 is; bytes?  A comment as to its arbitrary nature would be good as well so the next one through feels better if it needs to be altered.

src/test/clojure/clara/performance/test_compilation.clj
```diff
@@ -0,0 +1,154 @@
...
+  {:ns-name (symbol (str *ns*))})
+
+(defn generate-filter-productions
+  [seed-syms]
+  (into {}
+        (for [seed-sym seed-syms
+              :let [next-fact (symbol (str seed-sym "prime"))
+                    production (assoc base-production
+                                 :lhs [{:type (keyword seed-sym)
+                                        :constraints ['(= this ?binding) `(filter-fn ~'this)]}]
```
mrrodriguez: Use syntax quote so that `=` is expanded. Even though Clara will accept the bare `=` symbol in many places, I think it is a bad practice due to it being able to be shadowed. Clara should be using `clojure.core/=`, not a bare `=`.
```clj
`(= ~'this ~'?binding)
```

This applies elsewhere in this ns too.

src/test/clojure/clara/performance/test_compilation.clj
```diff
@@ -0,0 +1,154 @@
+(ns clara.performance.test-compilation
+  (:require [clojure.test :refer :all]
+            [clara.rules.compiler :as com]
+            [clara.rules :as r]
+            [clara.rules.durability.fressian :as dura])
+  (:import [java.io ByteArrayOutputStream ByteArrayInputStream]))
+
+(defn filter-fn
```
mrrodriguez: I don't understand how this would work. `gensym` should return a new and unique symbols every time it is called. So this `filter-fn` will always be false.

If that is intentional, some sort of explanation doc string would be helpful.

src/test/clojure/clara/performance/test_compilation.clj
```diff
@@ -0,0 +1,154 @@
...
+          (for [combo (partition-all 2 (shuffle seed-syms))]
+            (apply template combo)))))
+
+(defn generate-collection-productions
+  [seed-syms]
+  (into {}
+        (for [seed-sym seed-syms
+              :let [next-fact (symbol (str seed-sym "prime"))
+                    production (assoc base-production
+                                 :lhs [{:accumulator '(clara.rules.accumulators/all)
```
mrrodriguez: `clara.rules.accumulators` needs to be required explicitly by the ns running these rules, which I believe is this one. There shouldn't be assumptions about other namespaces already requiring your direct dependencies. That can lead to compilation-order dependent runs that fail occasionally.

src/test/clojure/clara/performance/test_compilation.clj
```diff
@@ -0,0 +1,154 @@
...
+                                 :rhs `(r/insert! (with-meta ~'?binding
+                                                             {:type ~(keyword next-fact)})))]]
+          [next-fact production])))
+
+(defn generate-queries
+  [seed-syms]
+  (into {}
+        (for [seed-sym seed-syms
+              :let [production (assoc base-production
+                                 :lhs [{:type [seed-sym]
```
mrrodriguez: Why is the `:type` wrapped in a vector here? I think it should be `:type seed-sym`.

src/test/clojure/clara/performance/test_compilation.clj
```diff
@@ -0,0 +1,154 @@
...
+         :func #(com/mk-session rules)
+         :iterations 50
+         :mean-assertion (partial > 5000)}))
+
+    (let [session (com/mk-session rules)
+          os (ByteArrayOutputStream.)]
+      (testing "Session serialization performance"
+        (run-performance-test
+          {:description "Generated Session serialization"
+           :func #(.serialize (dura/create-session-serializer (ByteArrayOutputStream.))
```
mrrodriguez: You shouldn't interact with the API via the impl detail that protocols may expose interfaces.
This should be called via the protocol fn var.

```clj
[clara.rules.durability :as dur]
...

(dur/serialize ...)
```

Also, I'd recommend not calling `[clara.rules.durability.fressian :as dura]` since that is confusing.
Perhaps, `[clara.rules.durability.fressian :as fres]` or something

src/test/clojure/clara/performance/test_compilation.clj
```diff
@@ -0,0 +1,154 @@
...
+      (testing "Session deserialization performance"
+        (.serialize (dura/create-session-serializer os)
+                    session
+                    {:rulebase-only? true})
+
+        (let [session-bytes (.toByteArray os)]
+          (run-performance-test
+            {:description "Generated Session deserialization"
+             :func #(-> (dura/create-session-serializer (ByteArrayInputStream. session-bytes))
+                        (.deserialize session-bytes {:rulebase-only? true}))
```
mrrodriguez: Same as above, call the protocol fn via it's var, not impl detail Java interop.

mrrodriguez: Why is `:rulebase-only?` only used in the last deserialize case? Shouldn't it be set for all serialize and deserialize calls?

src/test/clojure/clara/performance/test_compilation.clj
```diff
@@ -0,0 +1,154 @@
...
+    (testing "Session creation performance"
+      (run-performance-test
+        {:description "Generated Session Compilation"
+         :func #(com/mk-session rules)
+         :iterations 50
+         :mean-assertion (partial > 5000)}))
+
+    (let [session (com/mk-session rules)
+          os (ByteArrayOutputStream.)]
+      (testing "Session serialization performance"
```
mrrodriguez: May be more appropriate: "Session rulebase durability performance". This isn't testing the memory serialization aspects.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1263,105 +1439,70 @@
...
    new-bindings :- #{sc/Keyword}]
 
   (let [{:keys [condition production query join-bindings]} beta-node
 
         condition (if (symbol? condition)
                     (.loadClass (clojure.lang.RT/makeClassLoader) (name condition))
-                    condition)]
+                    condition)
+
+        retrieve-fn (fn [id field] (get id-key->expr [id field]))]
```
mrrodriguez: Recommended rename: `compiled-expr-fn`. This is too generic and confusing to read later on.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1263,105 +1439,70 @@
    id :- sc/Int
    is-root :- sc/Bool
    children :- [sc/Any]
-   compile-expr ; Function to do the evaluation.
+   id-key->expr :- {(schema/tuple sc/Int sc/Keyword) IFn}
```
mrrodriguez: Similar as I mentioned above in this ns, perhaps calling it `expr-fn-lookup` is more meaningful. In this local scope, it is less important that where I mentioned above though.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1372,83 +1513,51 @@
...
-  (let [;; A local, memoized function that ensures that the same expression of
-        ;; a given node :id is only compiled into a single function.
-        ;; This prevents redundant compilation and avoids having a Rete node
-        ;; :id that has had its expressions compiled into different
-        ;; compiled functions.
-        compile-expr (memoize (fn [id expr] (try-eval expr)))
-
-        ;; Sort the ids to compile based on dependencies.
+  [{:keys [id-to-production-node id-to-condition-node id-to-new-bindings forward-edges backward-edges]} :- schema/BetaGraph
+   id-key->expr :- {(schema/tuple sc/Int sc/Keyword) IFn}]
```
mrrodriguez: Same renaming suggestions as above.

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: Maybe. I'm not sure what utility there would be in it. It may be useful at some point.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1677,7 +1775,8 @@
    fact-type-fn
    ancestors-fn
    activation-group-sort-fn
-   activation-group-fn]
+   activation-group-fn
+   expressions]
```
mrrodriguez: Another place I'd consider renaming as mentioned above. `expressions` is probably the most vague.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1537,37 +1647,25 @@
...
-               :alpha-fn (binding [*file* (or (:file cmeta) *file*)
-                                   *compile-ctx* {:condition condition
-                                                  :env env
-                                                  :msg "compiling alpha node"}]
-                           (try-eval alpha-expr))
-               :children beta-children}
-        env (assoc :env env))
-      {:alpha-expr alpha-expr})))
+  [alpha-nodes :- [schema/AlphaNode]
+   id-key->expr :- {(schema/tuple sc/Int sc/Keyword) IFn}]
```
mrrodriguez: Same renaming suggestions as above.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+   create-id-fn :- IFn]
+  (reduce (fn [beta-graph production]
               (binding [*compile-ctx* {:production production}]
                 (add-production production beta-graph create-id-fn)))
 
             empty-beta-graph
-            productions)))
+            productions))
+
+(sc/defn extract-exprs :- {(schema/tuple sc/Int sc/Keyword) schema/SExpr}
```
mrrodriguez: This schema is referred to a lot. It'd be helpful to extract it with a `s/defschema` so it is easier to read through the many places it is used.

Suggested `NodeExprFnLookup`

src/test/clojure/clara/performance/test_compilation.clj
EthanEChristian: As these rules never fire, this is a placeholder.

src/test/clojure/clara/performance/test_compilation.clj
EthanEChristian: actually, `(keyword seed-sym)`

EthanEChristian:
I forgot to update the changelog, so i will be adding that shortly as well

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+                           (:id-to-production-node beta-graph))]
+    (reduce-kv
+      (fn [prev id beta-node]
+        (let [condition (:condition beta-node)
+              condition (if (symbol? condition)
+                          (.loadClass (clojure.lang.RT/makeClassLoader) (name condition))
+                          condition)]
+          (case (or (:node-type beta-node)
+                    ;; If there is no :node-type then the node is the ::root-condition
+                    ;; however, creating a case for nil could potentially casue weird effects if something about the
```
mrrodriguez: typo: "casue" -> "cause"

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+                                              {:compile-ctx {:condition condition
+                                                             :join-filter-expressions (:join-filter-expressions beta-node)
+                                                             :env (:env beta-node)
+                                                             :msg "compiling accumulate with join filter node"}}))
+            :query prev)))
+      exprs
+      (:id-to-condition-node beta-graph))))
+
+(sc/defn compile-exprs :- schema/NodeFnLookup
+  "Takes a map in form produced by extract-exprs and evaluates the values(expressions) of the map in a batched manner.
```
mrrodriguez: grammar:  "in form produced" -> "in the form produced"

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+   This allows the eval calls to be more effecient, rather than evaluating each expression on its own."
```
mrrodriguez: Maybe a line after this should say 
"See https://github.com/cerner/clara-rules/issues/381 for more details."

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+                                ;; size of the code trying to be evaluated has exceeded the limit
+                                ;; set by java.
+                                (throw (ex-info "The batching size of expressions has exceeded the limit set by java."
+                                                {:orginal-exception e})))))]
+    (into {}
+          cat
+          ;; Grouping by ns, most expressions will not have a defined ns, only expressions from production nodes.
+          ;; These expressions must be evaluated in the ns where they were defined, as they will likely contain code
+          ;; that is namespace dependent.
+          (for [[ns pairs] (group-by (comp :ns meta key) key->expr)
```
mrrodriguez: Suggestion: Maybe avoid naming confusion here, call this `[ns group]`
and then below bind the `partition-all` result to `expr-batch`.

It was a bit confusing to read `pairs` so much and also not that informative of what it is.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+                ;; Partitioning the number of forms to be evaluated, Java has a limit to the size of methods if we were
+                ;; evaluate all expressions at once it would likely exceed this limit and throw an exception.
+                pairs (partition-all partition-size pairs)
+                :let [node-keys (map #(nth % 0) pairs)]]
```
mrrodriguez: Are you avoiding using `first` here for some reason? If it is perf, I believe that `pairs` in this line is a seq of map entries. If that is the case, `key` is the fastest.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+                                            (with-meta (compile-action (:bindings production-node)
+                                                                       (:rhs production)
+                                                                       (:env production))
+                                                       (meta (:rhs production)))
+                                            id
+                                            :action-expr
+                                            ;; ProductionNode expressions can be arbitrary code, therefore we need the
+                                            ;; ns where the production was define so that we can compile the expression
+                                            ;; later.
+                                            {:ns (some-> (:ns-name production) str)
```
mrrodriguez: Why is the `:ns` stored as a string here? Also, I think it more typically to do 
`(some-> (:ns-name production) name)` if a string is desired from a ns sym.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+          cat
+          ;; Grouping by ns, most expressions will not have a defined ns, only expressions from production nodes.
+          ;; These expressions must be evaluated in the ns where they were defined, as they will likely contain code
+          ;; that is namespace dependent.
+          (for [[ns pairs] (group-by (comp :ns meta key) key->expr)
+                ;; Partitioning the number of forms to be evaluated, Java has a limit to the size of methods if we were
+                ;; evaluate all expressions at once it would likely exceed this limit and throw an exception.
+                pairs (partition-all partition-size pairs)
+                :let [node-keys (map #(nth % 0) pairs)]]
+            (mapv (fn [node-key expr]
```
mrrodriguez: Note: this is `mapv vector`, the anonymous fn just reimplements it's 2-arity case.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+                    [node-key expr])
+                  node-keys
+                  (with-bindings (if ns
+                                   {#'*ns* (-> ns symbol the-ns)}
+                                   {})
+                    (batching-try-eval node-keys (mapv #(nth % 1) pairs))))))))
```
mrrodriguez: Similar to above, `second`, or better, `val` if it is always a map entry.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,191 @@
...
+                                          (with-bindings
+                                            {#'*compile-ctx* (:compile-ctx cmeta)
+                                             #'*file* (:file cmeta *file*)}
+                                            (try-eval expr))))
+                                      exprs
+                                      node-keys)
+                                ;; If none of the rules are the issue, it is likely that the
+                                ;; size of the code trying to be evaluated has exceeded the limit
+                                ;; set by java.
+                                (throw (ex-info "The batching size of expressions has exceeded the limit set by java."
```
mrrodriguez: `ex-info` has an 3-arity version designed to take the `cause`, ie `ex-info [msg map cause]`

So 
```clj
(ex-info "The batching size of expressions has exceeded the limit set by java." {} e)

```

I'm not sure the `:original-exception` will add value. Perhaps if that exception has `ex-data` to include. Also, it may be useful to put the `node-keys` in this map.

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: The `group-by` will result in a non-deterministic order here. It may be best (for debugging perhaps), to be deterministic.
Perhaps a `sort-by key` would be reasonable to add?

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1769,11 +1867,26 @@
...
-        beta-tree (compile-beta-graph beta-graph)
+
+        ;; A stateful counter used for unique ids of the nodes of the graph.
+        id-counter (atom 0)
+        create-id-fn (fn [] (swap! id-counter inc))
+
+        ;; 1250 is an arbitrary number, this could be lower or higher depending on the
+        ;; rulebase that is being compiled, with regards to the average size of the
+        ;; forms being evaluated.
+        forms-per-eval (:forms-per-eval options 1250)
```
mrrodriguez: It'd be nice to have the 1250 in a var 
`forms-per-eval-default` somewhere and refer to that in all the places that you currently refer to 1250. That way the comments and the compiler vs deserialization impl doesn't get out of sync.

Also, how did you come up with 1250? Is this a threshold you found to work well in practice? It does depend on size of expressions of course.

As an aside to this discussion:
I think ideally, rules would be written in a way where there were not large inline code blocks in the LHS conditions, accumulators, and/or the RHS action forms.

I know from experience that people sometimes tend to make really large RHS forms though (I don't recommend it).

e.g.

Bad
```clj

(defrule my-rule
 
=>
(let [<lots of bindings and logic>]
   (insert! some-result))
```

Good
Good
```clj

(defrule my-rule
 
=>
;; `do-bigger-logic-fn` is evaluated under normal compilation conditions (better).
;; Also, things can be broken up and made more readable by naming functions around them.
;; Also, these pure fn's can be independently verified/tested.
(let [some-result (do-bigger-logic-fn <bindings needed from LHS>)]
   (insert! some-result))
```

If rules were written this way, I'd think the `:forms-per-eval` could be quite a bit higher. I realize that'd be a breaking change for impl's that have abused the RHS by making huge action expressions.

mrrodriguez: @EthanEChristian and I talked a lot about this offline. It actually may be the case that inline forms don't have an effect on the `:forms-per-eval`. This is nice if it proves to be the case.

Basically, the reason is that `batching-try-eval` eval's a vector of expressions of a `fn` form.
e.g.
```clj
[(fn [x] ...) (fn [y] ...) ...]
```

These each get their own classfile, so the big method size ends up just being in the method that creates this vector of `fn` references. There is constant overhead in terms of bytecodes for this method then. That constant overhead can be multiplied by teh `:forms-per-eval` to get a close approx of the correct value to default to. Keeping it configurable is likely still a good idea, due these sizes being possibly variable across Clj compiler versions, etc.

We can make sure to add those details here shortly once it is completely validated.

WilliamParker:
@EthanEChristian @mrrodriguez I'll try to take an initial look at this over the weekend.

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: Hold over from a previous pass, where this metadata was on the fn that would have been eval'd. Also a reason that 1250 was chosen in the other locations.

The addition of the metadata push the limit well below what we discussed on this comment 
https://github.com/cerner/clara-rules/pull/396#discussion_r197290343

src/main/clojure/clara/rules/durability/fressian.clj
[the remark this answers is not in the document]
EthanEChristian: No longer accurate, this has changed

src/main/clojure/clara/macros.clj
```diff
@@ -196,10 +196,11 @@
...
   (vec
-   (for [{:keys [condition beta-children env]} alpha-nodes
+   (for [{:keys [id condition beta-children env]} alpha-nodes
          :let [{:keys [type constraints fact-binding args]} condition]]
 
-     {:type (com/effective-type type)
+     {:id id
+      :type (com/effective-type type)
       :alpha-fn (com/compile-condition type (first args) constraints fact-binding env)
       :children (vec beta-children)})))
```
WilliamParker: It seems odd to me that we build a map here and then use it to construct alpha nodes later in assemble-session while building the beta nodes directly in this namespace, but changing that is probably out of scope here.  Maybe there's a reason for this that I'm not remembering offhand.

src/main/clojure/clara/rules/schema.cljc
```diff
@@ -1,7 +1,10 @@
 (ns clara.rules.schema
   "Schema definition of Clara data structures using Prismatic's Schema library. This includes structures for rules and queries, as well as the schema
    for the underlying Rete network itself. This can be used by tools or other libraries working with rules."
-  (:require [schema.core :as s]))
+  (:require [schema.core :as s])
+  #?(:clj
+     (:import [clojure.lang
+               IFn])))
```
WilliamParker: Minor: You could use the [ifn? predicate](https://clojuredocs.org/clojure.core/ifn_q) instead if you want to avoid the need for conditionals.  Looks like that has been in ClojureScript since 2012: [commit](https://github.com/clojure/clojurescript/commit/88a82720bd0642ace4d751343372c3106c8556ea)

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,192 @@
...
+        alpha-exprs (reduce (fn [prev alpha-node]
+                              (let [{:keys [id condition env]} alpha-node
+                                    {:keys [type constraints fact-binding args]} condition
+                                    cmeta (meta condition)]
+                                (handle-expr prev
+                                             (with-meta (compile-condition
+                                                          type (first args) constraints
+                                                          fact-binding env)
+                                                        ;; Remove all metadata but file and line number
+                                                        ;; to protect from evaluating usafe metadata
```
WilliamParker: spelling: "usafe" => "unsafe"

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,192 @@
...
+                                                        (select-keys cmeta [:line :file]))
+                                             id
+                                             :alpha-expr
+                                             {:file (or (:file cmeta) *file*)
+                                              :compile-ctx {:condition condition
+                                                            :env env
+                                                            :msg "compiling alpha node"}})))
+                            {}
+                            alpha-graph)
+        exprs (reduce-kv (fn [prev id production-node]
```
WilliamParker: IMO a more specific name here than "exprs" would be helpful.  This will be a map containing expressions if I'm understanding correctly?

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -31,7 +31,9 @@
             Accumulator
             ISystemFact]
```
WilliamParker: I'll have to think about if there are any ways this batching could interact negatively (i.e. in a bug-causing way) with node sharing.  We might need to add some tests around that.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,192 @@
...
+                                             :file (-> production :rhs meta :file)
+                                             :compile-ctx {:production production
+                                                           :msg "compiling production node"}})))
+                           alpha-exprs
+                           (:id-to-production-node beta-graph))]
+    (reduce-kv
+      (fn [prev id beta-node]
+        (let [condition (:condition beta-node)
+              condition (if (symbol? condition)
+                          (.loadClass (clojure.lang.RT/makeClassLoader) (name condition))
```
WilliamParker: Assuming that anything that is a symbol is a class seems likely to be a bug - we'll need to test it though.  Looks like there is some [existing logic along these lines](https://github.com/cerner/clara-rules/blob/0.18.0/src/main/clojure/clara/rules/compiler.clj#L181).  Also will we enter this path from ClojureScript at all?

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,192 @@
...
+                                          (with-bindings
+                                            {#'*compile-ctx* (:compile-ctx cmeta)
+                                             #'*file* (:file cmeta *file*)}
+                                            (try-eval expr))))
+                                      exprs
+                                      node-keys)
+                                ;; If none of the rules are the issue, it is likely that the
+                                ;; size of the code trying to be evaluated has exceeded the limit
+                                ;; set by java.
+                                (throw (ex-info "The batching size of expressions has exceeded the limit set by java."
```
WilliamParker: It is OK to say the likely cause in an exception message, but IMO we shouldn't state it with quite so much certainty.  For example, we could have a bug in our implementation that causes such an exception.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,192 @@
...
+  [key->expr :- schema/NodeExprLookup
+   partition-size :- sc/Int]
+  (let [batching-try-eval (fn [node-keys exprs]
+                            ;; Try to evaluate all of the expressions as a batch. If they fail the batch eval then we
+                            ;; try them one by one with their compilation context, this is slow but we were going to fail
+                            ;; anyway.
+                            (try
+                              (eval exprs)
+                              (catch Exception e
+                                (mapv (fn [expr node-keys]
```
WilliamParker: "doseq" seems more in line with your intent here since this return value isn't used.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1246,192 @@
...
+   See #381 for more details."
+  [key->expr :- schema/NodeExprLookup
+   partition-size :- sc/Int]
+  (let [batching-try-eval (fn [node-keys exprs]
+                            ;; Try to evaluate all of the expressions as a batch. If they fail the batch eval then we
+                            ;; try them one by one with their compilation context, this is slow but we were going to fail
+                            ;; anyway.
+                            (try
+                              (eval exprs)
+                              (catch Exception e
```
WilliamParker: I wonder if there are scenarios where the compilation could throw an Error other than a VirtualMachineError (which it wouldn't make sense to try to recover from, see the [docs](https://docs.oracle.com/javase/7/docs/api/java/lang/VirtualMachineError.html))

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Seems like it could be useful for debugging and REPL exploration.

EthanEChristian: Since there seems to be value, i will add them

src/main/clojure/clara/macros.clj
EthanEChristian: I agree that is a little weird the only thing i think that is forcing this is, the access to the compiled beta nodes. At this point the beta-children are just ids of the nodes that will be constructed later.

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: This is previous functionality, just shuffled around to work in my new approach. It does seem a little questionable, but i assumed that maintaining the old way would be better than changing functionality.

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: Used `mapv` to keep from having to zip the 2 lists back together, if you feel strongly i can use `doseq` instead

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: `catch Exception` shouldn't catch these sorts of errors anyway right? 

But you do make a good point, [InterruptedException](https://docs.oracle.com/javase/7/docs/api/java/lang/InterruptedException.html) would be a good candidate to not catch

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -408,31 +408,11 @@
...
-              (reify ReadHandler
-                (read [_ rdr tag component-count]
-                  (d/clj-record-idx->fact (.readObject rdr))))
-              "clara/alphanode"
-              (reify ReadHandler
-                (read [_ rdr tag component-count]
-                  (-> rdr
-                      (read-record d/add-alpha-fn)
-                      d/clj-record-holder-add-fact!)))}}
+   (create-cached-node-handler AlphaNode
```
EthanEChristian: Now that AlphaNodes have ids we should be able to leverage the cached-node-handlers

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: TODO: determine what exceptions to not catch

mrrodriguez: I don't see a reason why you shouldn't catch all exceptions and rethrow here. I think that the message needs to be clear that it could only "possibly" be a problem with compilation of the expression. However, I don't think it is beneficial to rethrow other exceptions in their "raw" form . It is more useful to wrap the exception and provide it with what context you can. Attach the root cause exception though so it is bubbles up.

More context should always be a better thing in this sort of case.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Yeah, it should be fine to leave as is with regards to this issue.  I have a suspicion there is a way to break this (as in make it throw ClassNotFoundException) but I don't think this change would make it any more broken since the assumption is the same before and after.  We can look at this separately.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Oh, I see, the anonymous function has two args.  Never mind, this is OK and it is a minor deal anyway.  It might be worth adding a quick comment that we want mapv rather than map to avoid laziness.

Do we have an existing test that validates this "rethrow if an individual expression actually is not compilable" workflow?  If not we should probably add one.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: I was actually musing on whether there are any Error types that we should catch, with the idea being to rethrow them with the specific expression that caused them rather than a gigantic collections that probably isn't too useful to the user.  I don't know that the traditional assumptions about not recovering from Error necessarily apply to a dynamic code generation and compilation framework - I particularly had [LinkageError](https://docs.oracle.com/javase/7/docs/api/java/lang/LinkageError.html) in mind.  We definitely wouldn't want to catch any VirtualMachineError instances though since the JVM's behavior when you do is specifically called out as undefined by the spec.

In practice this probably isn't a huge deal and I don't want to block this on investigating it - I'd expect most bad cases to throw Exception.  Maybe just a followup issue to think about later.

src/test/clojure/clara/performance/test_compilation.clj
WilliamParker: "because criterium makes assumptions based on the classes loaded and compiled" => Could you elaborate?  What sorts of assumptions?

EthanEChristian: @WilliamParker as @mrrodriguez mentioned `criterium` does try to warm up for JIT, [here](https://github.com/michalmarczyk/criterium/blob/0.4.0/src/criterium/core.clj#L396-L435). Again as Mike says, it seems that `criterium` is looking for a stable state before running benchmarks. While ideal for most situations, compilation in this case will never reach a steady state. Effectively making the test a really expensive `while (true)` block.

I will probably put a comment on this function mentioning the reasoning.

src/test/clojure/clara/test_rules.clj
```diff
@@ -2748,3 +2752,26 @@
                     (fire-rules))]
 
     (is (= [{:?x inner-one}] (query session test-query)))))
+
+(deftest test-maximum-forms-per-eval
```
EthanEChristian: @WilliamParker, A test demonstrating the batch eval failure with a method size too large.

This test is a little slow, i was considering putting it in the performance tests, but i would be worried that those test would not be run and we might miss changes to compilation that might break the limit.

WilliamParker: How long does this take to run?  Perhaps it should be in another test file - it would still run during Leiningen builds but not as part of the test-rules namespace.  I've been trying to move away from having everything in test-rules from a code organization POV anyway.

WilliamParker:
Looked over the compiler changes and left some comments, but overall looks good @EthanEChristian 

I haven't been through the durability changes yet.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1248,211 @@
...
+   if available, the compile context, file and ns. This metadata is not stored on the expression itself because it will
+   contain forms that will fail to eval."
+  [beta-graph :- schema/BetaGraph
+   alpha-graph :- [schema/AlphaNode]]
+  (let [backward-edges (:backward-edges beta-graph)
+        handle-expr (fn [id->expr s-expr id expr-key metadata]
+                      (assoc id->expr
+                        ;; Letting the key carry all of the metadata, this makes the batch evaluation of the expressions
+                        ;; easier because the metadata contains unevalable forms.
+                        (with-meta [id expr-key] (assoc metadata expr-key s-expr))
```
WilliamParker: This pattern of having the metadata carry fundamental semantic meaning of the data (without which durability would be broken) that isn't reflected in the NodeExprLookup schema seems like a suboptimal practice.  I'd like to at least have the metadata expectations reflected in the schema.  Alternatively you could have the key map to a tuple, map, etc. containing both the values, but my primary concern here is that as someone reading the code, I'd expect the schema to be a complete description of the data, or at least the mandatory parts of the data.  This seems to break that expectation.

```
because the metadata contains unevalable forms.
```
Are you referring to the issues discussed in https://github.com/cerner/clara-rules/pull/243 ?  If so could you reference it in the comment?

src/test/clojure/clara/test_rules.clj
WilliamParker: Yikes, that is quite a long-running test.  Perhaps we could put it in a "long_running_tests" file that wouldn't be part of a normal REPL workflow; if there are other instances of things like this it might be worthwhile to create a test profile that is run in Travis but exclude the file from "lein test" runs.

It also seems like this could assert on the exception message added about message size too large?  Or by the exception info, you could add a key if needed.  It would be pretty simple to just recursively get all the messages and make sure one of them matches a regex.

project.clj
```diff
@@ -68,8 +68,15 @@
...
-                   :generative (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.generative.*")))}
+                                          (let [blacklisted-packages #{"generative" "performance"}
+                                                patterns (into []
+                                                           (comp
+                                                             (map #(str "^clara\\." % ".*"))
+                                                             (interpose "|"))
+                                                           blacklisted-packages)]
+                                            (some->> x :ns ns-name str (re-matches (re-pattern (apply str patterns)))))))
+                   :generative (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.generative.*")))
+                   :performance (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.performance.*")))}
```
WilliamParker: Having this live in the main Clara project seems reasonable to me, but there is an existing (old) [clara-benchmark](https://github.com/rbrush/clara-benchmark) project.  FYI @rbrush in case you have suggestions, objections, etc.

src/main/clojure/clara/rules/compiler.clj
WilliamParker: I don't see any problem here on review.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1248,211 @@
...
+                                   {#'*ns* (the-ns nspace)}
+                                   {})
+                    (batching-try-eval node-keys (mapv val expr-batch))))))))
+
+(defn safe-get
+  "A helper function for retrieving a given key from the provided map. If the key doesn't exist within the map this
+   function will throw an exception."
+  [m k]
+  (let [v (get m k ::not-found)]
+    (if (identical? v ::not-found)
```
WilliamParker: I presume the reason you used identical? instead of = here was for performance?  @EthanEChristian : I realize that as we've discussed previously in another (non-Clara) context it is currently part of Clojure's implementation that equal keywords are identical, and so when comparing two things that are always keywords using identity checks is a way to gain marginally better performance.  However, the number of operations here is just the number of nodes in the rules session compiled, and then only at compile-time, correct?  I doubt that that will be any kind of serious problem for a number of checks (at compile time) on the order of thousands or tens of thousands, and this pattern isn't a part of Clara's codebase yet.  As is, this introduces a maintenance burden of keeping track of whether Clojure has changed this; I don't believe that property is part of any API guarantees.

In this particular case, I think we can resolve this one of two ways
- Using clojure.core/=.  I don't think this is likely to have any meaningful performance cost, but if you have data to the contrary please post.
- Creating a let-binding for the keyword so that it actually is guaranteed to be identical by Clojure's public contracts and commenting with the reasoning.  

i.e.

```
(defn safe-get
  "A helper function for retrieving a given key from the provided map. If the key doesn't exist within the map this
   function will throw an exception."
  [m k]
  (let [not-found-keyword ::not-found
        v (get m k not-found-keyword)]
    (if (identical? v not-found-keyword)
      (throw (ex-info "Key not found with safe-get" {:map map :key k}))
      v)))
```

project.clj
rbrush: I'm in favor of including this in the main Clara project. The test selectors are a nice solution to performance concerns here.

EthanEChristian:
@WilliamParker, i can see what you mean with the metadata.
I have changed the implementation to not rely on metadata, instead the lookups now us a tuple as the value, making the lookup something like:

```
{[Int Keyword] [SExpr/Fn CompilationContext]}
```

where `CompilationContext` is what the metadata used to be.

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -577,10 +580,13 @@
...
-            reconstruct-expressions (fn [ks]
+            reconstruct-expressions (fn [expr-lookup]
+                                      ;; Rebuilding the expr-lookup map from the serialized map:
+                                      ;; {[Int Keyword] {Keyword Any}} -> {[Int Keyword] [SExpr {Keyword Any}]}
                                       (into {}
-                                            (for [k ks]
-                                              [k (-> k meta (get (nth k 1)))])))
+                                            (for [[node-key compilation-ctx] expr-lookup]
+                                              [node-key [(-> compilation-ctx (get (nth node-key 1)))
+                                                         compilation-ctx]])))
```
WilliamParker: So the line and file info will be included with the serialized rulebase correct?  That will marginally increase the space consumed by the rulebase, but the size of sessions won't be impacted so I think this is OK.

EthanEChristian: The rest of the `:compile-ctx` will be there too, such as `:condition` and `:production`. Previously this information was discarded after compiling the rulebase the first time. We then were making the assumption that it would recompile after deserialization. This way if for some reason a `fn`  cannot be recompiled, it should give better context to what condition/production the function was being compiled for.

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: I think `resolve` would be more appropriate, however, `resolve` has a defect that can throw CNFE as well.
https://dev.clojure.org/jira/browse/CLJ-1403

this may be fixed in 1.10, but that is future thinking.

Even if `resolve` returns `nil` as expected here, handling the nil/exception case would be nicer than letting it bubble up without much context.

I agree though that this is not related to the work of this PR.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1248,196 @@
...
+        (let [condition (:condition beta-node)
+              condition (if (symbol? condition)
+                          (.loadClass (clojure.lang.RT/makeClassLoader) (name condition))
+                          condition)]
+          (case (or (:node-type beta-node)
+                    ;; If there is no :node-type then the node is the ::root-condition
+                    ;; however, creating a case for nil could potentially cause weird effects if something about the
+                    ;; compilation of the beta graph changes. Therefore making an explicit case for ::root-condition
+                    ;; and if there was anything that wasn't ::root-condition this case statement will fail rather than
+                    ;; failing somewhere else.
```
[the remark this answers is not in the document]
mrrodriguez: I like failing instead of waiting to later and being broken. However, I think a it'd be nicer to have that default `case` at the end that threw a more meaningful exception than "no branch of `case` matched".

eg
```clj
(case x
 y :foo
 z :bar
 (throw (ex-info "explanation"))
```

However, this isn't a critical change for the sake of this review. It's more of a thought about trying to give more meaningful compilation failure errors in the future.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1242,19 +1248,210 @@
...
+                                   {#'*ns* (the-ns nspace)}
+                                   {})
+                    (batching-try-eval compilation-ctxs exprs)))))))
+
+(defn safe-get
+  "A helper function for retrieving a given key from the provided map. If the key doesn't exist within the map this
+   function will throw an exception."
+  [m k]
+  (let [v (get m k ::not-found)]
+    (if (= v ::not-found)
```
mrrodriguez: To @WilliamParker 's point:

> The potentially unreliable use of identical? in the safe-get function.

The likelihood of this changing in clj is very slim. It is even slimmer that it'd change for keyword literals. However, if choosing to remain paranoid for a change like that to happen, I still think an `identical?`-based sentinel test is what would typically be done in this sort of case

```clj
(let [not-found ::not-found
       v (get m k not-found)]
  (if (identical? v not-found) <etc>))
```

Alternatively, I've even seen the more aggressive, guaranteed-unique sentinel used:
```clj
(let [not-found (Object.)      ;; <---- unique object identity
       v (get m k not-found)]
  (if (identical? v not-found) <etc>))
```
Using a keyword feels more idiomatic, but using the Object is a stronger guarantee, but feels more like idiomatic Java. 

I am not against the keyword style, but in a generic fn like `safe-get`, I'd lean towards using an approach where `identical?` is used instead of `=`.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1751,6 +1874,37 @@
       productions
       (throw (ex-info (str "Non-unique production names: " non-unique) {:names non-unique})))))
 
+(def forms-per-eval-default
+  ;; The default max number of forms that will be evaluated together as a single batch.
```
mrrodriguez: I like the analysis. The only thing that is odd to me is you used inline comments instead of making this a doc string of the var. I don't suppose it matters much though.

EthanEChristian:
@WilliamParker  @mrrodriguez ,
I think everything has been addressed. 🎉

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Yeah, I agree that the likelihood of it changing is low, but not entirely trivial.  I'd rather not take the risk when there is an easy way not to though.  I'm OK with using the same reference to enforce identity as you describe.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1875,34 +1881,34 @@
...
+   5000 is lower than the absolute max to allow for modifications to how clojure compiles without needing to modify this.
+   The current limit should be 5471, this is derived from the following opcode investigation:
+
+   Array creation:                                                    5B
+   Creating and populating the first 6 elements of the array:        60B
+   Creating and populating the next 122 elements of the array:    1,342B
+   Creating and populating the next 5343 elements of the array:  64,116B
+   Creating the vector and the return statement:                      4B
+
+   This sums to 65,527B just shy of the 65,536B method size limit."
```
WilliamParker: Not a blocker, but this increases the importance of setting up our tests against all relevant Clojure versions in Travis since potentially different/future Clojure versions could have differences here.  That is really something we should be doing anyway though.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1448,9 +1453,10 @@
   "A helper function for retrieving a given key from the provided map. If the key doesn't exist within the map this
    function will throw an exception."
   [m k]
-  (let [v (get m k ::not-found)]
-    (if (= v ::not-found)
-      (throw (ex-info "Key not found with safe-get" {:map map :key k}))
+  (let [not-found ::not-found
+        v (get m k not-found)]
+    (if (identical? v not-found)
+      (throw (ex-info "Key not found with safe-get" {:map m :key k}))
```
WilliamParker: +1 to this approach to enforcing identity equality

Issue #393, opened 2018-05-21
clara.rules.platform/group-by-seq used in engine may cause incorrect semantics


mrrodriguez:
`clara.rules.platform/group-by-seq` can cause a subtle defect when Clojure record are involved. More generally, the problem is that `c.r.p/group-by-seq` uses a `java.util.LinkedHashMap` to perform the grouping operation. This is a Java-based collection type and is only aware of Java based equality semantics, ie `Object.equals()` and `Object.hashCode()`.

Clojure has it's own equality semantics taken into account in `=` and `hash`. There are some occasions where the difference between the Clojure vs Java equality semantics can cause functionally different behavior.

`c.r.p/group-by-seq` is used internally within the Clara engine for various performance-sensitive operations. The `j.u.LinkedHashMap` was chosen to be used within `c.r.p/group-by-seq` to streamline this performance-sensitive operation. a `j.u.LinkedHashMap` is convenient since it is mutable, for fast building of the ephemeral data structure needed. It also provides insert-order ordering determinism. This is useful to guarantee consistent execution times through the Clara engine from one JVM instance to the next, given the same input.

A concrete example of a problem with `j.u.LinkedHashMap`'s use of Java-based equality semantics has came up with regard to Clojure record types. 
Clojure record types use their type information in `clojure.core/=` comparisons and `clojure.core/hash`. However, they also are meant to interop with Java, that is unaware of their record types, and instead they pose as plain `java.util.Map` impl's from the Java interop perspective. This means that Clojure records `Object.equals()` and `Object.hashCode()` implementations do not take the type of records into account.

e.g.
```clj
(defrecord Foo [])
(defrecord Bar[])

(= (->Foo) (->Bar)) ;;= false
(.equals (->Foo) (->Bar)) ;;= true
```

This is a problem in the Clara engine. One example is that the engine it uses `c.r.p/group-by-seq` on `:bindings` of a rule. The `:bindings` may contained Clojure records. The Clojure records should be compared with Clojure equality semantics, but instead they are using Java semantics.

@thegeez in the #clara Slack channel gave a reproducing case that can show this defect manifest itself. I'll copy it here:
```clj
(defrecord ClauseOne [b])
(defrecord ClauseTwo [b])

(defrule rule-one
  [ClauseOne (= ?b b)]
  [ClauseTwo (= ?b b)]
 =>
  (println "Match found:" ?b )
  )

(defrecord WrapAlice [num])
(defrecord WrapBob [num])

(comment
[...]

PR #397, opened 2018-06-22
ISSUE-393: Fix incorrect equality semantics when unifying similar obj…

rbrush:
Fix for #393 by simply wrapping potential keys in a Java object that explicitly uses Clojure's equality and hash logic.

### DIFF
--- a/src/main/clojure/clara/rules/platform.cljc
+++ b/src/main/clojure/clara/rules/platform.cljc
@@ -17,7 +17,8 @@
 
      Object
      (equals [this other]
-       (= wrapped (.wrapped ^JavaEqualityWrapper other)))
+       (and (instance? JavaEqualityWrapper other)
+            (= wrapped (.wrapped ^JavaEqualityWrapper other))))
 
      (hashCode [this]
        hash-code)))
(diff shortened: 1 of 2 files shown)


src/main/clojure/clara/rules/platform.cljc
```diff
@@ -7,10 +7,25 @@
...
+;; semantics are visible to Java code. This allows these Clojure objects
+;; to be safely used in things like Java Sets or Maps.
+;; This class also accepts and stores the hash code, since it almost always
+;; will be used once and generally more than once.
+#?(:clj
+   (deftype JavaEqualityWrapper [wrapped ^int hash-code]
+
+     Object
+     (equals [this other]
+       (= wrapped (.wrapped ^JavaEqualityWrapper other)))
```
mrrodriguez: You are assuming the `other` is always going to be a `JavaEqualityWrapper`, otherwise there could be cast exceptions. I believe that is fine though, since this is only intended to be used/compared in collections that contain only these `JavaEqualityWrapper` wrapper types.

To confirm: this is your reasoning here, correct?

rbrush: That is the only use case I had in mind, but it's a good idea to add an instance? check here. They're quite cheap and makes this class more general.

rbrush:
Merged!

PR #399, opened 2018-06-28
ISSUE-393: Use Java's hashCode to improve performance in this case

rbrush:
An addendum to #393 to improve performance by using Java's hashcode function.


src/main/clojure/clara/rules/platform.cljc
```diff
@@ -37,7 +37,12 @@
      [f coll]
      (let [^java.util.Map m (reduce (fn [^java.util.Map m x]
                                       (let [k (f x)
-                                            wrapper (JavaEqualityWrapper. k (hash k))
+                                            ;; Use Java's hashcode for performance reasons as
+                                            ;; discussed at https://github.com/cerner/clara-rules/issues/393
+                                            wrapper (JavaEqualityWrapper. k
+                                                                          (if (nil? k)
+                                                                              0
```
WilliamParker: Does this get passed in properly as an integer type?  IIRC Clojure will coerce numeric types in cases like this.  We probably have a test somewhere with a nil binding validating this too.

rbrush: This wasn't obvious to me either, so I tried a 100% Java implementation of this operation that guaranteed the use of primitive integers, and there was no measurable performance difference. So either Clojure is smart enough to be use the primitives here or it's fast enough that it doesn't make a difference.

Good idea on the explicit nil test. I went ahead and added that.

mrrodriguez: I'll demonstrate with some bytecode gen about what clj is capable of inferring. It is the case that clj generates bytecode that won't mistakenly pass a long type to `deftype` field declared with an int type.
However, as far as perf is concerned, there is some auto-boxing/unboxing in play (without a few tweaks here).

I'll repeat our wrapper signature here, just for clarity:
```clj

(deftype JavaEqualityWrapper [wrapped ^int hash-code]
 
  Object
  (equals [this other]
    (and (instance? JavaEqualityWrapper other)
         (= wrapped (.wrapped ^JavaEqualityWrapper other))))
 
  (hashCode [this]
    hash-code))


  (eval '(let [k (Object.)]
           (JavaEqualityWrapper. nil
                                 (if (nil? k)
                                   0
                                   (.hashCode ^Object k))))))

*compile-path* 
;;= /my/compile/path/classes                  ; <- using this below
```

Then 
```sh
# find the class that just got generated that we are interested in, something like
# user$eval49776.class

javap -v '/my/compile/path/classes/user$eval49776'
```

Relevant bytecode output:
```
  public static java.lang.Object invokeStatic();
    descriptor: ()Ljava/lang/Object;
    flags: ACC_PUBLIC, ACC_STATIC
    Code:
      stack=5, locals=1, args_size=0
         0: new           #13                 // class java/lang/Object
         3: dup
         4: invokespecial #14                 // Method java/lang/Object."<init>":()V
         7: astore_0
         8: new           #16                 // class user/JavaEqualityWrapper
        11: dup
        12: aconst_null
        13: aload_0
        14: aconst_null
        15: invokestatic  #22                 // Method clojure/lang/Util.identical:(Ljava/lang/Object;Ljava/lang/Object;)Z
        18: ifeq          28
        21: getstatic     #26                 // Field const__1:Ljava/lang/Object;
        24: goto          37
        27: pop
        28: aload_0
        29: aconst_null
        30: astore_0
        31: invokevirtual #30                 // Method java/lang/Object.hashCode:()I
        34: invokestatic  #36                 // Method java/lang/Integer.valueOf:(I)Ljava/lang/Integer;
        37: checkcast     #38                 // class java/lang/Number
        40: invokestatic  #44                 // Method clojure/lang/RT.intCast:(Ljava/lang/Object;)I
        43: invokespecial #47                 // Method user/JavaEqualityWrapper."<init>":(Ljava/lang/Object;I)V
        46: areturn

```

Notes, numbered bytecode number above:

* 21: `Field const__1:Ljava/lang/Object;` is the 0 long constant boxed as a Long. This is done in a static initialization block of the class (not shown here for brevity). The cost of the boxing isn't part of the cost of the runtime of this fn, so not a big deal on the boxing side of things.
* 34: `Object.hashCode` returned a primitive int, but it is now getting boxed - probably the worst part of this bytecode since larger ints won't be part of the `Integer.valueOf` pre-defined box Integer object cache.
* 37: At this point, either way the `if (nil? k)` branch goes, we have a boxed `java.lang.Number`
* 40: A clunkier/slower way (do to being a more generic method) to do `Integer.intValue` to unbox
* 43: Finally call `JavaEqualityWrapper` constructor with the correct type signature and runtime types incoming


--------

So things are ok, performance cost is probably negligible due to do the boxing not being a dominant concern in what else is happening (being dwarfed).

You can help the clj compiler not perform the boxing though.

Change the code to this
```clj(binding [*compile-files* true]
  (eval '(let [k nil]
           (JavaEqualityWrapper. nil
                                 (if (nil? k)
                                   (int 0)
                                   (int (.hashCode ^Object k)))))))

```

Figure out which class file was just generated (just sort `ls -ltr` in the `*compile-path*` should be fine)

Say it was `user$eval49895`

```sh
javap -v 'user$eval49895'
```

Relevant bytecode (same chunk as before, but with modifications now):
```
  public static java.lang.Object invokeStatic();
    descriptor: ()Ljava/lang/Object;
    flags: ACC_PUBLIC, ACC_STATIC
    Code:
      stack=5, locals=1, args_size=0
         0: aconst_null
         1: astore_0
         2: new           #13                 // class user/JavaEqualityWrapper
         5: dup
         6: aconst_null
         7: aload_0
         8: aconst_null
         9: invokestatic  #19                 // Method clojure/lang/Util.identical:(Ljava/lang/Object;Ljava/lang/Object;)Z
        12: ifeq          23
        15: lconst_0
        16: invokestatic  #25                 // Method clojure/lang/RT.intCast:(J)I
        19: goto          32
        22: pop
        23: aload_0
        24: aconst_null
        25: astore_0
        26: invokevirtual #31                 // Method java/lang/Object.hashCode:()I
        29: invokestatic  #34                 // Method clojure/lang/RT.intCast:(I)I
        32: invokespecial #37                 // Method user/JavaEqualityWrapper."<init>":(Ljava/lang/Object;I)V
        35: areturn
```

Comparison related comments:

* 16: clj still uses a primitive long for the `0` literal, but the `Method clojure/lang/RT.intCast:(J)I` has no boxing involved, just checks with an `if` for overflow to throw on if happens
* 29: `Object.hashCode` return value is prim int, odd call to `Method clojure/lang/RT.intCast:(I)I`, but it is a "do nothing" method that just returns its arg - so probably super easily inline-able by JIT


Conclusion, no boxing in the 2nd variant. It's a bit tricky to do, clj compiler is a bit too naive about types in a case like `Object.hashCode` returning an prim int, yet it still being boxed anyways.

I won't give into a bunch of different ways to alter things (too long), but you can also do 
```clj

  (eval '(let [k nil]
           (JavaEqualityWrapper. nil
                                 (if (nil? k)
                                   (int 0)
                                   (.hashCode ^Object k))))))
```

In this case, it looks like the clj compiler does recognize that both paths through the `if` result in a prim int, so the boxing of the `Object.hashCode` doesn't happen anymore either.

So it is less `int`s in the code, but I may just put an `int` to be safe. `int` is special in the clj compiler and is "inlined" so it doesn't result in runtime overhead or extra byte codes - it's similar to a type hint really.

rbrush:
Thanks for the detailed analysis, Mike! I went ahead and added the (int ) expressions as suggested to avoid the boxing.

src/test/clojure/clara/test_rules.clj
```diff
@@ -2748,3 +2748,19 @@
                     (fire-rules))]
 
     (is (= [{:?x inner-one}] (query session test-query)))))
+
+;; Ensure that nil values can be bound and unified as expected.
```
WilliamParker: Perhaps this comment should refer to [issue 393](https://github.com/cerner/clara-rules/issues/393) as well since it seems fairly specific to it as it has the nested record pattern.

rbrush:
I've moved the tests to the test-bindings namespace as suggested. It looks like this has incorporated all suggestions and is better than what current exists in master, so I'll plan on squash/merging this shortly.

WilliamParker:
It looks like you removed the "Test for issue 393" comment (probably inadvertently) when moving the tests over.  +1 to merge though.

rbrush:
Merged! (I also re-added the "test for issue 393 comment before doing so.) Thanks for the feedback, everyone!

PR #400, opened 2018-07-16
Issue 391: Update landing page to reflect ClojureScript support

WilliamParker:
This should make it more obvious to any new users landing on GitHub page that Clara has ClojureScript support.  Given the trivial nature of the change I'll plan to put it in tomorrow unless there are objections.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # _About_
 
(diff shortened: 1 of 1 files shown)


EthanEChristian:
+1

Issue #402, opened 2018-07-20
[performance] Remove usage of get-in in compiled rules


EthanEChristian:
When clara creates [join-filter-fns](https://github.com/cerner/clara-rules/blob/master/src/main/clojure/clara/rules/compiler.clj#L460) it typically creates a let block with assignments. These assignments are extracted from the token's bindings via `get-in`, using a simple tests:

``` Clojure
(ns clara.performance.test-fire-rules
  (:require [clojure.test :refer :all]
            [clara.rules.accumulators :as acc]
            [clara.rules :as r]
            [clara.rules.compiler :as com]))

<Performance test code written as part of #381 copied over>

(defrecord AFact [id])
(defrecord BFact [id])
(defrecord ParentFact [a-ids b-ids])

(def counter (atom {:a-count 0
                    :b-count 0}))

(deftest test-get-in-perf
  (let [accum-rule [{:ns-name (ns-name *ns*)
                     :lhs [{:type ParentFact
                            :constraints []
                            :fact-binding :?parent}
                           {:accumulator `(acc/all)
                            :from {:type AFact
                                   :constraints ['(contains? (:a-ids ?parent) id)]}
                            :result-binding :?as}
                           {:accumulator `(acc/all)
                            :from {:type BFact
                                   :constraints ['(contains? (:b-ids ?parent) id)]}
                            :result-binding :?bs}]
                     :rhs '(do (swap! counter update :a-count inc))}]

        parents (for [_ (range 5000)]
                  (->ParentFact
                    (set (range 400))
                    (set (range 400 800))))
        a-facts (for [id (range 5000)]
                  (->AFact id))
        b-facts (for [id (range 5000)]
                  (->BFact id))

        facts (doall (concat parents
                             a-facts
                             b-facts))

        session (com/mk-session [accum-rule])]
    (run-performance-test {:description "Slow get-in perf"
                           :func #(-> session
                                      (r/insert-all facts)
                                      (r/fire-rules))
                           :iterations 5
                           :mean-assertion (partial > 30000)})))
```

The rule in question in clara's dsl would look like:
``` Clojure
(r/defrule test-rule
  [?parent <- ParentFact]
  [?as <- (acc/all) :from [AFact (contains? (:a-ids ?parent) id)]]
  [?bs <- (acc/all) :from [BFact (contains? (:b-ids ?parent) id)]]
  =>
  (swap! counter update :a-count inc))
```

[...]

PR #404, opened 2018-07-28
Issue #402: [performance] Remove usage of get-in in compiled rules

EthanEChristian:
I have incorporated some of the performance test helpers created with #381, I placed them where both cljs and clj could leverage them. However running benchmarks on cljs on the scale of some of the clj benchmarks is considerably worse, thus i have created a benchmark only on the clj side.

I am not really sure how to go about profiling cljs, but it might be worth looking into how we could improve the execution of rules.

### DIFF
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -31,7 +31,9 @@
             Accumulator
             ISystemFact]
            [java.beans
-            PropertyDescriptor]))
+            PropertyDescriptor]
+           [clojure.lang
+            IFn]))
 
 ;; Protocol for loading rules from some arbitrary source.
 (defprotocol IRuleSource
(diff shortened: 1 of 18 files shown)


EthanEChristian:
#402

EthanEChristian:
Further explanation of my comment above:
```
However running benchmarks on cljs on the scale of some of the clj benchmarks is considerably worse
```

The clj test runs in ~400ms the cljs test runs in ~490000ms

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -378,10 +378,16 @@
              ~'?__bindings__ (atom ~initial-bindings)]
          ~(compile-constraints constraints)))))
 
+(defn build-token-assingment
+  "A helper function to build variable assignment forms for tokens."
+  [binding-key]
+  (list (symbol (name binding-key))
+        (list '-> '?__token__ :bindings binding-key)))
```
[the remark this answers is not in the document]
WilliamParker: +1 ; this is in case someone defines their own -> in the namespace of the production.  We'd still want this to be the clojure.core version.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -378,10 +378,16 @@
              ~'?__bindings__ (atom ~initial-bindings)]
          ~(compile-constraints constraints)))))
 
+(defn build-token-assingment
+  "A helper function to build variable assignment forms for tokens."
+  [binding-key]
+  (list (symbol (name binding-key))
```
WilliamParker: You could experiment with accessing the :bindings by the literal field as well if this is a performance hotspot.  I don't know if it will matter in any siginificant way, but the keyword lookup does have overhead versus a Java field member access at least on the JVM.  I think (not sure offhand) the syntax for that to work in both clj and cljs is .-fieldname.

src/test/clojure/clara/performance/test_rule_execution.clj
```diff
@@ -0,0 +1,39 @@
+(ns clara.performance.test-rule-execution
```
WilliamParker: Why is this Clojure (JVM)-only?

src/main/clojure/clara/rules/compiler.clj
WilliamParker: Note: not needed for merge, just an observation/food for thought.

src/test/clojure/clara/performance/test_rule_execution.clj
EthanEChristian: I put this in a JVM only location because the cljs test took 490 seconds, and it didn't seem that it would provide much value.

EthanEChristian: I think that this is the only outstanding comment, @WilliamParker  do you feel strongly that this should also run for CLJS?

EthanEChristian:
@WilliamParker I have updated the PR to cover the changes requested, if there is nothing further i will try to get around to merging this one.

EthanEChristian:
Logged #406 to setup the cljs build for performance.

PR #407, opened 2018-09-23
Issue 357: :test conditions could incorrectly be considered to create…

WilliamParker:
… bindings in the compiler


rbrush:
+1; LGTM. Thanks for the fix, @WilliamParker!

PR #409, opened 2018-11-01
Allow to run on jre9

sunilgunisetty:


project.clj
```diff
@@ -27,6 +27,8 @@
   :source-paths ["src/main/clojure"]
   :resource-paths []
   :test-paths ["src/test/clojure" "src/test/common"]
+  :jvm-opts ["-XX:+IgnoreUnrecognizedVMOptions"
+             "--add-modules=java.xml.bind"]
```
mrrodriguez: We could add a new profile named something like `:jdk9+` that added this "--add-modules=java.xml.bind" to `:jvm-opts`.

If that were done, then there wouldn't be a need to make this top-level `:jvm-opts` have to "-XX:+IgnoreUnrecognizedVMOptions" for all JVM versions. 

We could also probably autodetect if "--add-modules=java.xml.bind" should be added via some lein project load time-eval (if worthwhile).

Is there a particular advantage to setting it here at the root of the project for all JVMs instead that you see? I guess it is mostly fine, except if there a re more options, it hides the fact that they may be specified incorrectly.

mrrodriguez:
Be sure to check out the https://github.com/cerner/clara-rules/blob/master/CONTRIBUTING.md too

project.clj
WilliamParker: I agree that it would be better to add an explicit additional profile than risk having wrong arguments.  Also, some quick Googling suggests this is coming from a ClojureScript dependency: https://dev.clojure.org/jira/browse/CLJS-2377 . Is that the case or are we getting this from somewhere else?  Also, will this have any impact on the jar built on the JVM or is it just for the JVM running during the build?

sunilgunisetty: I changed the PR by adding java9 as a lein profile.

mrrodriguez:
> My thinking here is to merge this PR and log a followup issue to detect the Java version automatically in Leiningen/investigate what such options exists. This can't be the only project that has encountered this issue.

@WilliamParker That seems fine to me. I believe you could get that to work with some dynamic project eval stuff, but doesn't need to be done in this PR.

WilliamParker:
@sunilgunisetty I've merged this and logged https://github.com/cerner/clara-rules/issues/412 as discussed above for looking into other ways to address this later, such as the setup automatically determining what JVM is running and adjusting accordingly.  Thanks for bringing this up and sorry for the lengthy delay.

sunilgunisetty:
No problem. Thanks for the great work on Clara Rules !

Issue #260, opened 2017-01-22
Create optimized Fressian handlers for tokens and elements


WilliamParker:
Currently Clara elements and tokens use a general Fressian handler for arbitrary Clojure records.  This works, but is likely suboptimal for performance.  This includes both things that impact records in general and things specific to elements and tokens due to the structure of their contents.  Due to the sheer number of tokens and elements that can exist in large rules sessions I believe these optimizations have the potential to significantly improve performance of the Fressian durability implementation.

1. Since the handler is built for arbitrary record types, it has to serialize the record with the name of its constructor function as seen [here](https://github.com/cerner/clara-rules/blob/0.13.0-RC7/src/main/clojure/clara/rules/durability/fressian.clj#L117).  After deserialization, this name is then used as a symbol to resolve the function.  We've seen these calls to clojure.core/resolve be a hotspot during session deserialization.  Furthermore, the function name obviously takes up space in the serialized session.
2. Since the record is serialized as an arbitrary map [here](https://github.com/cerner/clara-rules/blob/0.13.0-RC7/src/main/clojure/clara/rules/durability/fressian.clj#L75) we have to serialize the keys, which takes space in the serialized session although probably not much since they are cached.  Deserializing the keys takes time as well.  Furthermore, when reading the record we create a map of the contents that are then passed to the constructor function [here](https://github.com/cerner/clara-rules/blob/0.13.0-RC7/src/main/clojure/clara/rules/durability/fressian.clj#L138).  Since this map is not used once the record is created, it would be more efficient to use something that could call the token or element's positional constructor without needing to build this map.
3. We know that we don't use metadata on tokens or elements but we still serialize and deserialize an empty map for them.
4. There are places where we it may make sense to use caching with an InternalMemIdx marker on structures where it doesn't make sense in general for records.  In particular, many of the tuples inside the :matches of tokens can be expected to be identical to other such tuples due to the many places in the engine where we take the existing vector of :matches and conj on a new one to build a new token, such as [here](https://github.com/cerner/clara-rules/blob/0.13.0-RC7/src/main/clojure/clara/rules/engine.cljc#L489). Bindings maps could benefit in some cases as well, although probably not as much as token match tuples.  For example, RootJoinNode just takes a binding map from an element and uses it in a token [here](https://github.com/cerner/clara-rules/blob/0.13.0-RC7/src/main/clojure/clara/rules/engine.cljc#L456).

mrrodriguez:
[...]

PR #410, opened 2018-11-01
Add retract-fn to accumulator

sunilgunisetty:
Its not obvious `retract-fn` is part of accumulator. Adding it make it more explicit.

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -14,7 +14,7 @@
 ;; The accumulator is a Rete extension to run an accumulation (such as sum, average, or similar operation)
 ;; over a collection of values passing through the Rete network. This object defines the behavior
 ;; of an accumulator. See the AccumulateNode for the actual node implementation in the network.
-(defrecord Accumulator [initial-value reduce-fn combine-fn convert-return-fn])
+(defrecord Accumulator [initial-value retract-fn reduce-fn combine-fn convert-return-fn])
 
 ;; A Rete-style token, which contains two items:
 ;; * matches, a vector of [fact, node-id] tuples for the facts and corresponding nodes they matched.
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -20,3 +20,4 @@ Community
 [@dgoeke]: https://github.com/dgoeke
 [@sparkofreason]: https://github.com/sparkofreason
 [@bfontaine]: https://github.com/bfontaine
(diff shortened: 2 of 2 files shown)


mrrodriguez:
Be sure to check out the https://github.com/cerner/clara-rules/blob/master/CONTRIBUTING.md too

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -14,7 +14,7 @@
 ;; The accumulator is a Rete extension to run an accumulation (such as sum, average, or similar operation)
 ;; over a collection of values passing through the Rete network. This object defines the behavior
 ;; of an accumulator. See the AccumulateNode for the actual node implementation in the network.
-(defrecord Accumulator [initial-value reduce-fn combine-fn convert-return-fn])
+(defrecord Accumulator [initial-value retract-fn reduce-fn combine-fn convert-return-fn])
```
mrrodriguez: Adding this may add clarity to someone reading it. Right now we just set the key on the record map via the factory fn `map->Accumulator`. Having this as a field is not strictly necessary and wouldn't really be of any real performance benefit (this field does not appear to be accessed enough to have significant performance implications).

I think it seems fine to add it here for readability. However, I was wondering if anyone else had any thoughts on if this could cause any sort of breakage anywhere (I don't think so, and I'd hope not).

Also, I think the retract-fn itself is poorly documented. I don't see many details on defining your own accumulators at all in our docs, so that'd be a good place to see improvements.

In particular, it recently reoccurred to me that the retract-fn isn't even always used, even if provided. It is more of an "optimization hint" that may be used if applicable. Examples of it not being used, would be the AccumulateWithJoinFilterNode that reaccumulates from the candidate facts each time.

WilliamParker: Not having the retract-fn in the record seems unclear to me so I agree with adding it.  As far as breaking anything, I think we're fine.  Since records take additional keys as you say the only way I think we'd break anything is with durability.  However, the [Fressian record handler](https://github.com/cerner/clara-rules/blob/0.19.0/src/main/clojure/clara/rules/durability/fressian.clj#L370) just writes the record out as a map with information to call its map-> constructor on deserialization.  That said, it is conceivable that someone has created a durability implementation that has optimizations that would be broken by this change - for an idea of the things that could be done see [issue 260](https://github.com/cerner/clara-rules/issues/260), but I don't think it is likely.  I'd suggest just making a note in the changelog that any users with their own durability implementation for the rulebase should look at this PR.  @mrrodriguez

Issue #388, opened 2018-03-14
Improve CLJS dev setup and testing org


mrrodriguez:
The CLJS dev setup is less than ideal. I'd like to get it working with a more modern figwheel/sidecar sort of setup.

Also, the cljs testing seems to be disorganized. I think it is not very clear where cljs & cljc files are found at. I'm not even sure it makes sense to have a "common", "clojurescript", and "clojure" directories under "src/test" anymore since read conditionals and the various file extensions are respected now. It makes the sharing of files more confusing and hard to find in my opinion.

I only bring up the testing setup since there seem to be issues getting Figwheel compiling these namespace and to get it all working again, I'd likely want to get this stuff sorted out.

WilliamParker:
+1 to doing this.  I haven't had the time or familiarity/more time to become familiar myself but I agree that the CLJS setup could use improvement and from what I read it sounds like the overall CLJS dev ecosystem has come a long way since Clara's CLJS build was set up.

sparkofreason:
You might take a look at [shadow-cljs](https://shadow-cljs.github.io/docs/UsersGuide.html) as well. We went from boot to figwheel and finally settled on shadow-cljs, have been very happy with it.

mrrodriguez:
@sparkofreason

I’ve been following shadow cljs a bit. I agree it seems useful in some cases. For Clara though there really isn’t any fancy JS interop dependency management to worry about. So I planned to just do a pretty straightforward lein figwheel setup for now. The biggest problem right now is some of these `defsession` sort of cljs namespaces aren’t compiling with figwheel. 

Also, Clara supports a quite old cljs version. For development time, I think I’m going to use a more recent one for dev tooling sake. We can still run tests against whatever old version we want to support. 

 I am going to look closer at the figwheel compilation problem stuff and try to figure out what to do about our `defsession` Usage once I understand it a bit more.

alex-dixon:
Here's the leiningen/ figwheel config I borrowed from Luminus that I posted a while back: https://github.com/cerner/clara-rules/issues/285#issuecomment-298922073.

I still use this for CLJS REPL with figwheel. There may be better alternatives these days but it works for me.

mrrodriguez:
Thanks @alex-dixon I'll check it out. I'm pretty familiar with cljs and figwheel stuff nowadays since I've been using it at work for a while now. It's good to see what you have though.

[...]

PR #413, opened 2018-12-10
Use JAXB from Maven

### LINKED ISSUE #412: Better options for Java 9 compatibility in project.clj
In https://github.com/cerner/clara-rules/pull/409 we added a profile with arguments necessary to run under Java 9.  This profile will only be used during development by those running a Java 9 JVM.  I suspect that it is possible to autodetect this in the project setup and adjust accordingly.  Alternatively other Clojure(Script) projects may have landed on other standard approach(es) that we might consider.

eraserhd:
Closes #412

This should work on all the JDKs.  I've tested it on 8, 10, and 11.


WilliamParker:
As I understand it the need for these dependencies comes from our (relatively old) version of ClojureScript.  According to [CLJS-2377](https://dev.clojure.org/jira/browse/CLJS-2377) and https://github.com/bhauman/lein-figwheel/issues/612#issuecomment-377210625 ClojureScript has now released a version that doesn't have this dependency anymore, so if Clara bumped its dependency to that ClojureScript version we wouldn't need to declare this dependency.  However, we'd be blind to anything that broke Clara in older ClojureScript versions unless we did further testing.  We could create an "old ClojureScript" build similar to the changes made in https://github.com/cerner/clara-rules/pull/358 to test against newer Clojure/ClojureScript, have that build add these dependencies, and then bump the version of ClojureScript in the primary build to a newer one.  This would ensure that we maintained compatibility with older ClojureScript versions.  

The reason I see for doing it this way is that as is, we'd be blind to any actual runtime dependency in the JVM version on these additional dependencies since all tests would pass and users would then be unable to use the jar (unless they added the dependencies).  I'm a bit hesitant to assume that no Clara JVM dependencies, now or in the future, won't have an actual dependency on these sorts of "systems level" dependencies being provided.  That said, I don't see it as a huge risk and in any event it is something that would quickly become apparent upon attempted use of Clara, so if what I'm describing won't work/has downsides I'm missing I don't mind merging this too much.

As an ancillary benefit, any performance testing we do against ClojureScript should really be against a newer version, even if we maintain compatibility against older versions.  

Thoughts?  I'm particularly interested in thoughts from anyone who keeps up with the ClojureScript ecosystem more than I do.

@mrrodriguez @eraserhd

mrrodriguez:
@WilliamParker It makes more sense to me for Clara to build against fairly modern versions of clj and cljs as the default and have a few "older" versions profiles to test out level of a minimum required version.

Perhaps something like 1.9 of clj, with some later 1.9.x of cljs as the default, then a profile for more cutting edge clj 1.10 and cljs 1.10.x and a profile for a minimum tested version we consider Clara to claim to have compatibility with.

eraserhd:
I figure this was a good thing for me to resolve for the next release, but here is the issue:

I can update Clojure and ClojureScript in the `:recent-clj` profile, and then I can run the unit tests without loading new libraries for JDK11.  This is good...

But then the PhantomJS tests fail, though, and it appears to be a problem with PhantomJS's ES6 support.  The issue is here: https://github.com/ariya/phantomjs/issues/13895 .

Then I discovered (did not know) that PhantomJS is officially dead, and Everyone Should Use Headless Chrome.  So this seems like a little bit of work, including remembering how to Travis CI again.

mrrodriguez:
Hi @eraserhd 
I have a long running issue I keep failing to solve and then not getting back to @ https://github.com/cerner/clara-rules/issues/388

The CLJS implementation really needs an overhaul due to pitfalls in how it is done.
Along with this, `phantomjs` should stop being used in favor of something supported as you mentioned.

I think `lein doo` with chrome headless would be sufficient, but we need to get the Travis CI setup to work for this.

What worries me some is that locally I've ran into a lot of CLJS sort of failures doing things like compiling and testing due to the issues w/how CLJS rules are compiled via a sneaky/not-really-supported `clojure.core/eval` during CLJS compilation. 

Perhaps I can try to revisit this yet again, or try to just do a single issue to stop using phantom and see if it still works as-is for now...

PR #415, opened 2019-01-15
Make negation tests cross-platform (clj/cljs)

WilliamParker:


WilliamParker:
"So it seems the same tests are still here, just retrofitted to the cross-platform style. Is that the full scope of the changes?" => correct @mrrodriguez .  I made a small change to test-complex-negation so that we could have an "object query" in ClojureScript, namely adding a common ancestor to all (user) facts via the ancestors-fn, but otherwise this should be a straightforward copy into the def-rules-test format.

Issue #149, opened 2015-12-15
Negated conjunctions and binding visibility


mrrodriguez:
I'd expect the following to work:

```clj
(let [q (dsl/parse-query []
                         [[:not [:and
                                 [?t <- Temperature]
                                 [Cold (= temperature (:temperature ?t))]]]])
      s (mk-session [q])]
  (-> s
      fire-rules
      (query q)))
```

However, I get:
```clj
ExceptionInfo Using variable that is not previously bound. This can happen when an expression uses a previously unbound variable, or if a variable is referenced in a nested part of a parent expression, such as (or (= ?my-expression my-field) ...). 
Unbound variables: #{?t}  clojure.core/ex-info (core.clj:4593)
```

I understand why this is though.  The Rete network is normalized in disjunctive normal form (DNF).
So this comes out to be something like:
```clj
(clara.rules.compiler/to-dnf '[:not [:and
                                     {:type Temperature
                                      :constraints []
                                      :fact-binding :?t}
                                     {:type Cold
                                      :constraints ['(= temperature (:temperature ?t))]}]])
;= [:or
;=   [:not {:type Temperature, :constraints [], :fact-binding :?t}]
;=   [:not {:type Cold, :constraints [(quote (= temperature (:temperature ?t)))]}]]
```

Since the two conditions get split across an `:or` the second condition no longer has visibility to the binding `?t`.

I think this one is going to be a little tricky to solve.

I know that I can manually extract the conjunction into another rule and get the behavior I'm looking for like:
```clj
(let [r (dsl/parse-rule [[?t <- Temperature]
                         [Cold (= temperature (:temperature ?t))]]
                        (insert! (with-meta {:marker true}
                                   {:type :__stub})))
      q (dsl/parse-query []
                         [[:not [:__stub]]])
      s (mk-session [r q])]
  (-> s
      fire-rules
      (query q)))

;= ({})
```

Given that, I was beginning to think that perhaps the simplest way to handle these sorts of scenarios would be for the Clara compiler to automatically extract out a rule like this before the network is put into DNF.  This would allow consumers to not have to get confused about why it doesn't work when it looks like it should.

This is just my first inclination though.  I haven't really wrapped my head around what else could be done in terms of doing something more sophisticated with the DNF or perhaps some sort of subnetwork.
[...]

Issue #344, opened 2017-09-26
Create function that uses tracing to return rules with potential performance problems


WilliamParker:
Currently we have [tracing functionality](https://github.com/cerner/clara-rules/blob/0.16.0/src/main/clojure/clara/tools/tracing.cljc) that contains abundant information about the flow of data through the session.  When rules sessions experience performance problems this tracing can provide information on what rule(s) need to be optimized.  However, this information is difficult to interpret; often significant knowledge of Clara's internals is required to do so.  The full trace necessarily contains the richest information, but it should be possible to provide utility function(s) that provide a quick snapshot of what rules are likely to be performance problems in simple cases.  I envision something along the lines of:

- Create a map of node IDs to the count of operations on these nodes.
- Associate each node ID with the rules and queries that it feeds into.
- Return some data structure representing rules/queries with associated counts.  For example, we could return a list with entries like [rules-and-queries-names-list node-type-keyword interaction-count], one entry per node with an ID, in descending order based on the interaction-count.

The objective of this would not be to represent all the information available in the tracing, but to provide a rough snapshot.  My experience is that in practice many rules performance problems are revealed by analysis of rules traces at this level.  For example, rules that are simply provided with a large amount of data, a small number of rules that repeatedly create work for each other, etc.  I'd like the interface to be simple enough that it could be used in logging to log out something to the effect of "Experienced a potential performance problem in rule X with maximum interaction count N".  It should be noted that full tracing needs to be turned on for this to be possible; under what circumstances the performance consequences of this are acceptable will vary depending on the use-case.

WilliamParker:
Pull request at https://github.com/cerner/clara-rules/pull/416

WilliamParker:
Closing issue since I've merged the PR

PR #416, opened 2019-01-17
Issue 344: Add a function for reporting potential performance problem…

WilliamParker:
…s in a traced session

Pull request for [issue 344](https://github.com/cerner/clara-rules/issues/344)
@mrrodriguez this is one of the tracing helpers we've discussed before.

@EthanEChristian this is probably of interest to you.  Possibly @eraserhd as well if I recall some discussions on the Slack channel in recent months correctly.

Some remaining items to address:
- How does this handle nodes with more than one descendant terminal node?  I'd like to add tests for this.
- This will report autogenerated rules created from the extraction added to fix https://github.com/cerner/clara-rules/issues/149 . This is an existing issue with session inspection as well though - I don't think addressing this needs to block a merge.

### DIFF
--- a/src/main/clojure/clara/tools/tracing.cljc
+++ b/src/main/clojure/clara/tools/tracing.cljc
@@ -114,3 +114,48 @@
                          (first))]
     (.-trace ^PersistentTracingListener listener)
     (throw (ex-info "No tracing listener attached to session." {:session session}))))
+
+(defn ^:private node-id->productions
+  "Given a session and a node ID return a list of the rule and query names associated
+   with the node."
+  [session id]
+  (let [node (-> session .-rulebase :id-to-node (get id))]
+    (into []
+          (comp
+           (map second)
+           cat
+           (map second))
+          (eng/get-conditions-and-rule-names node))))
+
+(defn traced-session->ranked-productions
+  "Given a session with tracing enabled, return a map of rule and query names
+  to a numerical index that represents an approximation of the proportional
+  amount of times Clara performed processing related to this rule.  This
+  is not intended to have a precise meaning, and is intended solely as a means
+  to provide a rough guide to which rules and queries should be considered
+  the first suspects when diagnosing  performance problems in rules sessions.
+  It is possible for a relatively small number of interactions to take a long
+  time if those interactions are particularly costly.  It is expected that
+  the results may change between different versions when Clara's internals change,
+  for example to optimize the rules network.  Nevertheless, it is anticipated
+  that this will provide useful information for a first pass at rules
+  performance problem debugging.  This should not be used to drive user logic.
+
+  This currently returns a Clojure array map currently in order to conveniently have the rules
+  with the most interactions printed first in the string representation of the map."
+  [session]
+  (let [node-ids (->> session
+                      get-trace
+                      (map :node-id))
+
+        production-names (into []
+                               (comp
+                                (map (partial node-id->productions session))
+                                cat)
+                               node-ids)
+
+        production-name->interactions (frequencies  production-names)
+
+        ranked-tuples (reverse (sort-by second production-name->interactions))]
+
+    (apply array-map (into [] cat ranked-tuples))))
--- a/src/main/clojure/clara/tools/tracing.cljc
+++ b/src/main/clojure/clara/tools/tracing.cljc
@@ -119,7 +119,7 @@
   "Given a session and a node ID return a list of the rule and query names associated
    with the node."
   [session id]
-  (let [node (-> session .-rulebase :id-to-node (get id))]
+  (let [node (-> session eng/components :rulebase :id-to-node (get id))]
     (into []
           (comp
            (map second)
@@ -127,7 +127,7 @@
            (map second))
           (eng/get-conditions-and-rule-names node))))
 
-(defn traced-session->ranked-productions
+(defn ranked-productions
   "Given a session with tracing enabled, return a map of rule and query names
   to a numerical index that represents an approximation of the proportional
   amount of times Clara performed processing related to this rule.  This
@@ -141,7 +141,7 @@
   that this will provide useful information for a first pass at rules
   performance problem debugging.  This should not be used to drive user logic.
 
-  This currently returns a Clojure array map currently in order to conveniently have the rules
+  This currently returns a Clojure array map in order to conveniently have the rules
   with the most interactions printed first in the string representation of the map."
   [session]
   (let [node-ids (->> session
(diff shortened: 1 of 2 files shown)


src/main/clojure/clara/tools/tracing.cljc
```diff
@@ -114,3 +114,48 @@
...
+  [session id]
+  (let [node (-> session .-rulebase :id-to-node (get id))]
+    (into []
+          (comp
+           (map second)
+           cat
+           (map second))
+          (eng/get-conditions-and-rule-names node))))
+
+(defn traced-session->ranked-productions
```
EthanEChristian: It might also be interesting to have something similar to pair with this. Once we have the production name it might be nice to be able to easily get the nodes and number of interactions that make up a specific production. This could probably be added in the future if we see the need though.

mrrodriguez: I think I'd probably prefer the name to just be`ranked-productions`.

I'm not sure there is a reason to express "conversion" that the `x->y` tends to imply.
I don't 100% agree [1] with the suggestions of this post [2], but do think it has some fairly good guidance in this regard

[1] It does mention 
> There are exceptions to every rule.
So it covers that base.

[2] https://stuartsierra.com/2016/01/09/how-to-name-clojure-functions

src/main/clojure/clara/tools/tracing.cljc
```diff
@@ -114,3 +114,48 @@
...
+  to provide a rough guide to which rules and queries should be considered
+  the first suspects when diagnosing  performance problems in rules sessions.
+  It is possible for a relatively small number of interactions to take a long
+  time if those interactions are particularly costly.  It is expected that
+  the results may change between different versions when Clara's internals change,
+  for example to optimize the rules network.  Nevertheless, it is anticipated
+  that this will provide useful information for a first pass at rules
+  performance problem debugging.  This should not be used to drive user logic.
+
+  This currently returns a Clojure array map currently in order to conveniently have the rules
```
mrrodriguez: "currently" is repeated in this sentence.

src/main/clojure/clara/tools/tracing.cljc
```diff
@@ -114,3 +114,48 @@
                          (first))]
     (.-trace ^PersistentTracingListener listener)
     (throw (ex-info "No tracing listener attached to session." {:session session}))))
+
+(defn ^:private node-id->productions
```
mrrodriguez: Similar to my comment for `traced-session->ranked-productions`,
I'd prefer `node-productions` since this doesn't seem like a "conversion" operation.

It's subjective of course.

src/main/clojure/clara/tools/tracing.cljc
```diff
@@ -114,3 +114,48 @@
...
+  "Given a session and a node ID return a list of the rule and query names associated
+   with the node."
+  [session id]
+  (let [node (-> session .-rulebase :id-to-node (get id))]
```
mrrodriguez: It'd be preferable to remove the reflective access here. It doesn't matter perf-wise in this case of course, but the less reflection warnings in the codebase the less noise there is to find the important ones.

src/main/clojure/clara/tools/tracing.cljc
```diff
@@ -114,3 +114,48 @@
...
+
+        production-names (into []
+                               (comp
+                                (map (partial node-id->productions session))
+                                cat)
+                               node-ids)
+
+        production-name->interactions (frequencies  production-names)
+
+        ranked-tuples (reverse (sort-by second production-name->interactions))]
```
mrrodriguez: The `reverse` could be avoided by using the 3 arity `sort-by`.
I believe that'd be
```clj
ranked-tuples (sort-by second > production-name->interactions)
```
or could generalize as
```clj
ranked-tuples (sort-by second #(compare %2 %1) production-name->interactions)
```

src/main/clojure/clara/tools/tracing.cljc
```diff
@@ -114,3 +114,48 @@
...
+
+    (apply array-map (into [] cat ranked-tuples))))
```
mrrodriguez: Side note: It could be a `sorted-map-by` with the value-info tied to metadata or something like that.
However, this is simpler and seems fine for now.

src/main/clojure/clara/tools/tracing.cljc
WilliamParker: That name is OK with me, will change.

src/main/clojure/clara/tools/tracing.cljc
WilliamParker: +1 good catch, will change

src/main/clojure/clara/tools/tracing.cljc
WilliamParker: Perhaps "productions-for-id"?

src/main/clojure/clara/tools/tracing.cljc
WilliamParker: Agreed that that would work.  IMO this is a little easier to understand though and performance doesn't seem like a concern here.  I'm inclined to leave it as is.

src/main/clojure/clara/tools/tracing.cljc
WilliamParker: Yeah, that could work if we used symbols as the keys.  Since strings don't support metadata you'd need to do something more complicated involving storing the "metadata" elsewhere and make sure you didn't create a memory leak in the process.  Alternatively the keys could be symbols instead.  We'd also need to make sure the compared value was never equal, perhaps by adding a tiebreak on alphabetical order, since sorted-map-by will consider equal values on the comparator to be equal keys.

src/main/clojure/clara/tools/tracing.cljc
WilliamParker: Agreed, there's lots of interesting views on the tracing data that one could provide.

src/main/clojure/clara/tools/tracing.cljc
WilliamParker: We can just use the components function on the session actually and avoid direct field access entirely.  It's better to use the protocols on the session anyway in case there was another implementation at some point. https://github.com/cerner/clara-rules/blob/0.19.0/src/main/clojure/clara/rules/engine.cljc#L1990

WilliamParker:
I'm going to go ahead and merge this; my last commit addressed review comments.  If there's anything outstanding to address I'm open to making further changes but it seems like we've covered everything of importance and this is not a change to the core compiler/engine/memory code demanding higher scrutiny.

Issue #419, opened 2019-02-05
[Durability] Increased memory usage post Serde.


EthanEChristian:
I have been investigating memory usage when creating and serializing sessions, and i ran across the fact that pre-serialization session consume less heap than after serialization. 

For example i have a session that has 31478 productions, looking at heap dumps:

Pre-Serialization:
![clara_local_session_pre-serialization](https://user-images.githubusercontent.com/3660185/52283409-04f35f80-2928-11e9-84b9-b6f5457d9ab5.png)

Post-Serialization(new JVM):
![clara_local_session_post-serialization](https://user-images.githubusercontent.com/3660185/52283472-218f9780-2928-11e9-891e-8bbd48367a10.png)

 Talking with @mrrodriguez, he pointed to the [fressian](https://github.com/cerner/clara-rules/blob/0.19.0/src/main/clojure/clara/rules/durability/fressian.clj#L208-L346) handlers for the clojure data structures. The problem with them is that pre-serialization a lot of the clara rulebase uses common references to constraints from rules, but when serialized these references are not recreated, meaning that large portions of productions become duplicated. 

I'm going to try and use an identity based cache for serializing the session in such a way that references can be maintained during deserialization, similar to the way we handle records today.

WilliamParker:
I'm surprised that the difference is so large actually, I suspect that there is indeed room for significant improvement.  Keep in mind that when you perform operations on a Clojure data structure, depending on the particular case in question both the original data structure and the new view will share some underlying Java objects.  This is how Clojure performs these operations efficiently - if it had to copy on write it would be much slower.  This blog is a good place to start if you haven't read on this before: https://hypirion.com/musings/understanding-persistent-vector-pt-1 . Unfortunately taking advantage of this sharing could be difficult or infeasible during SerDe, and you might have to weigh CPU vs memory costs.  However, IMO there are good odds that there is much lower hanging fruit available to reduce this memory use increase.

EthanEChristian:
@WilliamParker 
I agree that we probably can't reproduce the sharing at that level, but we should be able to replicate identical collections on SerDe. From what i can tell, the vast majority of duplication comes from the constraints of the rules themselves, as these will be held by the nodes and the pre-eval'd fns for those nodes.

mrrodriguez:
@WilliamParker 
That is a good point when it comes to altering large data structures. I wonder how much that is done at the ruleset serde level.

[...]

PR #420, opened 2019-02-06
#419: [Durability] Increased memory usage post Serde.

EthanEChristian:
I think this is the easiest solution for #419.

Comparing the Heap dumps provided on the original issue, the heap dumps now look like:
Pre-Serialization:
new_clara_local_session_pre-serialization
Post-Serialization:
new_clara_local_session_post-serialization

While I would like to see the after heap usage closer to that of the before, i believe that Will's [comment](https://github.com/cerner/clara-rules/issues/419#issuecomment-460848471) explains why this is unlikely possible.

Running the Compilation/Serialization performance tests:
Before:
lein_performance_tests_before

After:
lein_performance_tests_after

Another thing to note is that prior to these changes the Session took up ~137MB on disk and after it was ~62.4MB. (This metric applies to the Heap dumps above, not the performance tests.)

@WilliamParker  @mrrodriguez

### DIFF
--- a/src/main/clojure/clara/rules/durability/fressian.clj
+++ b/src/main/clojure/clara/rules/durability/fressian.clj
@@ -191,6 +191,40 @@
                       (read-record add-node-expr-fn)
                       d/cache-node)))}}))
 
+(defn- create-identity-based-handler
+  [clazz
+   tag
+   read-fn
+   write-fn]
+  (let [indexed-tag (str tag "idx")]
+    ;; Write an object a single time per object reference to that record.  The object is then "cached"
+    ;; with the IdentityHashMap `d/clj-struct-holder`.  If another reference to this object instance
+    ;; is encountered later, only the "index" of the object in the map will be written.
+    {:class clazz
+     :writer (reify WriteHandler
+               (write [_ w o]
+                 (if-let [idx (d/clj-struct->idx o)]
+                   (do
+                     (.writeTag w indexed-tag 1)
+                     (.writeInt w idx))
+                   (do
+                     (write-fn w tag o)
+                     (d/clj-struct-holder-add-fact-idx! o)))))
+     ;; When reading the first time a reference to an object instance is found, the entire object will
+     ;; need to be constructed.  It is then put into indexed cache.  If more references to this object
+     ;; instance are encountered later, they will be in the form of a numeric index into this cache.
+     ;; This is guaranteed by the semantics of the corresponding WriteHandler.
+     :readers {indexed-tag
+               (reify ReadHandler
+                 (read [_ rdr _ _]
+                   (d/clj-struct-idx->obj (.readInt rdr))))
+               tag
+               (reify ReadHandler
+                 (read [_ rdr _ _]
+                   (-> rdr
+                       read-fn
+                       d/clj-struct-holder-add-obj!)))}}))
+
 (def handlers
   "A structure tying together the custom Fressian write and read handlers used
    by FressianSessionSerializer's."
@@ -206,36 +240,29 @@
                   (resolve (.readObject rdr))))}}
 
    "clj/set"
-   {:class clojure.lang.APersistentSet
-    :writer (reify WriteHandler
-              (write [_ w o]
-                (write-with-meta w "clj/set" o)))
-    :readers {"clj/set"
-              (reify ReadHandler
-                (read [_ rdr tag component-count]
-                  (read-with-meta rdr set)))}}
+   (create-identity-based-handler
+     clojure.lang.APersistentSet
+     "clj/set"
+     (fn clj-set-reader [rdr] (read-with-meta rdr set))
+     write-with-meta)
 
    "clj/vector"
-   {:class clojure.lang.APersistentVector
-    :writer (reify WriteHandler
-              (write [_ w o]
-                (write-with-meta w "clj/vector" o)))
-    :readers {"clj/vector"
-              (reify ReadHandler
-                (read [_ rdr tag component-count]
-                  (read-with-meta rdr vec)))}}
+   (create-identity-based-handler
+     clojure.lang.APersistentVector
+     "clj/vector"
+     (fn clj-vec-reader [rdr] (read-with-meta rdr vec))
+     write-with-meta)
 
    "clj/list"
-   {:class clojure.lang.PersistentList
-    :writer (reify WriteHandler
-              (write [_ w o]
-                (write-with-meta w "clj/list" o)))
-    :readers {"clj/list"
-              (reify ReadHandler
-                (read [_ rdr tag component-count]
-                  (read-with-meta rdr #(apply list %))))}}
+   (create-identity-based-handler
+     clojure.lang.PersistentList
+     "clj/list"
+     (fn clj-list-reader [rdr] (read-with-meta rdr #(apply list %)))
+     write-with-meta)
 
    "clj/emptylist"
+   ;; Not using the identity based handler as this will always be identical anyway
+   ;; then meta data will be appended in the reader
    {:class clojure.lang.PersistentList$EmptyList
     :writer (reify WriteHandler
               (write [_ w o]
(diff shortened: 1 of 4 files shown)


src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -191,6 +191,40 @@
                       (read-record add-node-expr-fn)
                       d/cache-node)))}}))
 
+(defn- create-identity-based-handler
+  [clazz
+   tag
+   read-fn
+   write-fn]
+  (let [indexed-tag (str tag "idx")]
+    ;; Write an object a single time per object reference to that record.  The object is then "cached"
```
mrrodriguez: "to that record" should be "to that object" here.

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -191,6 +191,40 @@
                       (read-record add-node-expr-fn)
                       d/cache-node)))}}))
 
+(defn- create-identity-based-handler
+  [clazz
+   tag
+   read-fn
+   write-fn]
+  (let [indexed-tag (str tag "idx")]
```
mrrodriguez: "idx" will be against the tag str, which may be weird to read. Perhaps "-idx" instead?

src/main/clojure/clara/rules/durability/fressian.clj
EthanEChristian: +1, https://github.com/cerner/clara-rules/pull/420/commits/740953759448fbdfafc0a06f0f27e225235bfae0

src/main/clojure/clara/rules/durability/fressian.clj
EthanEChristian: +1, https://github.com/cerner/clara-rules/pull/420/commits/740953759448fbdfafc0a06f0f27e225235bfae0

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -191,6 +191,40 @@
...
+    {:class clazz
+     :writer (reify WriteHandler
+               (write [_ w o]
+                 (if-let [idx (d/clj-struct->idx o)]
+                   (do
+                     (.writeTag w indexed-tag 1)
+                     (.writeInt w idx))
+                   (do
+                     (write-fn w tag o)
+                     (d/clj-struct-holder-add-fact-idx! o)))))
```
WilliamParker: I think it is important here that the object is added to the cache _after_ it has been written (and implicitly after everything contained within it is written and cached) to match the ordering of the addition of items to the cache during read.  That is, we wouldn't want a write of 
````
[#{:test}]
````
To result in a cache of 
````
{[#{test}] 0]
  #{test} 1}
````
on write and a cache of 
````
[#{test} [#{test}]]
````
on read.

I ran the following test in my REPL:

````
(defn serde1 [x]
  (with-open [os (java.io.ByteArrayOutputStream.)
              ^FressianWriter wtr (fres/create-writer os :handlers df/write-handler-lookup)]
    ;; Write
    (pform/thread-local-binding [d/node-id->node-cache (volatile! {})
                                 d/clj-struct-holder (java.util.IdentityHashMap.)]
                                (fres/write-object wtr x)
                                (println "Write: " (.get d/clj-struct-holder))

                                )
    ;; Read
    (let [data (.toByteArray os)]
      (pform/thread-local-binding [d/clj-struct-holder (java.util.ArrayList.)]
                                  (with-open [is (java.io.ByteArrayInputStream. data)
                                              ^FressianReader rdr (fres/create-reader is :handlers df/read-handler-lookup)]
                                    (fres/read-object rdr)
                                    (println "Read" (.get d/clj-struct-holder)))))))
````
````
clara.test-fressian> (serde1 [[:a]])
Write:  #object[java.util.IdentityHashMap 0x71cdad81 {[[:a]]=1, [:a]=0}]
Read #object[java.util.ArrayList 0x36758cef [[:a], [[:a]]]]
````

This isn't a new issue, but it is a bigger one if we're doing caching on all Clojure data structures.  I suggest adding a comment here.  Thoughts/disagreements @mrrodriguez ?

File: src/test/clojure/clara/test_fressian.clj [hunk 4]
```diff
@@ -108,3 +108,14 @@
                      (serde (with-meta sm-custom {})))
             "cannot serialized custom sort comparators without name given in metadata")))))
 
+(deftest test-handler-identity
+  (let [v [1 2 3]
+        l (list 4 5 6)
+        m {:a 1 :b 2}
+        s #{:a :b :c}
```
WilliamParker: This seems to be missing some types, LazySeq for example.  I think it would be easy to add.
For example:
````
(let [ls (map inc [1 2]) ....
````

File: src/main/clojure/clara/rules/durability/fressian.clj [hunk 5]
```diff
@@ -191,6 +191,40 @@
                       (read-record add-node-expr-fn)
                       d/cache-node)))}}))
 
+(defn- create-identity-based-handler
+  [clazz
+   tag
+   read-fn
```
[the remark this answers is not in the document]
mrrodriguez: I think whatever we choose, we should be consistent across function signatures to avoid confusion.

Write before read sort of makes sense since writing does have to happen before read - technically.

src/main/clojure/clara/rules/durability/fressian.clj
mrrodriguez: @WilliamParker I believe you are wanting just to make it clearer in the docs/comments that the write traversal is carefully chosen so that it can be mirrored by the read traversal right?

If so, yeah, I think it should be explained better to avoid bad implementations being written. I also don't think that is something unique to this PR though. It's really just a call for better documentation around the way this works in general right?

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -206,36 +240,29 @@
...
-                  (read-with-meta rdr #(apply list %))))}}
+   (create-identity-based-handler
+     clojure.lang.PersistentList
+     "clj/list"
+     (fn clj-list-reader [rdr] (read-with-meta rdr #(apply list %)))
+     write-with-meta)
 
    "clj/emptylist"
+   ;; Not using the identity based handler as this will always be identical anyway
+   ;; then meta data will be appended in the reader
```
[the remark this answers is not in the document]
mrrodriguez: That is reasonable. It is sort of difficult to say it precisely here. Maybe even "reassociated"

src/main/clojure/clara/rules/durability/fressian.clj
WilliamParker: Right, I don't think it needs to be a blocker for merging this.  It is just a bigger issue now that it isn't just Clojure records being cached.

src/main/clojure/clara/rules/durability/fressian.clj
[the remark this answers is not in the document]
EthanEChristian: reordered: 
https://github.com/EthanEChristian/clara-rules/commit/bc8ba8fe933c6fd25d5911b6e80401f2563f8112#diff-97779c182cfdc7f57412ed5c7fe3ff07L197

src/main/clojure/clara/rules/durability/fressian.clj
EthanEChristian: added a minor comment

src/test/clojure/clara/test_fressian.clj
EthanEChristian: added a few others as well

src/main/clojure/clara/rules/durability/fressian.clj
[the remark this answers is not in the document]
EthanEChristian: changed to "added"

WilliamParker:
+1 to merge with the latest changes in https://github.com/cerner/clara-rules/pull/420/commits/bc8ba8fe933c6fd25d5911b6e80401f2563f8112 @EthanEChristian

Issue #422, opened 2019-02-07
Expose flag to retain compile-ctx instead of maintaining it


EthanEChristian:
As part of #381, i added batching to the evaluation of forms generated by the rules. Prior to my changes the compile-ctx was only bound at the time of evaluation and was discarded as it was "no longer needed". 

Compile-ctx acts as a way of providing additional information when a portion of a rule fails to compile, for instance an alpha-expr would also provide the constraints that were included in the alpha-expr. 

With the changes in #381, instead of discarding the compile-ctx it was instead stored on the rulebase along with the un-eval'd forms for later use with durability. This allows for the compile-ctx to be used as additional data for exceptions in the event that the environment where the rules are being deserialized doesn't maintain a state consistent to that of the pre-serialization one, ex. rules containing  symbols being able to be resolved pre-serialization but not post.

At the time, I didn't give much thought to the addition of the compile-ctx but with #419 i noticed that it might be holding unnecessary memory for consumers that will not be using durability or consumers that can provide an identical deserialization environment for durability.

The idea with this issue would be to expose a flag at session construction time that would opt into maintaining this data post rulebase construction.

EthanEChristian:
Now that #423 is merged, I am going to close this issue

PR #423, opened 2019-02-07
#422: Expose flag to retain compile-ctx instead of maintaining it

EthanEChristian:
Using #420 's results as a baseline, meaning that these numbers include those changes as well, and using the same session. I was able to get a few heap dumps of the session without the `:compile-ctx` being stored.

Pre-serialization:
clara_pre-serde_sans_cc
Post-serialization:
clara_post-serde_sans_cc

The thing to note here is that the pre-serialization, from my knowledge the standard use-case of clara, has dropped from ~523.6MB to ~504.0MB (again this is a Session containing 31k rules, so milage may vary).

The disk size of the Session also dropped from 62.4MB to 59.1MB.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 This is a history of changes to clara-rules.
 
+# 0.20.0 SNAPSHOT
+* Add flag to omit compilation context after Session compilation for use in serialization and deserialization. [issue 422](https://github.com/cerner/clara-rules/issues/422)
+
 # 0.19.1
 * Added a new field to the clara.rules.engine/Accumulator record.  This could be a breaking change for any user durability implementations with low-level performance optimizations.  See [PR 410](https://github.com/cerner/clara-rules/pull/410) for details.
 * Performance improvements for :exists conditions.  See [issue 298](https://github.com/cerner/clara-rules/issues/298).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 This is a history of changes to clara-rules.
 
 # 0.20.0 SNAPSHOT
-* Add flag to omit compilation context after Session compilation for use in serialization and deserialization. [issue 422](https://github.com/cerner/clara-rules/issues/422)
+* Add a flag to omit compilation context (used by the durability layer) after Session compilation to save space when not needed. Defaults to true. [issue 422](https://github.com/cerner/clara-rules/issues/422)
 
 # 0.19.1
 * Added a new field to the clara.rules.engine/Accumulator record.  This could be a breaking change for any user durability implementations with low-level performance optimizations.  See [PR 410](https://github.com/cerner/clara-rules/pull/410) for details.
--- a/src/main/clojure/clara/rules.cljc
+++ b/src/main/clojure/clara/rules.cljc
@@ -332,6 +332,8 @@
       * :forms-per-eval - The maximum number of expressions that will be evaluated per call to eval.
         Larger batch sizes should see better performance compared to smaller batch sizes. (Only applicable to Clojure)
         Defaults to 5000, see clara.rules.compiler/forms-per-eval-default for more information.
+      * :retain-compile-ctx - Retain compilation information for better compilation failures.
+        Defaults to false, see clara.rules.compiler/retain-compile-ctx-default for more information.
 
       This is not supported in ClojureScript, since it requires eval to dynamically build a session. ClojureScript
       users must use pre-defined rule sessions using defsession."
--- a/src/main/clojure/clara/rules.cljc
+++ b/src/main/clojure/clara/rules.cljc
@@ -332,8 +332,12 @@
       * :forms-per-eval - The maximum number of expressions that will be evaluated per call to eval.
         Larger batch sizes should see better performance compared to smaller batch sizes. (Only applicable to Clojure)
         Defaults to 5000, see clara.rules.compiler/forms-per-eval-default for more information.
-      * :retain-compile-ctx - Retain compilation information for better compilation failures.
-        Defaults to false, see clara.rules.compiler/retain-compile-ctx-default for more information.
+      * :omit-compile-ctx - When false Clara, in Clojure, retains additional information to improve error messages during
+        session deserialization at the cost of additional memory use.
+        By default this information is retained till the session is initially compiled and then will be discarded,
+        in the event that the Session will be serialized and deserialized (Durability) this information could be useful
+        if the session could not be compiled properly.
+        Defaults to true, see clara.rules.compiler/omit-compile-ctx-default for more information.
 
       This is not supported in ClojureScript, since it requires eval to dynamically build a session. ClojureScript
       users must use pre-defined rule sessions using defsession."
--- a/src/main/clojure/clara/rules.cljc
+++ b/src/main/clojure/clara/rules.cljc
@@ -334,9 +334,9 @@
         Defaults to 5000, see clara.rules.compiler/forms-per-eval-default for more information.
       * :omit-compile-ctx - When false Clara, in Clojure, retains additional information to improve error messages during
         session deserialization at the cost of additional memory use.
-        By default this information is retained till the session is initially compiled and then will be discarded,
-        in the event that the Session will be serialized and deserialized (Durability) this information could be useful
-        if the session could not be compiled properly.
+        By default this information is retained until the session is initially compiled and then will be discarded. This
+        information might prove useful for debugging compilation errors within the rulebase, eg. rulebase serialization
+        (ie. via Clara's durability support).
         Defaults to true, see clara.rules.compiler/omit-compile-ctx-default for more information.
 
       This is not supported in ClojureScript, since it requires eval to dynamically build a session. ClojureScript
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1966,8 +1966,11 @@
         retain-compile-ctx (:retain-compile-ctx options retain-compile-ctx-default)
         exprs (if retain-compile-ctx
                 exprs
-                (zipmap (keys exprs)
-                        (mapv (juxt first #(dissoc (second %) :compile-ctx)) (vals exprs))))
+                (into {}
+                      (map
+                        (fn [[k [expr ctx]]]
+                          [k [expr (dissoc ctx :compile-ctx)]]))
+                      exprs))
 
         beta-tree (compile-beta-graph beta-graph exprs)
         beta-root-ids (-> beta-graph :forward-edges (get 0)) ; 0 is the id of the virtual root node.
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1964,13 +1964,13 @@
         ;; uncertain deserialization environment and this sort of troubleshooting information would be useful
         ;; in diagnosing compilation errors in specific rules.
         omit-compile-ctx (:omit-compile-ctx options omit-compile-ctx-default)
-        exprs (if-not omit-compile-ctx
-                exprs
+        exprs (if omit-compile-ctx
                 (into {}
                       (map
                         (fn [[k [expr ctx]]]
                           [k [expr (dissoc ctx :compile-ctx)]]))
-                      exprs))
+                      exprs)
+                exprs)
 
         beta-tree (compile-beta-graph beta-graph exprs)
         beta-root-ids (-> beta-graph :forward-edges (get 0)) ; 0 is the id of the virtual root node.
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1917,14 +1917,14 @@
    This sums to 65,527B just shy of the 65,536B method size limit."
   5000)
 
-(def retain-compile-ctx-default
+(def omit-compile-ctx-default
   "During construction of the Session there is data maintained such that if the underlying expressions fail to compile
-   then this data can be used to explain the failure and the constraints of the rule whos expression is being evaluated.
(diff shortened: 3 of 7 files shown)


src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1948,6 +1957,18 @@
...
+
+        ;; If we have made it to this, it means that we have succeeded in compiling all expressions
+        ;; thus we can free the :compile-ctx used for troubleshooting compilation failures.
+        ;; The reason that this flag exists is in the event that this session will be serialized with an
+        ;; uncertain deserialization environment and this sort of troubleshooting information would be useful
+        ;; in diagnosing compilation errors in specific rules.
+        retain-compile-ctx (:retain-compile-ctx options retain-compile-ctx-default)
+        exprs (if retain-compile-ctx
+                exprs
+                (zipmap (keys exprs)
```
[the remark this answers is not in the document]
EthanEChristian: +1, updated here https://github.com/cerner/clara-rules/pull/423/commits/33d66565ee629aca688355b35a6519f8051c089c

src/main/clojure/clara/rules.cljc
```diff
@@ -332,6 +332,8 @@
       * :forms-per-eval - The maximum number of expressions that will be evaluated per call to eval.
         Larger batch sizes should see better performance compared to smaller batch sizes. (Only applicable to Clojure)
         Defaults to 5000, see clara.rules.compiler/forms-per-eval-default for more information.
+      * :retain-compile-ctx - Retain compilation information for better compilation failures.
+        Defaults to false, see clara.rules.compiler/retain-compile-ctx-default for more information.
```
WilliamParker: The reason to retain the compile-ctx is for use in durability, correct? I suggest that the docs here should mention that.  My primary concern is to make it easy for users who don't need to be concerned with the setting to rapidly skip over it (and dedicate mental bandwidth to things more relevant to their use-case).

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1917,6 +1917,15 @@
...
   5000)
 
+(def retain-compile-ctx-default
+  "During construction of the Session there is data maintained such that if the underlying expressions fail to compile
+   then this data can be used to explain the failure and the constraints of the rule whos expression is being evaluated.
+   The default behavior will be to discard this data, as there will be no use unless the session will be serialized and
+   deserialized into a dissimilar environment, ie function or symbols might be unresolvable. In those sorts of scenarios
+   it would be possible to construct the original Session with the `retain-compile-ctx` flag set to true, then the compile
+   context should aid in debugging the compilation failure on deserialization."
+  false)
```
WilliamParker: My initial instinct here would actually be that the default should be true, and only users who need to be concerned with memory optimizations on huge rulesets would be concerned with this.  That's a debatable point though.

Also, would it be possible to include a smaller set of information rather than the entire contents?  Would that achieve your memory use goals without introducing the complexity of a new user-facing setting?  I see [here](https://github.com/cerner/clara-rules/blob/0.19.0/src/main/clojure/clara/rules/compiler.clj#L1262) that it looks like we're binding the production to the compile-ctx dynamic var, perhaps we could retain something smaller like the name instead of the entire structure?

src/main/clojure/clara/rules/schema.cljc
```diff
@@ -167,23 +167,37 @@
                         (every? nil? (map s/check items tuple-vals))))
                  "tuple"))
 
-(def NodeCompilationContext
+(def NodeComplilationValue
```
WilliamParker: Spelling "Complilation" => "Compilation"

src/main/clojure/clara/rules.cljc
```diff
@@ -332,6 +332,8 @@
...
 
       This is not supported in ClojureScript, since it requires eval to dynamically build a session. ClojureScript
       users must use pre-defined rule sessions using defsession."
```
WilliamParker: What happens when a session is deserialized without the compile-ctx and there is an error?  Does it revert gracefully to previous behavior?  This seems like something that might merit a test.

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: I felt that it should default to false as the enhancement to batched compilation #396 could actually be seen as a degradation of memory for the typical use-cases (assuming non-durability is typical). Before this information was discarded right after evaluation, this way it would be putting it back the way it was before without requiring memory conscious consumers to have to evaluate the addition of the flag. I'm ok with it either way.

I'm not to worried about bindings that would fall out of scope after the Session has been created, rather i am trying to avoid the session itself holding this compile-ctx after it has been compiled.

WilliamParker: Discussed offline some with @EthanEChristian - I think the primary problem here is that the idea of a "compile-ctx" is something of a Clara/Clojure implementation detail that we shouldn't force users to dedicate thought to.  If we rename this ":deserialization-dbg-mode?" though, or something similar that more directly corresponds to the decision users need to make and document it accordingly, that seems like a reasonable tradeoff to me.  We'd preferably also add a note in the exceptions thrown if the extra data isn't present reminding the user that this mode exists i.e. "An exception was thrown while deserialization the rulebase.  Note that extra information is available when the :deserialization-dbg-mode is enabled".

mrrodriguez: I'm not a fan of how specific `deserialization-dbg-mode?` is when the data stored/not-stored is more generic "compilation context".

I actually think the name chosen seems appropriate, ie `retain-compile-ctx`

The only odd part to me is that it isn't stated in reverse perhaps, ie. `omit-compile-ctx` (or as clj/core likes to say "elide"). This would default to true. I think "omit" seems to be more common direction when referring to compiler opts - sort of like some opts in cljs.

WilliamParker: My primary concern here isn't exactly the name in and of itself, but that Clara be accessible and as easy to comprehend as possible for users who aren't familiar with Clojure or its ecosystem.  So for example, ideally a user need not even be aware that Clara makes calls to clojure.core/eval behind the scenes, or even of clojure.core/eval's existence.  My concern is that the current way this option appears makes users think they may need to be concerned with it when that is not in fact the case.  Perhaps we could reword the session options docs to something like ""retain compile context: When true Clara in Clojure retains additional information to improve error messages during session deserialization at the cost of additional memory use.  ... [rest of docs]. " Basically I'd prefer to make it easy for users only skimming the docstrings to quickly determine what settings concern them.  I think we can retain the name if preferred while still accomplishing that goal.

EthanEChristian: Updated the message in the latest commit, also changed the flag to `omit` per a comment by @mrrodriguez

src/main/clojure/clara/rules.cljc
EthanEChristian: tried to make it clearer in the latest commit.

src/main/clojure/clara/rules.cljc
EthanEChristian: Added a test to the test-durability ns with examples of the message that should be displayed.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 This is a history of changes to clara-rules.
 
+# 0.20.0 SNAPSHOT
+* Add flag to omit compilation context after Session compilation for use in serialization and deserialization. [issue 422](https://github.com/cerner/clara-rules/issues/422)
```
mrrodriguez: I think this reads in a misleading way. The "omit compilation context" would be something you would not want to set to true for serde, since it'd be broken.
Maybe:

> Add a flag to omit compilation context (used by the durability layer) after Session compilation to save space when not needed.

src/main/clojure/clara/rules.cljc
```diff
@@ -332,6 +332,12 @@
       * :forms-per-eval - The maximum number of expressions that will be evaluated per call to eval.
         Larger batch sizes should see better performance compared to smaller batch sizes. (Only applicable to Clojure)
         Defaults to 5000, see clara.rules.compiler/forms-per-eval-default for more information.
+      * :omit-compile-ctx - When false Clara, in Clojure, retains additional information to improve error messages during
+        session deserialization at the cost of additional memory use.
+        By default this information is retained till the session is initially compiled and then will be discarded,
```
mrrodriguez: till > until ?

src/main/clojure/clara/rules.cljc
```diff
@@ -332,6 +332,12 @@
       * :forms-per-eval - The maximum number of expressions that will be evaluated per call to eval.
         Larger batch sizes should see better performance compared to smaller batch sizes. (Only applicable to Clojure)
         Defaults to 5000, see clara.rules.compiler/forms-per-eval-default for more information.
+      * :omit-compile-ctx - When false Clara, in Clojure, retains additional information to improve error messages during
```
mrrodriguez: This should say "When true" right?

Also, why does it say ", in Clojure,"? What's the relevance of that?

src/main/clojure/clara/rules.cljc
```diff
@@ -332,6 +332,12 @@
...
+        session deserialization at the cost of additional memory use.
```
mrrodriguez: I don't know if it has to be specifically stated for deserialization.
It retains compile context that can be useful for certain types of inspection. Deserialization error message reporting is one example.

src/main/clojure/clara/rules.cljc
```diff
@@ -332,6 +332,12 @@
...
+        By default this information is retained till the session is initially compiled and then will be discarded,
+        in the event that the Session will be serialized and deserialized (Durability) this information could be useful
```
mrrodriguez: I think this can just say "serialized", the "and deserialized" is not necessary in cases where you don't really need to distinguish the 2 sides of serialization.

Also, instead of "(Durability)", how about "(ie. via Clara's durability support)".

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1948,6 +1957,21 @@
...
         ;; This is a performance optimization, see Issue 381 for more information.
         exprs (compile-exprs (extract-exprs beta-graph alpha-graph) forms-per-eval)
+
+        ;; If we have made it to this point, it means that we have succeeded in compiling all expressions
+        ;; thus we can free the :compile-ctx used for troubleshooting compilation failures.
+        ;; The reason that this flag exists is in the event that this session will be serialized with an
+        ;; uncertain deserialization environment and this sort of troubleshooting information would be useful
+        ;; in diagnosing compilation errors in specific rules.
+        omit-compile-ctx (:omit-compile-ctx options omit-compile-ctx-default)
+        exprs (if-not omit-compile-ctx
```
mrrodriguez: nit: Why `if-not`, instead just flip the conditions with an `if`?

src/test/clojure/clara/test_durability.clj
```diff
@@ -379,3 +380,27 @@
...
+                             (println "here"))
+        without-compile-ctx (com/mk-session [[rule]])
+        with-compile-ctx (com/mk-session [[rule] :omit-compile-ctx false])]
+    (ns-unmap 'clara.test-durability 'some-var)
+    (try
+      (rb-serde without-compile-ctx nil)
+      (is false "Error not thrown when deserializing the rulebase without ctx")
+      (catch Exception e
+        (is (= (.getMessage e)
+               "Failed compiling.\n{:expr (clojure.core/fn [?__fact__ ?__env__] (clojure.core/let [this ?__fact__ ?__bindings__ (clojure.core/atom {})] (if (== this clara.test-durability/some-var) (clojure.core/deref ?__bindings__) nil)))}\n"))))
```
mrrodriguez: It's be nice to regex this perhaps instead of dump the literal output.
The thing is that this could easily change with future compiler work and having to fix things like this can be annoying.

WilliamParker: +1, this should be made less brittle somehow IMO.  I tend to prefer looking at the ex-info data map rather than the message where applicable and requiring that some Throwable in the stack match the predicate, not that the top one do so, in case the exception is later caught and re-thrown as the cause of something else.

CHANGELOG.md
mrrodriguez: You could conceivably like to have the `compile-ctx` for other types of debugging? I don't think the flag is directly coupled to serialization is my point. It's just one example (the main one) of where it can be used.

src/main/clojure/clara/rules.cljc
mrrodriguez: Ok, I guess it reads correctly after all. I guess I'm more used to flags described about what they do _when true_. So it's a bit weird that it starts of "when false", but it semantically is correct still.

src/main/clojure/clara/rules.cljc
mrrodriguez: Perhaps just say "eg. during deserialization" or as part of the doc, not making it sound like it is exclusively a durability flag (of course that's the only use _we_ have for it in the compiler right now).

WilliamParker:
@EthanEChristian the comment/doc updates look good.  The test is less brittle this way, but it still depends on the order of exceptions in the exception chain.  It is OK to merge this as is, but I've created a branch with what I was thinking of if that helps:  https://github.com/WilliamParker/clara-rules/commit/42e8d240489f8f5a79fbc99d5bcaaa05d0f42e5f Feel free to cherry-pick that in or use as an example here.

mrrodriguez:
I have no further requests from it.

WilliamParker:
Regarding Travis CI, I think it might be confused by having a commit that isn't the same as the author of the PR, but note that the Travis CI build on my branch succeeded: https://travis-ci.org/WilliamParker/clara-rules/builds/542879601

I suggest just merging this in, in any case Travis will build the squashed commit on master; as long as it builds before a release is done I think we're fine.

PR #425, opened 2019-03-11
Issue 275: Thow exception on infinite loops

WilliamParker:
…s not addressed yet.

This is still a fairly rough draft at this point.  Please don't merge it; I expect the durability tests to fail.  I wanted to give some visibility into my thinking here though.

Non-exhaustive lists of items for consideration:

-  The value of the :max-cycles setting is just a number, which means we could serialize it along with the session in durability.   Should we?  Pro:  it is one less thing for a user to think about upon deserialization.  Con: The Fressian implementation doesn't serialize other options that can't be reliably serialized.  Also this option doesn't impact the semantics of the ruleset, while other options do, so it arguably is different in kind.
- What should the default value be?

### DIFF
--- a/src/main/clojure/clara/rules.cljc
+++ b/src/main/clojure/clara/rules.cljc
@@ -35,6 +35,12 @@
 
    This take an additional map of options as a second argument.  Current options:
 
+   :max-cycles (positive integer): Each time Clara determines that there are no more insertions process in a given activation group (rules with the same salience)
+    it records that one cycle of processing rules has been performed.  When the max-cycles limit is reached Clara throws an exception indicating that the rules
+    are in an infinite loop.  This exception can be regarded as analogous to a StackOverflowError.  The default is 600000, which is a level at which it is unlikely that
+    most users will encounter this error in cases where the rules actually would eventually terminate.  See issue 275 for details on how this level was chosen.
+    Nevertheless, it is possible that such cases exist, and so an option is provided to increase the limit.  Furthermore, some infinite loop scenarios will cause memory
+    errors before this ceiling is reached, and so users who do not expect to need a large number of rule cycles may wish to lower this limit.
    :cancelling true (EXPERIMENTAL, subject to change/removal.  Not supported in ClojureScript.):  
     Simultaneously propagate insertions and retractions through the rules network, at every step using the insertion and retractions of equals facts to cancel each
     other out and avoid operations deeper in the rules network.  The behavior of unconditional insertions and RHS (right-hand side) retractions
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -1719,7 +1719,7 @@
 
 (defn fire-rules*
   "Fire rules for the given nodes."
-  [rulebase nodes transient-memory transport listener get-alphas-fn update-cache]
+  [rulebase nodes transient-memory transport listener get-alphas-fn update-cache max-cycles]
   (binding [*current-session* {:rulebase rulebase
                                :transient-memory transient-memory
                                :transport transport
@@ -1729,7 +1729,12 @@
                                :listener listener}]
 
     (loop [next-group (mem/next-activation-group transient-memory)
-           last-group nil]
+           last-group nil
+           flushed-updates-count 0]
+
+      (when (> flushed-updates-count max-cycles)
+        (throw (ex-info "It appears that the rules are in an infinite loop."  {:clara-rules/infinite-loop-suspected true})))
+                        
 
       (if next-group
 
@@ -1739,7 +1744,7 @@
           ;; group before continuing.
           (do
             (flush-updates *current-session*)
-            (recur (mem/next-activation-group transient-memory) next-group))
+            (recur (mem/next-activation-group transient-memory) next-group (inc flushed-updates-count)))
 
           (do
 
@@ -1817,15 +1822,15 @@
                   (when (some-> node :production :props :no-loop)
                     (flush-updates *current-session*)))))
 
-            (recur (mem/next-activation-group transient-memory) next-group)))
+            (recur (mem/next-activation-group transient-memory) next-group flushed-updates-count)))
 
         ;; There were no items to be activated, so flush any pending
         ;; updates and recur with a potential new activation group
         ;; since a flushed item may have triggered one.
         (when (flush-updates *current-session*)
-          (recur (mem/next-activation-group transient-memory) next-group))))))
+          (recur (mem/next-activation-group transient-memory) next-group (inc flushed-updates-count)))))))
 
-(deftype LocalSession [rulebase memory transport listener get-alphas-fn pending-operations]
+(deftype LocalSession [rulebase memory transport listener get-alphas-fn pending-operations max-cycles]
   ISession
   (insert [session facts]
 
@@ -1844,7 +1849,8 @@
                      transport
                      listener
                      get-alphas-fn
-                     new-pending-operations)))
+                     new-pending-operations
+                     max-cycles)))
 
   (retract [session facts]
 
@@ -1859,7 +1865,8 @@
                      transport
                      listener
                      get-alphas-fn
-                     new-pending-operations)))
+                     new-pending-operations
+                     max-cycles)))
 
   ;; Prior to issue 249 we only had a one-argument fire-rules method.  clara.rules/fire-rules will always call the two-argument method now
   ;; but we kept a one-argument version of the fire-rules in case anyone is calling the fire-rules protocol function or method on the session directly.
@@ -1912,7 +1919,8 @@
                        transport
                        transient-listener
                        get-alphas-fn
-                       (uc/get-ordered-update-cache)))
+                       (uc/get-ordered-update-cache)
+                       max-cycles))
 
         #?(:cljs (throw (ex-info "The :cancelling option is not supported in ClojureScript"
                                  {:session session :opts opts}))
@@ -1961,14 +1969,16 @@
                                  get-alphas-fn
                                  ;; This continues to use the cancelling cache after the first batch of insertions and retractions.
                                  ;; If this is suboptimal for some workflows we can revisit this.
-                                 update-cache)))))
+                                 update-cache
+                                 max-cycles)))))
 
       (LocalSession. rulebase
                      (mem/to-persistent! transient-memory)
                      transport
                      (l/to-persistent! transient-listener)
                      get-alphas-fn
-                     [])))
+                     []
+                     max-cycles)))
 
   (query [session query params]
     (let [query-node (get-in rulebase [:query-nodes query])]
@@ -1994,7 +2004,8 @@
      :listeners (if (l/null-listener? listener)
                   []
                   (l/get-children listener))
-     :get-alphas-fn get-alphas-fn}))
+     :get-alphas-fn get-alphas-fn
+     :max-cycles max-cycles}))
 
 (defn assemble
   "Assembles a session from the given components, which must be a map
@@ -2006,7 +2017,7 @@
    :listeners A vector of listeners implementing the clara.rules.listener/IPersistentListener protocol
    :get-alphas-fn The function used to return the alpha nodes for a fact of the given type."
 
-  [{:keys [rulebase memory transport listeners get-alphas-fn]}]
+  [{:keys [rulebase memory transport listeners get-alphas-fn max-cycles]}]
   (LocalSession. rulebase
                  memory
                  transport
@@ -2014,7 +2025,8 @@
                    (l/delegating-listener listeners)
                    l/default-listener)
                  get-alphas-fn
-                 []))
+                 []
+                 max-cycles))
 
 (defn with-listener
   "Return a new session with the listener added to the provided session,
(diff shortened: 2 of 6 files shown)


WilliamParker:
After some reflection I think the :max-cycles setting actually makes more sense as an option to fire-rules.  We already added options to fire-rules for [issue 249](https://github.com/cerner/clara-rules/issues/249).  The maximum cycles that are plausible is obviously partly dependent on the rules in question, but the specific input data and environment where the rules are run (i.e. prod, dev, or anything that impacts the cost of a false positive versus a false negative in detection of infinite cycles) are also important.  If users want to have a session always have the same max-cycles option for all fire-rules calls it should be readily possible for them to wrap the fire-rules function to always have the option given.

WilliamParker:
As usual I'll want to take a second look with fresh eyes before merging, but this can be reviewed as a candidate PR for merge now. @mrrodriguez @EthanEChristian

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1729,7 +1729,25 @@
...
+                     " If the rules are not in fact in a loop it is possible to increase the ceiling by use of the :max-cycles setting."
+                     " See the clara.rules/fire-rules docstring for details.")
+                {:clara-rules/infinite-loop-suspected true
+                 :listeners (try
+                              (let [p-listener (l/to-persistent! listener)]
+                                (if (l/null-listener? p-listener)
+                                  []
+                                  (l/get-children p-listener)))
+                              (catch #?(:clj Exception :cljs :default)
+                                  listener-exception
```
mrrodriguez: Indentation makes this confusing.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1817,13 +1835,13 @@
                   (when (some-> node :production :props :no-loop)
                     (flush-updates *current-session*)))))
 
-            (recur (mem/next-activation-group transient-memory) next-group)))
+            (recur (mem/next-activation-group transient-memory) next-group flushed-updates-count)))
```
mrrodriguez: Why no `inc` this time?'
Should it be set back to 0?

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1867,7 +1885,8 @@
   (fire-rules [session opts]
 
     (let [transient-memory (mem/to-transient memory)
-          transient-listener (l/to-transient listener)]
+          transient-listener (l/to-transient listener)
+          max-cycles (get opts :max-cycles 600000)]
```
mrrodriguez: I'm not sure on the default chosen here. It could be breaking. 

I haven't thought it completely through, but is the counting no specifically counting cycles of rules happening, but instead just steps happening during fire-rules? Perhaps that is not granular enough.

src/main/clojure/clara/tools/testing_utils.cljc
```diff
@@ -152,3 +152,20 @@
...
+     [t]
+     (let [throwables  ((fn append-self
+                          [prior t1]
+                          (if t1
+                            (append-self (conj prior t1) (.getCause ^Throwable t1))
+                            prior))
+                        []
+                        t)]
+       (into []
+             (comp (map ex-data)
```
mrrodriguez: I think you can just use `(keep ex-data)` here and remove the `comp` and `filter` line.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: My reasoning here was that this path is reached every time a :no-loop rule is fired, while the other paths where I incremented the count are only reached upon the end of an activation group.  I expect activation group switches to be much less numerous than individual rules firing, and I thought the risk of throwing in ultimately terminating cases here to be too high.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: How would you indent it?  This looks fairly normal to me but open to suggestions.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: The count is intended to be incremented each time an activation group is fully fired.  I'm expecting that to be much smaller than the number of times a rule is fired, and so safer in terms of avoiding false positives and incorrectly throwing.

WilliamParker: I'll have to post more details later on how I chose the count and maybe change it - I measured the count for 30 seconds of a simple case on my device machine.  Open to discussion on what the count should be.

src/main/clojure/clara/tools/testing_utils.cljc
WilliamParker: I wanted to filter out empty but non-nil exception data too, but I agree that using keep can simplify this a bit.

src/main/clojure/clara/rules/engine.cljc
EthanEChristian:
```
(catch #?(:clj Exception :cljs :default)  listener-exception
     listener-exception)
```
is probably how i would write it, but that just because i am used to seeing:
```
(try 
  <some logic>
  (catch Exception ex
    <do something>))
```
though the reader conditional makes this a bit different.

src/main/clojure/clara/rules/engine.cljc
EthanEChristian: I think I agree with Mike here:
> I'm not sure on the default chosen here. It could be breaking.

Given the fabricated set of rules
```
(ns clara.loop-detection
  (:require [clojure.test :refer :all]
            [clara.rules :as r]
            [clara.rules.accumulators :as acc]))

(defrecord Total [x])

(defrecord Subs [y])
(defrecord Sum [z])

(r/defrule sum-rule
  [Total (= ?total x)]
  [?s <- (acc/sum :y) :from [Subs]]
  [:test (> ?total ?s)]
  =>
  (r/insert! (->Sum (- ?total ?s))))

(r/defrule subs-rule
  [Sum]
  =>
  (r/insert-unconditional! (->Subs 1)))

(r/defquery sum-query
  []
  [?s <- (acc/count) :from [Subs]])

(-> (r/mk-session)
    (r/insert (->Total 600000))
    (r/fire-rules)
    (r/query sum-query))
```
This will break due to loop detection, when in fact it is a finite execution. While I won't expect to find this exact pattern in rules that we(cerner) have, I wouldn't be surprised if someone was doing something similar or using unconditional inserts to break loops.

On one hand, i would think that the addition of this would maintain the previous functionality of the rules engine by default(loop forever, or till done), but on the other I understand that someone just starting clara might not be aware of the functionality thus miss out on the detection option. 

Perhaps a large default is correct, but also allowing a sentinel value to disable loop detection to maintain passivity.

WilliamParker: My reasoning here was that it just didn't seem likely to hit it, but I see your points that it is always going to be possible and there is enough usage of Clara for odd use cases like this to exist.

How would we feel about something like the following?

- Having the limit be configurable, as is.  We could revisit the limit as well.
- Making the behaviour on reaching the limit configurable, with the options being to throw an exception, log a warning, or do nothing, with the log option being the default.  If a warning just went to standard out that would probably take care of most of the confusion on the part of users who misunderstand looping behaviour in Clara, while being less likely to cause problems for users than simply throwing an exception.  That said, the problems wouldn't go down to zero since if it bubbled up to the end-user (not a developer) in an app of some kind it could be confusing, as well as if Clara was in an app that had specific requirements around printing to standard out.
- In the future possibly adding a comprehensive "development mode" that would do things like enable throwing here, enable tracing, etc.  This would be an easier thing to explain in the docs than lots of different options requiring an understanding of Clara to make good choices.

If we go down the logging path, [tools.logging](https://github.com/clojure/tools.logging) might be an option.  I'm not delighted about adding dependencies, but it looks pretty lightweight and has no dependencies itself other than Clojure.  In essence it inspects the environment at runtime and uses whatever JVM logging framework is available, falling back to [java.util.logging])https://docs.oracle.com/javase/7/docs/api/java/util/logging/package-summary.html) if nothing else is available.  In ClojureScript I think we'd probably need to just print to the console, but I don't think the JS console is exposed to website users who aren't using developer tools so perhaps that isn't a major issue.

Sorry to take so long to respond here... I've been a bit preoccupied lately.

Thoughts @EthanEChristian @mrrodriguez ?

src/main/clojure/clara/rules.cljc
```diff
@@ -35,6 +35,12 @@
 
    This take an additional map of options as a second argument.  Current options:
 
+   :max-cycles (positive integer): Each time Clara determines that there are no more insertions process in a given activation group (rules with the same salience)
+    it records that one cycle of processing rules has been performed.  When the max-cycles limit is reached Clara throws an exception indicating that the rules
+    are in an infinite loop.  This exception can be regarded as analogous to a StackOverflowError.  The default is 600000, which is a level at which it is unlikely that
```
mrrodriguez: It really is indicating that the rules _may_ be in an infinite loop. I think that it'd be better stated that way.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: I think it was weird to see the first `listener-exception` on the next line. It just makes it blend with the body.
I don't consider it to be a critical point though if others thing it is fine how it is. I know reader-conditionals can make things get a bit weird.

src/main/clojure/clara/rules/engine.cljc
mrrodriguez: Ok, that explanation makes sense. However, you may want to note that in a comment above this since I think it could be hard to understand looking at it later.

src/main/clojure/clara/rules/engine.cljc
EthanEChristian: I think that approach seems reasonable to me, particularly:
> Making the behavior on reaching the limit configurable

I feel like this allows us to have this functionality and still maintain passivity for consumers that have "odd" rules.

I don't think I have any objections to the addition of logging to clara's dependencies.

mrrodriguez: @WilliamParker I like the default to be logging, however I don't think this should require a new dependency. 

I think it may actually be best to expose 2 args for this feature:

`max-cycles` - what you already have
`max-cycles-exceeded-fn` - tentative name, but it gives the consumer the ability to decide what to do

If they do not provide anything, perhaps we do a "print once" default fn. We can also provide a fn that can be used to throw the exception if that is what someone wants. These can be exposed as plain fn's or "special" keywords the argument accepts.

```clj
:max-cycles 60000
:max-cycles-exceeded-fn r/max-cycles-exceeded-print-once
;; Alternatives:
;; Special kw :clara-rules/max-cycles-exceeded-print-once
;; r/max-cycles-exceeded-throw
```
If no `:max-cycles-exceeded-fn` is given, use a default.

If we go this path, we should consider what args to this fn would be useful.

I think giving a handler to the consumer for cases like this is a common clj pattern.

@WilliamParker @EthanEChristian

EthanEChristian: Per an offline conversation with @mrrodriguez, for my use-case the addition of the dependency wouldn't be an issue. After that conversation, I think that i can agree with Mike's hesitance on clara mandating a logger on it's consumers.

MrMikeRodriguez: Thanks for summarizing @EthanEChristian 

I'll re-iterate my concern is that I'm not a particular fan of adding dependencies in general in a general-use library - unless they are providing enough value to justify. Logging in a single case doesn't seem too compelling in that area.

Also, I know that `org.clojure/tools.logging` is quite flexible (I use it fairly often), but it can still lead to issues when added to any given classpath. It does do some implicit logger impl searching/loading and I think it could get particularly awkward if someone was using a logger framework that doesn't fit into `org.clojure/tools.logging`'s pattern at all. It may end up logging to it's own logger instead of their chosen one/may be annoying to try to silence or integrate with.

mrrodriguez: Thanks for summarizing @EthanEChristian 

I'll re-iterate my concern is that I'm not a particular fan of adding dependencies in general in a general-use library - unless they are providing enough value to justify. Logging in a single case doesn't seem too compelling in that area.

Also, I know that `org.clojure/tools.logging` is quite flexible (I use it fairly often), but it can still lead to issues when added to any given classpath. It does do some implicit logger impl searching/loading and I think it could get particularly awkward if someone was using a logger framework that doesn't fit into `org.clojure/tools.logging`'s pattern at all. It may end up logging to it's own logger instead of their chosen one/may be annoying to try to silence or integrate with.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: Is this easier to understand now @mrrodriguez ? https://github.com/cerner/clara-rules/pull/425/commits/13ef0dfde268bd618092d34eb64f25edd2f00be8

WilliamParker:
Travis seems to be confused because I pushed the branch to cerner/clara-rules first, realized my mistake, then pushed it to my fork instead, but failed to build the branch that I deleted.  A build on the PR did succeed though.

src/main/clojure/clara/rules/engine.cljc
WilliamParker: OK it sounds like there is a consensus for a default behaviour of printing to standard out then.  I'll proceed with that.

WilliamParker:
This has been superseded by a listener-based approach in https://github.com/cerner/clara-rules/pull/447 

Closing without merge.

Issue #426, opened 2019-04-15
Join binding followed by usage of binding leads to unresolvable symbol at compilation


EthanEChristian:
Given the following rule:
```
(defrecord A [a])
(defrecord B [b])

(defrule a-rule 
  [A (= ?a a)]
  [B (= ?r (conj b ?a))
     (not-empty? ?r)]
  =>
  (println "doesn't matter"))
```
This rule will fail at compilation:
```
Caused by: java.lang.RuntimeException: Unable to resolve symbol: ?r in this context
```

The reason this occurs is because the way that the graph gets built. Currently `(= ?r (conj b ?a))` is identified as a `join-filter-expression` therefore it makes it into the ExpressionJoinNode. `(not-empty? ?r)` is not identified as a dependent on the binding above, so it is added as a constraint to the AlphaNode for `B`. Since the AlphaNode has no binding for `?r` it will fail to compile when trying to resolve the binding. 

The issue seems to stem from how we determine the [join-expression-filter](https://github.com/cerner/clara-rules/blob/0.19.1/src/main/clojure/clara/rules/compiler.clj#L889-L900).
[non-equality-unifications](https://github.com/cerner/clara-rules/blob/0.19.1/src/main/clojure/clara/rules/compiler.clj#L855-L862) classifies the variable usage at the beginning, however I think that it will also need to reclassify bound-variables if it encounters a binding that is a non-equality-unification.

EthanEChristian:
posted #427 as a fix

EthanEChristian:
During the implementation of the PR above, i ran across a scenario where the binding in question was not passed to the production node correctly on the CLJS side of clara. This issue will encompass that issue as well, as they are too tightly coupled to not fix both.

PR #427, opened 2019-04-16
(#426) Join binding followed by usage of binding leads to unresolvabl…

EthanEChristian:
…e symbol at compilation

### DIFF
--- a/src/main/clojure/clara/macros.clj
+++ b/src/main/clojure/clara/macros.clj
@@ -174,8 +174,10 @@
                                                      ;; that a rule wouldn't, or at least shouldn't for
                                                      ;; clarity, start the names of other locals or vars
                                                      ;; with "?".
-                                                     (mapv (comp symbol name) all-bindings))]
-              (com/compile-action all-bindings
+                                                     (mapv (comp symbol name) (:bindings beta-node)))]
+              ;; using the :bindings defined on the beta-node rather than `all-bindings`, as all-bindings
+              ;; does not account for bindings that reside within expression join nodes
+              (com/compile-action (:bindings beta-node)
                                   ;; Using private function for now as a workaround.
                                   (if (:ns-name production)
                                     (if (com/compiling-cljs?)
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -9,7 +9,6 @@
             [clojure.core.reducers :as r]
             [clojure.reflect :as reflect]
             [clojure.set :as s]
-            [clojure.set :as set]
             [clojure.string :as string]
             [clojure.walk :as walk]
             [schema.core :as sc]
@@ -803,8 +802,8 @@
 
         ;; Unsatisfied conditions remain, so find ones we can satisfy.
         (let [satisfied? (fn [classified-condition]
-                           (clojure.set/subset? (:unbound classified-condition)
-                                                bound-variables))
+                           (s/subset? (:unbound classified-condition)
+                                      bound-variables))
 
               ;; Find non-accumulator conditions that are satisfied. We defer
               ;; accumulators until later in the rete network because they
@@ -812,8 +811,8 @@
               ;; in the network are satisfied.
               satisfied-non-accum? (fn [classified-condition]
                                      (and (not (:is-accumulator classified-condition))
-                                          (clojure.set/subset? (:unbound classified-condition)
-                                                               bound-variables)))
+                                          (s/subset? (:unbound classified-condition)
+                                                     bound-variables)))
 
               has-satisfied-non-accum (some satisfied-non-accum? remaining-conditions)
 
@@ -825,16 +824,16 @@
                                   (remove satisfied-non-accum? remaining-conditions)
                                   (remove satisfied? remaining-conditions))
 
-              updated-bindings (apply clojure.set/union bound-variables
+              updated-bindings (apply s/union bound-variables
                                       (map :bound newly-satisfied))]
 
           ;; If no existing variables can be satisfied then the production is invalid.
           (when (empty? newly-satisfied)
 
             ;; Get the subset of variables that cannot be satisfied.
-            (let [unsatisfiable (clojure.set/difference
-                                 (apply clojure.set/union (map :unbound still-unsatisfied))
-                                 bound-variables)]
+            (let [unsatisfiable (s/difference
+                                  (apply s/union (map :unbound still-unsatisfied))
+                                  bound-variables)]
               (throw (ex-info (str "Using variable that is not previously bound. This can happen "
                                    "when an expression uses a previously unbound variable, "
                                    "or if a variable is referenced in a nested part of a parent "
@@ -855,11 +854,30 @@
 (defn- non-equality-unifications
   "Returns a set of unifications that do not use equality-based checks."
   [constraints]
-  (let [[bound-variables unbound-variables] (classify-variables constraints)]
-    (into #{}
-          (for [constraint constraints
-                :when (non-equality-unification? constraint bound-variables)]
-            constraint))))
+  (let [[bound-variables _] (classify-variables constraints)]
+    (loop [[cur & more] constraints
+           constraints #{}
+           bound-variables bound-variables]
+      (if cur
+        (let [non-equity? (non-equality-unification? cur bound-variables)
+
+              ;; In the event that the unification is also a binding, then we will need to remove
+              ;; the bound variable from the previously bound variables. This prevents further constraints
+              ;; that use this variable from being considered "equality-unifications".
+              ;; See https://github.com/cerner/clara-rules/issues/426 for more info.
+              bound-variables (if (and non-equity?
+                                       (seq cur)
+                                       (equality-expression? cur))
+                                (s/difference bound-variables
+                                              (into #{}
+                                                    (filter is-variable?)
+                                                    cur))
+                                bound-variables)
+              constraints (if non-equity?
+                            (conj constraints cur)
+                            constraints)]
+          (recur more constraints bound-variables))
+        constraints))))
 
 (sc/defn condition-to-node :- schema/ConditionNode
   "Converts a condition to a node structure."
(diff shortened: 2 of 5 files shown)


src/main/clojure/clara/rules/compiler.clj
```diff
@@ -9,7 +9,6 @@
             [clojure.core.reducers :as r]
             [clojure.reflect :as reflect]
             [clojure.set :as s]
-            [clojure.set :as set]
```
EthanEChristian: tried to clean up this binding

src/main/clojure/clara/macros.clj
```diff
@@ -174,8 +174,10 @@
                                                      ;; that a rule wouldn't, or at least shouldn't for
                                                      ;; clarity, start the names of other locals or vars
                                                      ;; with "?".
-                                                     (mapv (comp symbol name) all-bindings))]
-              (com/compile-action all-bindings
```
EthanEChristian: Looking at this again, i don't think `all-bindings` is required at all after these changes.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -855,11 +854,30 @@
...
+           constraints #{}
+           bound-variables bound-variables]
+      (if cur
+        (let [non-equity? (non-equality-unification? cur bound-variables)
+
+              ;; In the event that the unification is also a binding, then we will need to remove
+              ;; the bound variable from the previously bound variables. This prevents further constraints
+              ;; that use this variable from being considered "equality-unifications".
+              ;; See https://github.com/cerner/clara-rules/issues/426 for more info.
+              bound-variables (if (and non-equity?
```
mrrodriguez: I struggled with this a bit. I will summarize my understanding for clarity.

The issue #426  description had an example with this form of 
```clj
  [A (= ?a a)]
  [B (= ?r (conj b ?a))
     (not-empty? ?r)]
```
Now consider the constraints for `B` fact condition:
```clj
(= ?r (conj b ?a))
(not-empty? ?r)
```
What is happening is that the `(= ?r (conj b ?a))` is true for `(non-equality-unification? cur bound-variables)` because it has the sub-expression `(conj b ?a)` which is unifying on an unknown to this condition variable, in a way that is not using `=` (which has a simplified/optimized/hash-based impl).

I think it is important to note that this particular fn , `non-equality-unifications` is used to find all constraints that will be separated away from the "simple" expressions and moved into a "join expression" that deals with the non-equality unification.

Now `(classify-variables constraints)` is classifying this new variable binding `?r` as a "bound variable" in `bound-variables`. However, it isn't necessarily "previously bound" if constraints are moved around during compilation. Subsequent constraints that _are not_ moved with the non-equality unifications, could be broken, since the classification isn't accurate anymore if they are not also moved.

Concretely, the next constraint is `(not-empty? ?r)` and `?r` is classified as previously bound now.
This means that:
```clj
(non-equality-unification? '(not-empty? ?r) #{?r}) 
;;= false
```

This would result in `(not-empty? ?r)` being "skipped" in this fn, and consequently not "moved" with the rest of the "join filter" expression constraints later.

This approach works around this, by doing what Ethan appropriately called "re-classification". When a non-equality unification is detected, it has to remove it's newly introduced bindings from the classification as a mechanism to make other dependent constraints move along with it.

I can follow this approach. However, it does seem like a sneaky way to utilize how bound variables are used to detect non-equality unifications.
The reason I find it "sneaky" is that the logic it is doing is somewhat assuming that the caller is planning to move constraints around. It has the semantics that if it returns a constraint, it needs to also return all the constraints that may depend on that constraint's new bindings.

-----

However, it seems that this may _not_ work for some cases still:

```clj
(defrecord A [a])
(defrecord B [b])

(defrule a-rule2
  [A (= ?a a)]
  [B (= ?r (conj b ?a))
     (= ?r #{:a :b})]
  =>
  (println "doesn't matter"))
```

ie. the 2nd constraint is an equality binding on `?r`.

This compiles, but it isn't producing a semantically correct network.

The equality unification on `?r` in `(= ?r #{:a :b})` isn't being considered dependent on the `?r` introduced by `(= ?r (conj b ?a))` that is going to be moved into the join filter later.
You can see this by running above with:

```clj
(-> (mk-session [a-rule2] :cache false)
    (insert (->A :c)
            (->B #{:b}))
    fire-rules)

;; Prints "doesn't matter"

```

I think the solution to this is in not trying to trying to manipulate only the classified `bound-variables` to find the dependent constraints. A more "brute-force" search may be needed. In an interation where this condition is detected of the form of `(= ?r (conj b ?a))`, I think all subsequently dependent constraints that refer to `?r` need to be added ot the `constraints` immediately and removed from `more` - in terms of the locals used in the changes to the `non-equality-unifications` fn.

WilliamParker: Regarding the a-rule2 case you describe above @mrrodriguez it looks to me like the problem is that the alpha node is creating the ?r binding, but the join filter is then rebinding it rather than testing for equality with the value created in the alpha network.

````
 [5 :alpha-expr]
 [#object[clara.test_rules$eval17454$fn__17457 0x2fe5411 "clara.test_rules$eval17454$fn__17457@2fe5411"]
  {:file "*cider-repl localhost*",
   :compile-ctx
   {:condition
    {:type clara.test_rules.B,
     :constraints [(= ?r #{:b :a})],
     :original-constraints [(= ?r (conj b ?a)) (= ?r #{:b :a})]},
    :env nil,
    :msg "compiling alpha node"},
   :alpha-expr
   (clojure.core/fn
    [?__fact__ ?__env__]
    (clojure.core/let
     [this ?__fact__ ?__bindings__ (clojure.core/atom {})]
     (clojure.core/let
      [?r #{:b :a}]
      (clojure.core/swap! ?__bindings__ clojure.core/assoc :?r ?r)
      @?__bindings__)))}],
 [2 :join-filter-expr]
[2 :join-filter-expr]
 [#object[clara.test_rules$eval17454$fn__17459 0x60678a61 "clara.test_rules$eval17454$fn__17459@60678a61"]
  {:compile-ctx
   {:condition
    {:type clara.test_rules.B,
     :constraints [(= ?r #{:b :a})],
     :original-constraints [(= ?r (conj b ?a)) (= ?r #{:b :a})]},
    :join-filter-expressions
    {:type clara.test_rules.B, :constraints [(= ?r (conj b ?a))]},
    :env nil,
    :msg "compiling expression join node"},
   :join-filter-expr
   (clojure.core/fn
    [?__token__ ?__fact__ ?__element-bindings__ ?__env__]
    (clojure.core/let
     [this
      ?__fact__
      b
      (.-b ?__fact__)
      ?a
      (clojure.core/-> ?__token__ :bindings :?a)
      ?r
      (get ?__element-bindings__ :?r)
      ?__bindings__
      (clojure.core/atom {})]
     (clojure.core/let
      [?r (conj b ?a)]
      (clojure.core/swap! ?__bindings__ clojure.core/assoc :?r ?r)
      @?__bindings__)))}]
````

[compile-constraints](https://github.com/cerner/clara-rules/blob/eb8cf88cbae6d4493948e87a57c393dadb130b89/src/main/clojure/clara/rules/compiler.clj#L246) takes a collection of variables that it should not rebind, but in [compile-join-filter this only includes variables from the ancestors](https://github.com/cerner/clara-rules/blob/eb8cf88cbae6d4493948e87a57c393dadb130b89/src/main/clojure/clara/rules/compiler.clj#L500).  At first glance the solution would seem to be to pass along the element-bindings passed to compile-join-filter to compile-constraints.  However, on closer examination, it doesn't appear to me that the element-bindings are necessarily correct here.  They are [populated from the :new-bindings of the beta network](https://github.com/cerner/clara-rules/blob/eb8cf88cbae6d4493948e87a57c393dadb130b89/src/main/clojure/clara/rules/compiler.clj#L1361), which are just the bindings in the condition that aren't from the ancestors as can be seen in their population [here](https://github.com/cerner/clara-rules/blob/eb8cf88cbae6d4493948e87a57c393dadb130b89/src/main/clojure/clara/rules/compiler.clj#L928).

In order to fix this problem I think we need to store (correct) information on what bindings are created in the alpha network and which are created in the beta network, rather than just new-bindings information.  

Regarding this immediate PR, my current thinking is that there shouldn't be any need to start from the viewpoint of all bindings which are created in the condition; we can start from a "blank slate" of no bindings and proceed from there.  We could then generate a list of bindings created in the alpha network and of bindings created in the beta network and use this to resolve this issue and the one described by @mrrodriguez above.  Intuitively, the current solution seems a bit like we have incorrect binding analysis and are "cleaning it up" after the fact, while I'd suggest the cleaner solution would be to correct the binding analysis upfront.  Note also that we do have the list of bindings from the ancestors available should it be needed, although at the moment I don't think we do.  I also haven't tried implementing this concretely so it's possible that I'm missing something.

This may also bring up issues of whether we want the constraints to be order-independent as well and figuring out what the current behaviour is.

File: src/test/common/clara/test_bindings.cljc [hunk 4]
```diff
@@ -369,3 +371,70 @@
...
+
+;; https://github.com/cerner/clara-rules/issues/426
+;; A join binding used in the same condition but different constraint will fail to compile.
+(defn calc-wind-chill
+  [temp wind-speed]
+  (+ 35.74
+     (- (* 0.6215 temp)
+        (* 35.75
+           (Math/pow wind-speed 0.16)))
+     (* 0.4275 temp (Math/pow wind-speed 0.16))))
```
WilliamParker: I'm guessing this is a formula for wind chill?  It seems unnecessarily complicated to use a real formula here since what we're testing has nothing to with this.  Also does "Math/pow" work in ClojureScript?

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -855,11 +854,30 @@
...
-    (into #{}
-          (for [constraint constraints
-                :when (non-equality-unification? constraint bound-variables)]
-            constraint))))
+  (let [[bound-variables _] (classify-variables constraints)]
+    (loop [[cur & more] constraints
+           constraints #{}
+           bound-variables bound-variables]
+      (if cur
+        (let [non-equity? (non-equality-unification? cur bound-variables)
```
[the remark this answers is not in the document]
WilliamParker: +1

src/main/clojure/clara/macros.clj
WilliamParker: If I understand this correctly all-bindings is being populated from the constraints after [this transformation](https://github.com/cerner/clara-rules/blob/0.19.1/src/main/clojure/clara/rules/compiler.clj#L906) and that is why this change is needed?

Issue #435, opened 2019-09-25
Requesting clarification re: multiple equal facts (I found this counterintuitive)


blak3mill3r:
I am just beginning to try to familiarize myself with this abstraction. I read all of the documentation, and I've just discovered that inserting a fact is not idempotent. My intuition was that it would be.

After re-reading the documentation and some of the issues history, I didn't find any mention of this specifically. #84 is related, but my question is simpler still. I'd appreciate some clarification on the reasoning behind storing duplicate facts (and allowing them transitively to produce multiple identical inferred facts).

I am not sure what it might mean to store and reason about two or more exactly identical "facts". They are all the same fact, no? If they are not the same fact, in spite of being equal, then how can I model that two equal facts do not carry any more information than just one of them?

My expectation was that I could model, for example, two separate observations of the same information as two records which differ only in an `observation-timestamp` field, and that if several different combinations of rules happen to imply exactly the same fact, that this would be equivalent in its downstream implications to any single piece of evidence that would imply the truth of that one fact. Instead it seems that there is a difference between the same fact being inserted once, twice, or n times. Why is that?

```clojure
(ns confused
  (:require [clara.rules :refer :all]))

(defrecord BottlesOfBeer [count time])

(defquery get-bottle-count [:?time]
  [?count <- BottlesOfBeer (= ?time time)])

(def ts (.getTime #inst "2020-01-01T00:00:00.000-00:00"))

(-> (mk-session)
    (insert (->BottlesOfBeer 99 ts))
    (insert (->BottlesOfBeer 99 ts))
    (fire-rules)
    (query get-bottle-count :?time ts))
;; =>
;; ({:?count #confused.BottlesOfBeer{:count 99, :time 1577836800000}, :?time 1577836800000},
;;  {:?count #confused.BottlesOfBeer{:count 99, :time 1577836800000}, :?time 1577836800000})


(let [[f1 f2] (-> (mk-session)
                  (insert (->BottlesOfBeer 99 ts))
                  (insert (->BottlesOfBeer 99 ts))
                  (fire-rules)
                  (query get-bottle-count :?time ts))]
  (= f1 f2))
;; => true
```

Should I query the session first and only `insert` if the fact is not already present? Or, do I need to use an accumulator to `distinct` them before inferring anything further? The example above is simpler than the code I was actually working with (I was initially confused by the presence of duplicate inferred facts) but I would like first to understand what it means to have `f1` and `f2` returned by the `query` above.

[...]

PR #438, opened 2019-10-09
idempotent fact insertion proposal for discussion   (query -> conditional insert)

blak3mill3r:
an implementation of `idempotent-insert` related to #435

I'm not expecting this to be merged. It has no tests and doesn't support ClojureScript. This is more to start a discussion about the implementation. I do think this would be a useful feature for `clara.rules` to have, but I'm not sure that the implementation in this PR is the right approach.

Perhaps a "mode" for the session would be better. I understand that clara.rules is not storing every single fact necessarily, but that what is stored depends on the rules a fact might match.

The code here works (and I am going to use something like this for now), however:

* it's clojure only
* must call `fire-rules`
* it will only work with records, so that it can conveniently create
  an "exact match" query which matches on every field of the record
* doesn't include an `insert-idempotent!` for use in RHS

### DIFF
--- a/src/main/clojure/clara/idempotent.clj
+++ b/src/main/clojure/clara/idempotent.clj
@@ -1,0 +1,32 @@
+(ns clara.idempotent
+  "Implementation of insert-idempotent and defrecord-idempotent, such that duplicate identical facts will not be inserted."
+  (:require [clara.rules :refer [fire-rules insert]]))
+
+(defn dynamic-basis [record-class]
+  (clojure.lang.Reflector/invokeStaticMethod (.getName record-class) "getBasis" (into-array [])))
+
+(defprotocol ExactQueryable
+  (present? [fact session]))
+
+(defmacro prepare-idempotency [klass-sym]
+  (let [klass (resolve klass-sym)
+        fields (dynamic-basis klass)
+        constraints (into [] (map #(-> `(~'= ~(symbol (str "?" %)) ~%)))     fields) 
+        args        (into [] (comp (map #(str "?" (name %))) (map keyword))  fields)
+        query-sym   (symbol (str "exact-match-"(.getSimpleName klass)))
+        match-expr  (into [] cat (for [f fields] [(keyword (str "?" f)) (list (keyword f) 'fact)]))]
+    `(do
+       (defquery ~query-sym ~args [~'?f ~'<- ~klass-sym ~@constraints])
+       (extend-type ~klass-sym ExactQueryable
+                    (~'present? [~'fact ~'session]
+                     (-> ~'session (query ~query-sym ~@match-expr)
+                         first :?f))))))
+
+(defn insert-idempotent [session fact]
+  (if (present? fact session) session
+      (-> session (insert fact) fire-rules)))
+
+(defmacro defrecord-idempotent [& args]
+  `(do
+     (defrecord ~@args)
+     (prepare-idempotency ~(first args))))


blak3mill3r:
TLDR; the code in this PR adds a new namespace which provides a macro

`defrecord-idempotent` which expands into:

* a `defrecord`
* a `defquery` that will match on every field of the record
* a protocol implementation of `(present? [fact session])`

So to use it, if you're already using records, you replace them with `defrecord-idempotent`, and then you can use `insert-idempotent` with them.

eraserhd:
Is the goal here the ability to have only one fact in the store (and triggered) when there is multiple supporting evidence?

If so, this only requires one retraction for two `insert-idempotent`s, which I think messes up the semantics you are going for.  Also, "insert insert retract insert" will trigger.  I think you would need to reference count the fact.

blak3mill3r:
@eraserhd The whole rules-engine abstraction is relatively new to me, and I came across it in a search for the right abstraction for a particular project I'm working on, which was evading me...

It's an X window manager that also manages processes and automates development tasks (or something along those lines). Maybe it's a process manager that automates window management. It's written in Clojure with an XMonad component that's dynamically recompiled/reloaded with some Clojure data slipped into that compilation.

The rules engine is intended to allow the process/window/resource automation to be user-extensible, and hopefully both composable and intuitive.

However the fact that `clara.rules` allows multiple identical facts turned out to be inconvenient for me as I started implementing it. Several pieces of evidence of the same fact produce multiple downstream facts, and while I could filter them out, or use accumulators for everything, something about that approach struck me as unclean (and a little inefficient).

I love that it's all immutable datastructures in there, and I do not want to pile up inferred facts for things that aren't relevant anymore (processes that have exited, windows that have closed, for example).

I am picturing an atom holding a rules session that persists for quite some time, the lifetime of the (window manager) session.

blak3mill3r:
Perhaps I should instead use the rules engine to determine what actions need to be taken short-term, and not expect a rules session to last as long as a WM session...

eraserhd:
Some rules engines have the semantics like so:

* If a fact is inserted, it is "true".
* If it is inserted twice, that means that it supported by multiple rules.  But it is no more or less true, so there is no need to tigger again.
* If it is then retracted twice, that means that all supporting evidence was lost, and the fact is no longer true.

For example, say I have a rule:

```clojure
(defrule socks-on-sale
  [:or [SockColor (= id ?id) (= color :red)]
       [SockMaterial (= id ?id) (= type :wool)]]
  =>
  (insert! (->SockSale ?id)))
```

In this case, I probably want semantics where red, wool socks are on sale.  But if I later spill blue dye on the socks, and they become purple then I want to:

```
(-> session
  (retract (->SockColor 42 :red))
  (insert (->SockColor 42 :purple)))
```

The socks should still be on sale (the kind of semantics some systems have).

blak3mill3r:
Thank you for that explanation, I will take some time to integrate it.

The example you gave makes good sense, that all supporting facts must be retracted for the supported fact to cease to be true.

However, what confused me was that inserting 5 duplicate copies of "the same fact" (per clojure's equality) required 5 retractions (and, space in memory and on disk).

My intuition, which turned out to be wrong, was that 5 different facts about the same thing that indicated the same thing would all have to be retracted (similar to the example you gave) but that these would not compare equal to each other. For example, 5 different observations of a patient's blood sugar, all happened to give the exact same blood sugar reading...

but they differed by timestamp, or by who administered the test

otherwise if they do not differ in either regard, how can we tell that it wasn't the same fact mistakenly entered a few times?

Or, perhaps due to network partitions and retry behavior?

That's another reason I think that idempotent insert semantics would be useful. They wouldn't prevent modeling the 5-blood-sugar-readings case, but they would allow the network-retries-dont-duplicate-facts.

If `clara.rules` lacking such a feature, if I wanted to use it inside a system built for redundancy and reliability that might replay messages if they _might_ have been missed, then I would have to protect the rules engine by, say, storing a set of all facts before giving them to `clara.rules`.

That might be the better approach, too, since it doesn't seem like the implementation of `clara.rules` has any straightforward way to test for the presence of a fact, and I think the code in this PR must add to what is stored, by adding those `exact-match-` rules.

mrrodriguez:
I haven't reviewed the code. I just want to point out a few things.

The popular rules engine Drools (Java-based now), has a concept of not duplicating "logically inserted facts". It comes with it's own issues in both performance and semantics when it isn't what you want, which is what was mentioned about how Clara just defaults to not doing any of this for you.

However, if it were to be done, I am much more in favor of it being a mode activated during the creation of a session (and the working memory that goes with it, etc).
I think there is more to this than just how facts can be externally inserted into the session as well.

You could have a rule like:
```clj
(defrule my-rule
 [A]
 => 
 (insert! (->B)))
```

If there are 2 facts of type `A` inserted, and this is fired, the fact `B` would be inserted and supported, with regards to the truth maintenance system (ie. the TMS), by 2 different rule invocations (the 2 different facts of type `A`).

If one of the facts of type `A` were then retracted, the `B` fact should remain, since it still has some TMS support. If the final `A` was then retracted, then `B` would also need to be retracted.

The common way to implement this non-duplication of logically  inserted facts is to do something like a mapping from the `fact -> count of supporting reasons`.
This typically is done via some sort of hash map where the `fact` is the key. It needs to end up being a fairly performant key to do this, which will require some stuff from the engine/memory side.

Drools defaults basically to this behavior for TMS managed inserts. External inserts are a bit different there - they can be duplicated and have no TMS management.

I think the initial issue proposed here wished to have externally inserted facts, also managed in a "set-like" (no duplicates) way. This brings up questions.

I think there is this to consider:
a) If I insert it 5 times, do I need to retract it 5 times to get rid of it?
I think the answer is "no", but it is unfortunate that that differs in semantics from facts inserted in the RHS of a rule.

So I think it ends up being this
(b) No matter how many times the same fact is inserted, it's just considered to be "supported" 1 time, as far as the rest of the engine is concerned when a retraction happens on it.


Also, what happens if you have a fact that you externally insert, that is also inserted by the rules themselves in the RHS.

Taking my initial example, what if I insert `(->B)` from outside the rules, but then fire `my-rule` with some `A`'s present. Now I have a "logically inserted from the RHS" copy of `B`, but also an externally inserted one. How much support does it now have? Can it never be retracted until unless it is externally retracted?

Hopefully this sheds some more thoughts on it and isn't confusing to read through.

blak3mill3r:
@mrrodriguez very helpful, thank you!

I would say instead that there should be a difference between facts inserted "from outside" which are only given 1 support, regardless of multiple insertions of identical facts.

Perhaps internally, for inferred facts in the rules engine, it should still count multiple facts as multiple pieces of supporting evidence ... so different rules could produce two identical facts, and those would be stored as two pieces of supporting evidence.

In which case, perhaps what I want to achieve is best achieved by just putting a set-test before my rules insertions...

blak3mill3r:
IOW managing and storing a set of facts outside `clara.rules` just to protect against double-insertions "from the outside" as opposed to inferred facts via rules, where the existing semantics make sense (now that I have had my eyes opened by this discussion)

PR #439, opened 2019-10-21
Issue 436: Add new node sharing test namespace

WilliamParker:


src/test/common/clara/test_node_sharing.cljc
```diff
@@ -0,0 +1,303 @@
...
+                          (->Cold -20))
+                  fire-rules
+                  (query cold-windy-query))))
+  ;; Note that even though the parent condition isn't met, the WindSpeed will still be evaluated
+  ;; to determine if it meets the constraints since it isn't dependent on the parent condition.
+  ;; It is possible that this would be change in the future if we decided to use lazy evaluation,
+  ;; but in any case it should not be more than 1.
+  (is (= @tu/side-effect-holder 1) "The constraints on WindSpeed should be evaluated exactly once"))
+
+
```
WilliamParker: Todo: eliminate this whitespace.

src/test/common/clara/test_node_sharing.cljc
```diff
@@ -0,0 +1,303 @@
...
+  (let [session (-> empty-session
+                    (insert {:fact-type ::b})
+                    (insert {:fact-type ::d})
+                    fire-rules)]
+
+    (is (= (query session c-query)
+           [{:?c {:fact-type ::c}}]))))
+
+
+;; FIXME: Log an issue for this bug and uncomment when it is resolved.  Since an :or
```
WilliamParker: This fails both before and after https://github.com/cerner/clara-rules/pull/434 so I don't think it is a regression.

WilliamParker:
I've updated the PR according to the todo's I added.

WilliamParker:
@mrrodriguez I'm going to go ahead and merge this, but if you want to discuss changes, additional tests, etc. later I'm happy to do so - I do see that you self-requested a review a while ago.  I just suspect that you're fairly busy right now from offline conversations. :)

PR #445, opened 2020-01-11
implementation of option to blacklist rules when creating sessions.

wandersoncferreira:
The idea is to enable the users to blacklist a specific rule inside a namespace for example.

An example that motivated this implementation was because I have a namespace with many rules defined, however, for some clients I wish to "disable" some specific rules from being loaded into the session.

The workaround I have to do now is to load all the individual rules in the session, instead I wanted to load the entire namespace and only blacklist the ones I wish to remove.

Example of the implemented API:

```clojure
(defrecord Values [id])
  (defrule testing->include
    [?value <- Values [v] (= ?id (:id v))]
    =>
    (println "INCLUDING!!! " ?value))
  (defrule testing->exclude
    [?value <- Values [v] (= ?id (:id v))]
    =>
    (println "EXCLUDING!!! " ?value))
  (def session-without (rl/mk-session 'clara.rules.compiler/testing->include
                                      'clara.rules.compiler/testing->exclude
                                      :blacklist ['clara.rules.compiler/testing->exclude]))
```

This is a working implementation. I don't know all the internals of the library to antecipate more drawbacks that this feature could bring. All the tests are working fine.

### DIFF
--- a/CONTRIBUTORS.md
+++ b/CONTRIBUTORS.md
@@ -21,3 +21,4 @@ Community
 [@sparkofreason]: https://github.com/sparkofreason
 [@bfontaine]: https://github.com/bfontaine
 [@sunilgunisetty]: https://github.com/sunilgunisetty
+[@bartuka]: https://github.com/wandersoncferreira
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -2036,6 +2036,14 @@
          (vary-meta production assoc ::rule-load-order (or n 0)))
        (range) productions))
 
+
+(defn remove-blacklisted-rules
+  "If an option :blacklist was provided by the user, we filter out the rules indicated."
+  [options rules]
+  (let [blacklist (set (map str (:blacklist options)))
+        not-contains? (complement contains?)]
+    (filter #(not-contains? blacklist (:name %)) rules)))
+
 (defn mk-session
   "Creates a new session using the given rule source. The resulting session
   is immutable, and can be used with insert, retract, fire-rules, and query functions."
@@ -2047,6 +2055,7 @@
                           (mapcat #(if (satisfies? IRuleSource %)
                                      (load-rules %)
                                      %))
+                          (remove-blacklisted-rules options)
                           add-production-load-order
                           ;; Ensure that we choose the earliest occurrence of a rule for the purpose of rule order.
                           ;; There are Clojure core functions for distinctness, of course, but none of them seem to guarantee
--- a/src/main/clojure/clara/rules.cljc
+++ b/src/main/clojure/clara/rules.cljc
@@ -338,6 +338,8 @@
         information might prove useful for debugging compilation errors within the rulebase, eg. rulebase serialization
         (ie. via Clara's durability support).
         Defaults to true, see clara.rules.compiler/omit-compile-ctx-default for more information.
+      * :blacklist, a vector of qualified namespace rules that will be blacklisted from the created session.
+        Defaults to [], all the rules provided will be considered.
 
       This is not supported in ClojureScript, since it requires eval to dynamically build a session. ClojureScript
       users must use pre-defined rule sessions using defsession."
--- a/src/test/common/clara/test_blacklist.cljc
+++ b/src/test/common/clara/test_blacklist.cljc
@@ -1,0 +1,107 @@
+#?(:clj
+   (ns clara.test-blacklist
+     (:require [clara.tools.testing-utils :refer [def-rules-test] :as tu]
+               [clara.rules :refer [fire-rules
+                                    insert
+                                    insert!
+                                    query]]
+
+               [clara.rules.testfacts :refer [->Temperature
+                                              ->Cold
+                                              ->Hot
+                                              ->TemperatureHistory
+                                              ->ColdAndWindy]]
+               [clojure.test :refer [is]]
+               [clara.rules.accumulators])
+     (:import [clara.rules.testfacts
+               Temperature
+               Cold
+               Hot
+               ColdAndWindy
+               TemperatureHistory]))
+
+   :cljs
+   (ns clara.test-blacklist
+     (:require [clara.rules :refer [fire-rules
+                                    insert
+                                    insert!
+                                    query]]
+               [clara.rules.testfacts :refer [->Temperature Temperature
+                                              ->Cold Cold
+                                              ->Hot Hot
+                                              ->ColdAndWindy ColdAndWindy
+                                              ->TemperatureHistory TemperatureHistory]]
+               [clara.rules.accumulators]
+               [cljs.test])
+     (:require-macros [clara.tools.testing-utils :refer [def-rules-test]]
+                      [cljs.test :refer [is]])))
+
+(def-rules-test test-blacklist-a-rule
+  {:rules [cold-rule [[[Temperature (< temperature 20) (= ?t temperature)]]
+                      (insert! (->Cold ?t))]
+           history-rule [[[Temperature (< temperature 20) (= ?t temperature)]]
+                         (insert! (->TemperatureHistory [?t]))]]
+
+   :queries [cold-query [[] [[Cold (= ?c temperature)]]]
+             history-query [[] [[TemperatureHistory (= ?t temperatures)]]]]
+
+   :sessions [empty-session [cold-rule cold-query history-rule history-query] {:blacklist ['cold-rule]}]}
+
+  (let [session (-> empty-session
+                    (insert (->Temperature 10 "MCI"))
+                    (fire-rules))]
+
+    (is (empty? (query session cold-query))) ;removed rule
+
+    (is (= #{{:?t [10]}}
+           (set (query session history-query))))))
+
+(def-rules-test test-blacklist-many-rules
+  {:rules [cold-rule [[[Temperature (< temperature 20) (= ?t temperature)]]
+                      (insert! (->Cold ?t))]
+           history-rule [[[Temperature (< temperature 20) (= ?t temperature)]]
+                         (insert! (->TemperatureHistory [?t]))]
+           hot-rule [[[Temperature (> temperature 50) (= ?t temperature)]]
+                     (insert! (->Hot ?t))]
+           cold-and-windy-rule [[[Cold (> temperature 30) (= ?t temperature)]]
+                                (insert! (->ColdAndWindy ?t 60))]]
+
+   :queries [cold-query [[] [[Cold (= ?c temperature)]]]
+             history-query [[] [[TemperatureHistory (= ?ht temperatures)]]]
+             hot-query [[] [[Hot (= ?h temperature)]]]
+             cold-windy-query [[] [[ColdAndWindy (= ?cw temperature)]]]]
+
+   :sessions [empty-session [cold-rule cold-query
+                             history-rule history-query
+                             hot-rule hot-query
+                             cold-and-windy-rule cold-windy-query] {:blacklist ['history-rule
+                                                                                'hot-rule
+                                                                                'cold-and-windy-rule]}]}
+
+  (let [session (-> empty-session
+                    (insert (->Temperature 10 "MCI"))
+                    (insert (->Cold 40))
+                    (insert (->Temperature 70 "MCI"))
+                    (fire-rules))]
+
+    (is (empty? (query session history-query)))
+    (is (empty? (query session hot-query)))
+    (is (empty? (query session cold-windy-query)))
+
+    (is (= #{{:?c 10} {:?c 40}}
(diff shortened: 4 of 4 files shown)


EthanEChristian:
Not really related to the code, but more thinking of maintenance of consumer code(projects that contain rules). 

Looking at the changes they seem pretty straight forward, remove productions from the productions loaded from all namespaces given a list of production names to be removed. Though this makes me wonder, given the scenario:
>I have some superset of rules in a given namespace, two clients use this namespace. One of the clients(Client A) use the entire superset, where as the other(Client B) only uses a subset. Thus I have to define a blacklist for Client B, easy enough. Time goes by and Client A requires maintenance/enhancement, so rules are added to the namespace to achieve this. This would then require me to append any/all new productions to the Blacklist for Client B, which may or may not live close to the namespace itself.

I guess the root of my comment/question lies in the work required to maintain the blacklist when the namespace is ever changing. While it probably wouldn't be too big of a deal to maintain a handful of these "subset" clients, it would become a burden to handle them all in the event that the number was increasing. So in short I question wether it should be Blacklisting vs Whitelisting. I believe the latter is supported, but requires the productions to be provided rather than the namespace.

I don't think the answer to the question/comment above should block this enhancement as it is opt-in functionality, and won't affect consumers that don't opt in.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -2036,6 +2036,14 @@
          (vary-meta production assoc ::rule-load-order (or n 0)))
        (range) productions))
 
+
+(defn remove-blacklisted-rules
+  "If an option :blacklist was provided by the user, we filter out the rules indicated."
+  [options rules]
+  (let [blacklist (set (map str (:blacklist options)))
+        not-contains? (complement contains?)]
+    (filter #(not-contains? blacklist (:name %)) rules)))
```
EthanEChristian: Instead of always filtering the productions regardless of if the option was provided, could we first do a check if the option is provided then filter if it was?

wandersoncferreira: I think it would be better to check, in case we have many rules the time wasted doing nothing can be significant. 

is the following approach reasonable?
```clojure
(if (empty? blacklist) rules (filter...))
```

wandersoncferreira:
@EthanEChristian your comment about blacklist vs whitelist is very relevant. I endup in this situation because of shacky business rules. I have a namespace with "mandatory"-rules to all contracts that is acquired by the company. This namespace has only a small "critical"-rules that *should* be true to 100% of these contracts. However, 99% of time this assumption holds, and I did not want to break the logic of this namespace because of the 1% and this 1% can change any of the let's say 10 must-follow rules I have in the namespace.

So in the end, I could endup defining in 99% of the time each of the mandatory rules one-by-one, because of the 1%.

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: Yes, i think that will work perfectly.

wandersoncferreira: Should I rename the variable `rules` to productions to follow the reasoning of the project?

CONTRIBUTORS.md
```diff
@@ -20,4 +20,5 @@ Community
 [@dgoeke]: https://github.com/dgoeke
 [@sparkofreason]: https://github.com/sparkofreason
 [@bfontaine]: https://github.com/bfontaine
-[@sunilgunisetty]: https://github.com/sunilgunisetty
\ No newline at end of file
+[@sunilgunisetty]: https://github.com/sunilgunisetty
+[@bartuka]: https://github.com/wandersoncferreira
```
EthanEChristian: You should also add a line like line 13. so that your name will be displayed as well

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: Looks good to me

EthanEChristian:
@wandersoncferreira, 
I think that answers my question, in the event that the "subset" clients are the minority case then it definitely makes sense to allow blacklisting. 

Thanks for the clarification.

WilliamParker:
Is your need for this in Clojure or Clojurescript?  My first thought for implementing this in Clojure would be to implement the RuleSource protocol.  Something like ```` (defrecord NamespaceWithRulesBlacklist [ns blacklist] RulesSource (load-rules [] (remove (set blacklist) (load-rules ns))) ````

However this is only in Clojure, unfortunately ClojureScript is more restricted. I'd have to think about if there is a practical analogue there.  I do see the utility of something like this, though I wonder if ClojureScript is likely to have the large rulebases that would make this most useful.  This by itself isn't too complex, but I do worry about a growing core API to support similar cases.  Perhaps also something like

````
:rule-filters [[:blacklist [rulename-1 rulename-2]]
````

in the session options, with the filters applied in order.  This would have the benefit of allowing a single API for additions and combinations of the filters in the future if needed.

Note that when I refer to core API, that doesn't necessarily mean everything in the repo.  I don't personally mind having some additional "ecosystem" in the repo, but there's a cost in complexity every time the core engine, compiler, etc. are grown, as well as additional cognitive burden for users when top-level APIs grow.  That isn't to say we shouldn't ever do these things, just that there is always a cost-benefit calculation to be made.

Apologies for syntax, spelling, etc. errors, I'm on mobile at the moment. :)

**Comment by wandersoncferreira** [comment]
My current need is in the Clojure spectrum only, however I see the benefits of adding this option to ClojureScript as well. I understand your point of growing the core API and I think the option of adding a `:rule-filters` is very interesting from the perspective that it provides space for productive ideas in the future.

it has the same burden for top level API readers and is more open to extensions.

**Comment by mrrodriguez** [comment]
I like the idea of `:rule-filters` although, I think it should be `:production-filters` to be inclusive of possibly queries (they don't need to be special cased). I have more to say on naming at the end of this though.

I think the question is then what can go in that sort of filter. I don't know that I'm a big fan of a "mini-DSL" in it like `[:blacklist [:rule1]]` etc.

I think perhaps better would be to allow 3 possible things:

1) rule structures - filters rules that are `=` to that rule structure
eg.
```clj
;; `rule-1` is a symbol resolving to a `defrule rule-1` sort of rule var in the `*ns*`
:production-filters [rule-1]
```

2) symbols - filters rules with a name `=` to that symbol
```clj
:production-filters ['my.ns/rule-1]
```

3) strings - filters rules with a name `=` to that string
```clj
:production-filters ["my-rule-1"]
```
Note: you can define productions without `defrule` & `defquery`, so there is nothing saying we have to have a ns-bound-symbol naming the rule - and strings are also supported by the schema I believe.

4) regex - filter rules matching the regex
```clj
:production-filters [#"my.ns/rule.*"]
```

----

Points on above:

I'm not sure we need (4) at this point.
When matching with symbols, I'm not sure if we should do the `=` check with `name` on the symbols & the production name. I'm not sure in some cases that we strictly only use symbols vs strings in the production names. It may make sense to normalize both sides of the comparison with `name` first to compare the underlying strings.


A "filter" is Clojure should be semantically similar to the `filter` fn. This notably is a filter on rules to _include_, not _exclude_. So going down this path interestingly doesn't solve the "blacklist" issue very directly.

So I think it may be clearer and easier to work with to have `:production-inclusions` vs `:production-exclusions` as the options we provide.

wandersoncferreira:
@mrrodriguez Your point about the `filter` semantics is true, it should behave similar to Clojure fn. However, I am pro "mini-DSL" when it is well defined and its power is very restricted to some activities. The change in naming from `rule` to `productions` I think is very sane to keep the semantics of the project.

I think we can provide a simpler "mini-DSL" to use filtering from strings, symbols, structure and regexes [I like this one very much =)]

```
:production-filters [[:exclusions [rulename-1 rulename-2]
                             [:inclusions ['my.ns/rule-1 "my.ns/rule-1" #"my.ns/rule.*"]]]

``` 
Or if we want to provide better "operators" to the DSL, a `:not` operator would be nice too instead of "exclusions" for example.

I think the point of providing an expressive DSL over production filtering has the side effect to enable more use-cases for the future. All the development can be done in a different namespace from `compiler` and probably only providing some interface to handle these cases in there.

mrrodriguez:
@wandersoncferreira So it sounds like our opinions differ on the subject of to mini-DSL or not.

I'm still more of a fan, DSL or not, of having `:production-exclusions` and `:production-inclusions` be separate. However, overall I'm just waiting to see if anyone else has any thoughts at this point.

@EthanEChristian and/or @WilliamParker in particular may be interested in weighing in.

wandersoncferreira:
@mrrodriguez as a fact, I don't have a strong opinion though. I was trying to accomodate the previous comments as well, e.g the increase in core API described by @WilliamParker when he suggested the DSL.

I am also waiting for some agreement before changing the implementation =)

"If you don't have enough information and you **can** postpone a decision, is also clever to do so" rsrs

WilliamParker:
My reason for suggesting having the exclusions and inclusions in a single ordered data structure was to allow for having well-defined priority between them.  Otherwise expressing use-cases like "blacklist all rules matching regex A, but include rule B that matches it" and "whitelist all rules matching regex A, but exclude rule B that matches it" would require a new addition to the top-level API.  That said, I share @mrrodriguez 's misgivings about adding new DSLs and their attendant complexity.  My intent was to try to contain the number of new options and/or DSLs we might need later. :) These requirements seem likely to differ between users quite a lot, but I recognize that because of the limitations in the ClojureScript compilation model it is difficult to do this without core support.

Perhaps just having a :filter-rules-fn option that for ClojureScript would be required to be something that could be eval'ed in a Clojure environment would be an option.  This could be a fully-qualified symbol pointing to a function in a Clojure namespace, or an s-expression that could be evaluated with nothing except clojure.core present.  In any individual use-case the requirements for filtering can probably be expressed in a modest amount of code; the complexity here I think comes from trying to support such use-cases in general.  Does this seem reasonable to you @mrrodriguez ?  You've probably spent more time thinking about the ClojureScript compilation model than I have, is this something that could be reliably done with ClojureScript's API contracts?

mrrodriguez:
@WilliamParker 

> My reason for suggesting having the exclusions and inclusions in a single ordered data structure was to allow for having well-defined priority between them. Otherwise expressing use-cases like "blacklist all rules matching regex A, but include rule B that matches it" and "whitelist all rules matching regex A, but exclude rule B that matches it" w

Yeah, I considered the clashes between option issue, but I think trying to facilitate this in via some  a "well-defined priority" is just overly complex. We don't need to facilitate an option to be overly used/abused in my opinion.

In many other clj libs out there with a concept of inclusions/exclusions (pretty common really), it is typically documented that one is prioritized over the other, or that one is done first. I think the semantics for that are clearer than some ordered sequences of inclusion/exclusion rules interleaved by the user.

> Perhaps just having a :filter-rules-fn option that for ClojureScript would be required to be something that could be eval'ed in a Clojure environment would be an option. This could be a fully-qualified symbol pointing to a function in a Clojure namespace, or an s-expression that could be evaluated with nothing except clojure.core present. 

I don't like any option that tries to do clojure-side user-given `eval` things during cljs compilation. (We need to stop doing that in general in the clara cljs impl, as discussed in #388  ) It is only accidental that you can get away with that sometimes. It seems to be an abuse of the compiler leaking the clj environment during macro expansion anyway you try it. `eval` during cljs macroexpansion is not something that it really supports, it's more "accidental" that it sometimes works out (Relates to [this comment](https://github.com/cerner/clara-rules/issues/388#issuecomment-489747471) )

If we are instead restricted to something like described like I suggested with (1)-(4) in [the above comment](https://github.com/cerner/clara-rules/pull/445#issuecomment-573712345) it'd just be normal macroexpansion clj-side logic.
Your option of providing a symbol instead of a form is better than arbitrary form. However, I'd hope it wouldn't need to be fully-qualified. It'd also probably be a little confusing from an API perspective that it was restricted to that. We'd probably at least have to name it `filter-rules-fn-name`. Even then, it's a bit odd from an API perspective.

WilliamParker:
OK fair enough on the issues surrounding ClojureScript eval.  The API change probably doesn't add too much cognitive overhead for users as an option ":production-exclusions" is easily understood even without additional doc once one understands that a production is either a rule or query.  I'm fine with a :production-exclusions option as you outlined @mrrodriguez ; I also agree that we should maintain consistency with clojure.core in the semantic meaning of "filter".  I do suggest that we be thoughtful in writing the doc on this in the session-options so that this is as quickly understood as possible for someone just quickly skimming through the options.

@EthanEChristian  thoughts?

EthanEChristian:
@WilliamParker,
I am good with that, changing the name to `production-exclusions` is probably for the best and i agree we should be clear with the doc of all these opts we are adding

wandersoncferreira:
cool! I will try to submit a PR with the implementation this weekend! Great discussions around how to evolve an API. o/

Issue #444, opened 2019-12-13
Add unused binding warning to the compiler


EthanEChristian:
Per community feedback from the slack channel, It would be helpful if during compilation Clara would warn when a rule bound a variable but didn't use it. This sort of warning would be helpful to  track down minor spelling mistakes or logical mistakes that would cause rules to misbehave.

[add production](https://github.com/cerner/clara-rules/blob/0.20.0/src/main/clojure/clara/rules/compiler.clj#L1181) seems like a likely place that we could add logic to determine unused bindings.

wandersoncferreira:
The idea of this issue is to be able to diferentiate between the two cases below?

```clojure
(rl/defrule rule-testing
    [?cold <- ColdAndWindy (= ?w windspeed) (< temperature 10)
                           (= ?w 20)]
    =>
    (println "All bindings used"))

  (rl/defrule rule-testing-warning
    [?cold <- ColdAndWindy (= ?w windspeed) (< temperature 10)]
    =>
    (println "Warning, ?w not used"))

```

In the `rule-testing-warning` I should print something warning the user that `?w` is not being used, is that right?

EthanEChristian:
Correct. ?cold is technically bound and unused as well

wandersoncferreira:
Do you have any idea about the implementation? I did something potentially ugly depending on how to approach the problem.

I thought of not interfering with anything that is already happening inside `add-production` so, I created a function to receive the production and parse the `:lhs` and `:rhs` as strings and use a simple regex to find for every binding.

```clojure
(defn warning-unused-bindings [production]
    (let [re-bindings #"\?[\w-]+"
          constraints (map str (mapcat :constraints (:lhs production)))
          constraints-bindings (filter some? (map (partial re-find re-bindings) constraints))
          fact-bindings (map name (map :fact-binding (:lhs production)))
          rhs-bindings (re-seq re-bindings (str (:rhs production)))]
      (doseq [[binding freq] (->> fact-bindings
                                  (concat constraints-bindings rhs-bindings)
                                  frequencies)]
        (when (= 1 freq)
          (println (format "WARNING: %s NOT BEING USED" binding))))))
```

If something appears only once in the whole definition, it gets a warning message. There are space for improving the solution with performance concerns, but I will get in the right track before it.

mrrodriguez:
[...]

PR #448, opened 2020-01-17
warning the user at compile time about the unused bindings

wandersoncferreira:
A first solution to #444 , however there are some questions:

i) How would we handle switching it on/off? dynamic var? include a logging library and provide control through log levels?
ii) What will be the default behavior? I think it should be off. People may notice a difference in response time due to additional code that she does not sign up for.
iii) It need an explicit test or does the fact that it is running for all the tests in the suite already good enough?

I was curious to see how the clara-rules project itself would perform on this new flag. I uploaded it to a [webpaste service](http://ix.io/27BP/text) but I can paste it here if someone want it. It has 1k lines in total.

### DIFF
--- a/src/main/clojure/clara/rules/logger.cljc
+++ b/src/main/clojure/clara/rules/logger.cljc
@@ -1,0 +1,25 @@
+(ns clara.rules.logger)
+
+(defn- warn-unused-binding
+  [{:keys [lhs rhs] :as production}]
+  (let [re-binding #"\?[\w-]+"
+        constraints-bindings (->> lhs
+                                  (mapcat :constraints)
+                                  (map (comp (partial re-find re-binding) str))
+                                  (filter some?))
+        fact-bindings (->> lhs
+                           (map :fact-binding)
+                           (filter some?)
+                           (map name))
+        rhs-bindings (re-seq re-binding (str rhs))]
+    (run! (fn [[?bind freq]]
+            (when (= freq 1)
+              (println (format "WARNING: binding %s defined at %s is not being used" ?bind (:name production)))))
+          (->> constraints-bindings
+               (concat fact-bindings rhs-bindings)
+               (filter some?)
+               frequencies))))
+
+(defn warn-unused-bindings!
+  [productions]
+  (run! warn-unused-binding productions))
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -5,6 +5,7 @@
   (:require [clara.rules.engine :as eng]
             [clara.rules.listener :as listener]
             [clara.rules.platform :as platform]
+            [clara.rules.logger :as logger]
             [clara.rules.schema :as schema]
             [clojure.core.reducers :as r]
             [clojure.reflect :as reflect]
@@ -2059,6 +2060,8 @@
                                       (conj! previous new-production)))
                                   (transient #{}))
                           persistent!)]
+
+     (logger/warn-unused-bindings! productions)
 
      (if-let [session (get @session-cache [productions options])]
        session


src/main/clojure/clara/rules/logger.cljc
```diff
@@ -0,0 +1,25 @@
+(ns clara.rules.logger)
```
WilliamParker: A binding in an accumulator constraint will cause accumulation to occur over each value of that binding, so even if it is a new binding that isn't used elsewhere it actually does impact rule semantics.  So these two rules don't behave the same way:

```
(defrule get-current-temperature
  [?current-temp <- newest-temp :from [Temperature (= ?location location)]]
  =>
  (insert! (->CurrentTemperature ?current-temp))
  )
```

```
(defrule get-current-temperature
  [?current-temp <- newest-temp :from [Temperature]]
  =>
  (insert! (->CurrentTemperature ?current-temp))
  )
```

http://www.clara-rules.org/docs/accumulators/ discusses this, specifically the paragraph after the first example.   So I think we need logic that excludes from this check bindings that only appear once, but where that appearance is in an accumulator.

src/main/clojure/clara/rules/logger.cljc
```diff
@@ -0,0 +1,25 @@
...
+
+(defn- warn-unused-binding
+  [{:keys [lhs rhs] :as production}]
+  (let [re-binding #"\?[\w-]+"
```
WilliamParker: I'd suggest using the [is-variable?](https://github.com/cerner/clara-rules/blob/0.20.0/src/main/clojure/clara/rules/compiler.clj#L82) function in the interest of consistency; we could factor it out into say an "analysis-utils" namespace to avoid a cyclic dependency.

src/main/clojure/clara/rules/logger.cljc
mrrodriguez: Ah! I'm glad you pointed this out. I should have paid more attention. Can't this be the case for more than just an accumulator then? This may make the whole effort here need to shift to something within the compiler more than an external pass over the production structures as currently being done. 

I thought this external "linting" approach was fine since we could just ignore nearly all structure and find variables. However, seeing this accumulator case reminds me there are more semantics to perhaps be aware of. The detection may be more "context dependent" than I thought...

mrrodriguez: If we go with this approach, I'd not call the ns "logger".

Something like `analyzer` or `analysis` seems more typical. Logging is a separate concern.

mrrodriguez:
> i) I don't think we want to force a particular log library on users; there was some discussion of this on #425 (comment) in the context of issue 275. If there are more things to log I like the idea of an overall "development mode" where these warnings are turned on and the user supplies a logging implementation, possibly with Clara including implementation(s) for likely choices. This could be as simple as a function that takes a level keyword and a string to log.

Agreed that we should not attempt to supply any sort of default logging infrastructure. If we want to make the logging of warnings/etc pluggable, we likely can just provide some functions via an API option and perhaps default those to stdout.

src/main/clojure/clara/rules/logger.cljc
wandersoncferreira: Interesting corner case, I will write some code to handle the accumulator approach. Maybe if something else appear that is more dependent of context, we can revisit this approach of external linting.

src/main/clojure/clara/rules/logger.cljc
wandersoncferreira: In the example below, `?cold` should get a warning for not being used. But in the production structure, the `:fact-binding` is a `keyword` therefore the definition of `is-variable?` does not work for it. I just verified with some other approach, but it should be considered a variable or not? [thus `is-variable` needs to be extended to handle not only symbols but also keywords]

```clj
(rl/defrule rule-testing
    [?cold <- ColdAndWindy (= ?w windspeed) (< temperature 10)
                           (= ?w 20)]
    =>
    (println "All bindings used"))
```

src/main/clojure/clara/rules/logger.cljc
```diff
@@ -0,0 +1,25 @@
...
+        constraints-bindings (->> lhs
```
EthanEChristian: super nit/style: Could these be replaced by transducers?

Issue #446, opened 2020-01-11
Bean accessors don't account for IndexedPropertyDescriptors


EthanEChristian:
The compiler supports syntactic sugar when accessing fields off a bean-able object. To support this clara uses bean [introspection](https://github.com/cerner/clara-rules/blob/master/src/main/clojure/clara/rules/compiler.clj#L175-L185) to get the field accessors. However, this introspection allows for IndexedPropertyDescriptors to be returned.

IndexedPropertyDescriptors expect `getIndexedReadMethod` instead of `getReadMethod`, thus causing a NullPointerException to be thrown even if the field that is "Indexed" is not being used.  

Currently non indexed fields look something like:
```
(defrule a-rule 
  [SomeBeanableClass (= "someFieldValue" theField)]
  ...)
```

However to support indexing, it would have to look something like:
```
(defrule a-rule 
  [SomeBeanableClass (= "someIndexedFieldValue" (theIndexedField 0))]
  ...)
```
That would mean that the compiler would have to understand the context of the indexed sugar, where today the non-indexed sugar can simply be replace by the appropriate java interop call, ie. 
```
(.theField ?__fact__)
```

I would purpose that we ignore Indexed properties for the time being, as supporting it would convolute the syntax and the parsing of accessors today.

mrrodriguez:
> I would purpose that we ignore Indexed properties for the time being, as supporting it would convolute the syntax and the parsing of accessors today.

I agree with this.

EthanEChristian:
#449 opened for this issue

EthanEChristian:
As #449 has been merged, i am going to close this issue

PR #449, opened 2020-02-01
Issue446: Bean accessors don't account for IndexedPropertyDescriptors

EthanEChristian:
A PR for #446


project.clj
```diff
@@ -78,6 +78,9 @@
                                             (some->> x :ns ns-name str (re-matches (re-pattern (apply str patterns)))))))
                    :generative (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.generative.*")))
                    :performance (fn [x] (some->> x :ns ns-name str (re-matches #"^clara\.performance.*")))}
+
+  ;; Remove any "Test Facts" from being included in the jar.
+  :jar-exclusions [#".*TestFact\.(java|class)"]
```
mrrodriguez: They really shouldn't even be in the "main" src path, but that's a topic for another build related issue I think.

Any reason this was added in this particular PR though?

src/main/java/clara/test/facts/BeanTestFact.java
```diff
@@ -0,0 +1,38 @@
+package clara.test.facts;
+
+/**
+ * A Java Pojo for the express purpose of testing the behavior of compilation and execution of rules.
+ *
+ * This class should not be included in the released artifact, if it did make it into a released artifact it should be
```
mrrodriguez: Why can't it be under `src/test/java` instead? ie, a dev-time-only `:source-paths`?

src/test/clojure/clara/test_java_facts.clj
```diff
@@ -0,0 +1,29 @@
...
+(tu/def-rules-test test-basic-rule
+  {:rules [kansas-rule [[[BeanTestFact
+                          (= ?locs locations)
+                          (some #(= "Kansas" %) ?locs)]]
+                        (rules/insert! "Kansas Exists")]]
+   :queries [string-query [[] [[?s <- String]]]]
+
+   :sessions [empty-session [kansas-rule string-query] {}]}
+
+  (let [locs (make-array String 2)]
```
mrrodriguez: You could use `doto` around `make-array` to do the `aset`'s all in one go and make it seem somewhat more "functional/immutable" in style

src/test/clojure/clara/test_java_facts.clj
```diff
@@ -0,0 +1,29 @@
...
+            [clara.rules :as rules]
+            [clojure.test :refer [is deftest run-tests testing use-fixtures]])
+  (:import [clara.test.facts
+            BeanTestFact]))
+
+;; A test to demonstrate that a Pojo with indexed property accessors can be used as an alpha root in a session,
+;; see https://github.com/cerner/clara-rules/issues/446
+(tu/def-rules-test test-basic-rule
+  {:rules [kansas-rule [[[BeanTestFact
+                          (= ?locs locations)
```
mrrodriguez: So would direct field access on `roadConditions` not work since it has no standard accessor?

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -31,7 +31,8 @@
             Accumulator
             ISystemFact]
            [java.beans
-            PropertyDescriptor]
+            PropertyDescriptor
+            IndexedPropertyDescriptor]
```
mrrodriguez: This new import does not seem like it is needed.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -179,10 +180,13 @@
...
-
-          [(symbol (string/replace (.. property (getName)) #"_" "-")) ; Replace underscore with idiomatic dash.
-           (symbol (str "." (.. property (getReadMethod) (getName))))])))
+                                                    (getPropertyDescriptors)))
+              :let [read-method (.getReadMethod property)]
+              ;; In the event that there the class has an indexed property without a basic accessor we will simply skip
+              ;; the accessor as we will not know how to retrieve the value. see https://github.com/cerner/clara-rules/issues/446
+              :when read-method]
+          [(symbol (string/replace (.getName property) #"_" "-")) ; Replace underscore with idiomatic dash.
+           (symbol (str "." (.getName read-method)))])))
```
WilliamParker: You'll want to be careful that nothing here adds new reflection.  Perhaps compile a session (thus doing evals) with reflection warnings turned on.  This would be a good test to have actually though perhaps a bit out of scope here; as far as I know we don't have a test for this.

src/main/java/clara/test/facts/BeanTestFact.java
WilliamParker: Agreed, I don't understand why this needs to be in the main source path and then removed from the jar afterward rather than just put in a test source path from the start.

src/test/clojure/clara/test_java_facts.clj
WilliamParker: That's what I'd think yes.  A test that an actual invocation of the indexed accessor in a rule for the property without a standard accessor still works might be nice to have and wouldn't be hard to create but isn't crucial.  I realize it should still pass with this implementation, I just tend to prefer having tests be implementation agnostic where possible.

src/test/clojure/clara/test_java_facts.clj
```diff
@@ -0,0 +1,29 @@
+(ns clara.test-java-facts
+  (:require [clara.tools.testing-utils :as tu]
+            [clara.rules :as rules]
+            [clojure.test :refer [is deftest run-tests testing use-fixtures]])
+  (:import [clara.test.facts
+            BeanTestFact]))
+
```
WilliamParker: This test seems like it could go in the test_java namespace as well, but I don't have strong feelings about it.

src/test/clojure/clara/test_java_facts.clj
```diff
@@ -0,0 +1,29 @@
...
+;; A test to demonstrate that a Pojo with indexed property accessors can be used as an alpha root in a session,
+;; see https://github.com/cerner/clara-rules/issues/446
+(tu/def-rules-test test-basic-rule
```
WilliamParker: Note that using def-rules-test isn't really needed here since this presumably isn't going to be tested in ClojureScript, but it doesn't do any harm.

src/test/clojure/clara/test_java_facts.clj
```diff
@@ -0,0 +1,29 @@
...
+    (aset locs 0 "Florida")
+    (aset locs 1 "Kansas")
+    (let [session-strings (into #{}
+                                (map :?s)
+                                (-> empty-session
+                                    (rules/insert (BeanTestFact. locs))
+                                    (rules/fire-rules)
+                                    (rules/query string-query)))]
+      (is (= 1 (count session-strings)))
+      (is (contains? session-strings "Kansas Exists")))))
```
WilliamParker: A slightly more concise option:

```
(let [session-strings (map :?s (-> empty-session
                                    (rules/insert (BeanTestFact. locs))
                                    (rules/fire-rules)
                                    (rules/query string-query)))]
      (is (= ["Kansas"] session-strings)))
```

src/main/java/clara/test/facts/BeanTestFact.java
```diff
@@ -0,0 +1,38 @@
...
+        return locations;
+    }
+    public void setLocations(int pos, String location) {
+        locations[pos] = location;
+    }
+    public String getLocations(int pos){
+        return locations[pos];
+    }
+
+    // Partial Indexed property accessor, ie. no standard accessor
```
WilliamParker: Perhaps add the issue link to this comment as well.

project.clj
EthanEChristian: removed with:
https://github.com/cerner/clara-rules/pull/449/commits/b41ffee72a58f19e94b2169f4e9b7ed35736551c

src/main/java/clara/test/facts/BeanTestFact.java
EthanEChristian: My lein is not that strong, but yeah that works:
https://github.com/cerner/clara-rules/pull/449/commits/b41ffee72a58f19e94b2169f4e9b7ed35736551c

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: cleaned up:
https://github.com/cerner/clara-rules/pull/449/commits/27e120f65dbaa595505e0f858c31921af9124e86

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: Adding 
```
:global-vars {*warn-on-reflection* true}
```
to the project and building doesn't show this as a reflection.
To test that this is correct, i removed the type hint for PropertyDescriptor and i get the following:

> Reflection warning, clara/rules/compiler.clj:179:33 - reference to field getReadMethod on java.lang.Object can't be resolved.
Reflection warning, clara/rules/compiler.clj:183:36 - reference to field getName on java.lang.Object can't be resolved.
Reflection warning, clara/rules/compiler.clj:184:29 - reference to field getName can't be resolved.
Reflection warning, clara/rules/compiler.clj:179:33 - reference to field getReadMethod can't be resolved.
Reflection warning, clara/rules/compiler.clj:183:36 - reference to field getName can't be resolved.
Reflection warning, clara/rules/compiler.clj:184:29 - reference to field getName can't be resolved.

So, i think we are good for this example.

src/main/java/clara/test/facts/BeanTestFact.java
EthanEChristian: added:
https://github.com/cerner/clara-rules/pull/449/commits/2d81c3e5bfcd05b970509ac015065582850ddca2

src/test/clojure/clara/test_java_facts.clj
EthanEChristian: I did this for consistency

src/test/clojure/clara/test_java_facts.clj
EthanEChristian: updated:
https://github.com/cerner/clara-rules/pull/449/commits/1a120651ad7f82c4ca32b6b46a0f60aaf84f94b0

src/test/clojure/clara/test_java_facts.clj
EthanEChristian: done:
https://github.com/cerner/clara-rules/pull/449/commits/1a120651ad7f82c4ca32b6b46a0f60aaf84f94b0

src/test/clojure/clara/test_java_facts.clj
EthanEChristian: added a test demonstrating that a compilation error is thrown when trying to use the standard accessor:
https://github.com/cerner/clara-rules/pull/449/commits/1a120651ad7f82c4ca32b6b46a0f60aaf84f94b0

src/test/clojure/clara/test_java_facts.clj
EthanEChristian: I was on the fence about that, I kind of felt like the test_java namespace was testing the Java apis rather than the usage of java roots in the rulebase and thats why i created a new namespace.

src/test/clojure/clara/test_java_facts.clj
mrrodriguez: Might as well test all that we can test on all platforms.

project.clj
```diff
@@ -6,7 +6,9 @@
   :dependencies [[org.clojure/clojure "1.7.0"]
                  [prismatic/schema "1.1.6"]]
   :profiles {:dev {:dependencies [[org.clojure/math.combinatorics "0.1.3"]
-                                  [org.clojure/data.fressian "0.2.1"]]}
+                                  [org.clojure/data.fressian "0.2.1"]]
+                   :java-source-paths ["src/main/java"
```
mrrodriguez: You don't have to repeat "src/main/java".
`lein` merges collections together if you don't have metadata on them like `:replace` or `:displace`.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -3,17 +3,11 @@
    This is the Clara rules compiler, translating raw data structures into compiled versions and functions.
    Most users should use only the clara.rules namespace."
   (:require [clara.rules.engine :as eng]
-            [clara.rules.listener :as listener]
```
mrrodriguez: Were these just unused ns's to clean up?

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: Yep, doesn't look like reflection.

I think we should just have
```clj
:global-vars {*warn-on-reflection* true}
```
at the top-level of our lein project.clj though in general

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: yes, might as well clean up while im here

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: why not:
https://github.com/cerner/clara-rules/pull/449/commits/dc754f90f170c0ff6b95133443bda1d361a91463

project.clj
EthanEChristian: updated:
https://github.com/cerner/clara-rules/pull/449/commits/dc754f90f170c0ff6b95133443bda1d361a91463

src/main/clojure/clara/rules/compiler.clj
WilliamParker: To be clear, my concern wasn't necessarily with reflection warnings in Clara itself but rather with reflection when invoking accessors on Java objects.  For example:

```
(defrule testrule
    [JavaBeanWithFieldA (= a "some value")]
   =>
  ;; do something
)
```
reflecting when invoking the .getA method behind the scenes.  However in any case if that were happening I'd expect to see more warnings in the Travis build with global reflection warnings enabled - in particular, warnings interspersed through test execution when eval calls build rule networks.  So +1.

src/main/java/clara/test/facts/BeanTestFact.java
WilliamParker: +1

src/test/clojure/clara/test_java_facts.clj
WilliamParker: +1

src/test/clojure/clara/test_java_facts.clj
WilliamParker: +1

src/test/clojure/clara/test_java_facts.clj
WilliamParker: +1

src/main/clojure/clara/rules/compiler.clj
EthanEChristian: Ah, yeah that makes sense. I believe the reason that the rules network fns don't experience reflection is because we add the metadata a bit later:
https://github.com/cerner/clara-rules/blob/1bafa05c2502eb452bdd765c1b45c158725f575e/src/main/clojure/clara/rules/compiler.clj#L390

Issue #237, opened 2016-11-17
AlphaNode should keep information on its type and have a node-id


WilliamParker:
https://github.com/cerner/clara-rules/issues/236 provided motivation for this, but it isn't really the same issue.  Our AlphaNodes in the engine currently don't have node ids; it also isn't possible to automatically inspect them to determine what type they are for.  I propose that they should have both node ids and information on their type.  So something like

````
(defrecord AlphaNode [env children activation node-id type] ...)
````

instead of the current

````
(defrecord AlphaNode [env children activation] ...)
````

The node-id would be the same as for the other nodes.  Regarding the type, suppose we had a rule

````
(defrule cold-rule
    [?c <- [Cold]]
    =>
    (insert! (->LousyWeather)))
````

The AlphaNode would have a type of Cold.

This sort of automatic inspectability would potentially be helpful for https://github.com/cerner/clara-rules/issues/236.  More generally, I think such inspectability and ability to identify a particular node would put us in a better place to write various sorts of tooling and make manual inspection of rule networks during debugging easier.  There would be some cost added to SerDe on rulebases but it should be quite small and in any case I don't think rulebase SerDe time (as opposed to session SerDe time) is particularly sensitive.

EthanEChristian:
#450 has been merged, i will close this issue out

Issue #291, opened 2017-05-09
Add node ID and/or rule name info to compiled RHS action functions


WilliamParker:
[Compile-action](https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/rules/compiler.clj#L389) creates unevaluated code for an anonymous function that executes the RHS of a rule. Since the [current namespace is bound to the namespace from which the rule came when this code is evaluated](https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/rules/compiler.clj#L1372) the class of the evaluated function will reflect this namespace.  However, the class won't indicate which rule this came from, which would be helpful in profiling.  Demonstration:

````
(in-ns 'clojure.core)
clojure.core=> (def fn1 (eval `(fn [~'x] (+ ~'x 1))))
#'clojure.core/fn1
clojure.core=> (def fn2 (eval `(fn meaningful-name [~'x] (+ ~'x 1))))
#'clojure.core/fn2
clojure.core=> (class fn1)
clojure.core$eval10594$fn__10595
clojure.core=> (class fn2)
clojure.core$eval10598$meaningful_name__10599
````

Since the [node-id is available](https://github.com/cerner/clara-rules/blob/0.14.0/src/main/clojure/clara/rules/compiler.clj#L1367) and compile-action is only called in ProductionNode compilation we can just give the anonymous function a name like "production-rhs-(NODE_ID_HERE)", something similar involving the name, or a combination of both.

The benefit of this change would be that it would be easy to tell from profiling e.g. in VisualVM what rule these functions correspond to.  I just had a case where one of these functions showed up as a hotspot.

This is similar to https://github.com/cerner/clara-rules/issues/261 but I think that issue is more for the alpha node activations, for which this fix is blocked by https://github.com/cerner/clara-rules/issues/237, whereas for production RHS actions there isn't anything blocking the change.

WilliamParker:
I think this was addressed by https://github.com/cerner/clara-rules/pull/451/files so I'm going to close it for now but feel free to reopen it if there is more that you want to address here @EthanEChristian .

Issue #261, opened 2017-01-22
User-intelligible class names for anonymous rulebase functions


WilliamParker:
Clojure anonymous functions are compiled into classes with names based on the namespace and the name in the fn body if one is given, but if there is no name in the fn body then the only user-intelligible part of the name is the namespace.  Normally in a Clojure development workflow the class name of a function isn't important, but this class name is what shows up in a profiler snapshot when the function is invoked.  When Clara's eval'ed rule network functions such as alpha node activations are performance hotspots it can thus be difficult to determine in a profiler what rules those function calls correspond to.

````
user=> (class (fn myfunction [a] (+ a 1)))
user$eval10502$myfunction__10503
user=> (class (fn [a] (+ a 1)))
user$eval10506$fn__10507
````

I propose that we give these functions names using node ID(s) as part of the name.  After https://github.com/cerner/clara-rules/pull/201 I believe node-ids should be deterministic.  If this is the case, a user could compare the node ID(s) derived from a function class name in a profiler snapshot whose VM is no longer running with a new instance of the rulebase built with the same commands in order to determine what conditions are causing performance problems.

This could be done for all the evaluated functions, although my particular need is mostly for such information on AlphaNode activations functions at this point.  Note that this is blocked on giving node IDs to AlphaNode for which I have logged work at https://github.com/cerner/clara-rules/issues/237.

rbrush:
Just a thought, but could we encode information about the expression in the class name itself, so this could be useful in scenarios where we can't easily track down node ids? So if there's an alpha expression of ```(< temperature 32)```,  we create a class name alpha__lt_temperature_32.

We'd have to encode everything to valid class names and might have to truncate long expressions, but this could give a good idea of performance simply by looking at the profiles or call stack samples.

MrMikeRodriguez:
It isn't clear to me how there could be encoded a meaningful name like "lt" for less than unless we are special casing certain known functions?  Maybe I was misunderstanding part of that.

Overall on this issue, though, I will say that I've added function names to these inner-compiled functions in Clara's compiler before so that I could troubleshoot better.  I believe I took an approach similar to what Will has said about using a node id.

mrrodriguez:
It isn't clear to me how there could be encoded a meaningful name like "lt" for less than unless we are special casing certain known functions? Maybe I was misunderstanding part of that.

[...]

PR #450, opened 2020-02-04
Issue237: Add type to AlphaNodes for introspection

EthanEChristian:
A quick addition for #237 

@mrrodriguez  @WilliamParker

### DIFF
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -524,7 +524,7 @@
 ;; Record representing alpha nodes in the Rete network,
 ;; each of which evaluates a single condition and
 ;; propagates matches to its children.
-(defrecord AlphaNode [id env children activation]
+(defrecord AlphaNode [id env children activation type]
 
   IAlphaActivate
   (alpha-activate [node facts memory transport listener]
--- a/src/main/clojure/clara/rules/engine.cljc
+++ b/src/main/clojure/clara/rules/engine.cljc
@@ -524,7 +524,7 @@
 ;; Record representing alpha nodes in the Rete network,
 ;; each of which evaluates a single condition and
 ;; propagates matches to its children.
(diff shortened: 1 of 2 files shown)


src/main/clojure/clara/rules/engine.cljc
```diff
@@ -524,7 +524,7 @@
 ;; Record representing alpha nodes in the Rete network,
 ;; each of which evaluates a single condition and
 ;; propagates matches to its children.
-(defrecord AlphaNode [id env children activation]
+(defrecord AlphaNode [id env children activation type]
```
mrrodriguez: So the id part of #237 has already been done?

Also, should it perhaps be `fact-type` to be clearer?

EthanEChristian: yes, and i have updated it to be fact-type

EthanEChristian:
@mrrodriguez,
For the Durability question, I would expect that any type provided would already have to be serializable as we have to serialize the alpha-roots, ie.
```
{fact-type [AlphaNode]}
``` 

It wouldn't be unreasonable to believe that a user could provide a fact-type-fn that produced something that wasn't serializable, but that would have tanked durability before these changes as well. Additionally, i would think that based on the contract(fressian-handlers), that in the event that a fact-type-fn like this was provided then the user should also provide the appropriate handlers.

mrrodriguez:
https://github.com/cerner/clara-rules/pull/450#issuecomment-582134204

@EthanEChristian that makes sense to me. I wasn't positive we were already serializing this same value - but indeed that should be the same thing.

mrrodriguez:
When this is done, does #237 need closed?

EthanEChristian:
Yes, once this goes in i was going to close it.

I was trying to knock this out in an attempt to get better generated function names for #261 and #291

EthanEChristian:
Alright, I am going to merge this and get #237 closed out

PR #451, opened 2020-02-11
#291 and #261: Add names to anonymous rulebase functions

EthanEChristian:
As the title mentions, this is a PR for adding usable names for the anonymous functions generated by the compiler(#291 and #261). 

The main goal here would be for performance trouble shooting, most performance analysis tools should be able to provide the hotspot function names. This enhancement should allow those function names to be use to walk back to the troubled nodes/rules.


EthanEChristian:
I wrote a simple example project:
https://github.com/EthanEChristian/clara-aot

for testing out some simple aot rules. However, i'm not sure how useful this really is in answering your point on class name lengths though, as we are at the whim of the user. We could try and shrink the name i have provided but we would always be subject to exceeding the max length due to how the last portion of the namespace is prepended to the class name. For example:

`split_string$eval7309$ProductionNode_7_ActionExpr__7310.class` 
if a user defined a lengthy namespace we could easily exceed the length.

mrrodriguez:
@EthanEChristian 
https://github.com/cerner/clara-rules/pull/451#issuecomment-585911933

Thanks for adding details. I know it's always arbitrary and dependent on the ns length. My point is for us to do our best to not make it a more _likely_ problem.

It looks like limits have been around 255 characters (not considering longer unicode ones). So if we make our auto-generated portion of names, eg. 50 characters long - we are already taking a good sized chunk out of the max. So it's just something I try to be conscious of.

EthanEChristian:
@mrrodriguez, I have abbreviated the function names. As this is likely only going to matter for troubleshooting or power users i think its ok to think that they should be able to trace the compiler code to find the shortened version of the node type and function type.

Zylox:
> Ideally perhaps we could supply a function in the future that would take a class name and map it back to a rule
this is something i think would have a lot of value. Might make sense to have an issue logged for this.

@EthanEChristian i'm not sure if there is a good place for this but documenting what the node abbreviations are in one spot might be nice. I agree these are for power users but even power users like a reference sheet.

EthanEChristian:
@Zylox, 
rather than allowing the callers to the new `mk-node-fn-name` function specify the abbreviation, perhaps the abbreviation would be managed by a centralized table(map). This should give a decent location for where we can document all of the abbreviations.

Zylox:
Either that or the callers retrieve it from that table if you wanted to maintain the same flow. I like the idea of a lookup table though.

EthanEChristian:
@Zylox @WilliamParker,
I would be lying if i said that we could log an issue and get around to it, so i have added a commit with the added function in the inspect namespace and some tests to accompany it.

project.clj
```diff
@@ -51,6 +51,7 @@
                        {:id "advanced"
                         :source-paths ["src/test/clojurescript" "src/test/common"]
                         :compiler {:output-to "target/js/advanced.js"
+                                   :anon-fn-naming-policy :mapped
```
mrrodriguez: What’s this for?

EthanEChristian: It seems that `:advanced` optimizations by default removes the anonymous function names, i had to provide this new property such that the fn names would remain to properly test the changes.

mrrodriguez: Ah ok. I overlooked some of the fn to name stuff happening in like tests and inspect. I think during inspection the cljs would still have an issue right? 

I’m a little uncertain of the parts of these changes than involve pulling fn names from their fn impl. But I suppose if it seems to work for now. The clj side I’d consider a bit more reliable than the cljs.

src/main/clojure/clara/tools/inspect.cljc
```diff
@@ -300,3 +305,53 @@
         (println "  qualified because")
         (explain-activation explanation "    "))
       (println))))
+
+(let [inverted-type-lookup (zipmap (vals eng/node-type->abbreviated-type)
+                                   (keys eng/node-type->abbreviated-type))]
+  (defn node-fn-name->production-name
```
EthanEChristian: TODO:
Document this function

EthanEChristian:
@mrrodriguez @WilliamParker @Zylox,
If there are no other concerns here, i intend on merging this PR pretty soon.

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1702,6 +1702,18 @@
       (filter (fn [fact] (join-filter-fn token fact fact-bindings {}))
               unfiltered-facts))))
 
+;; This lives here as it is both close to the node that it represents, and is accessible to both clj and cljs
+(def node-type->abbreviated-type
```
mrrodriguez: Curious why this ended up being in this ns instead of compiler?

mrrodriguez: @EthanEChristian answered me offline, only this ns can be reached from cljs currently. That's fine for now.

Issue #454, opened 2020-05-08
Add validate that queries provide the correct parameters


EthanEChristian:
With the default implementation of the session([LocalSession](https://github.com/cerner/clara-rules/blob/0.20.0/src/main/clojure/clara/rules/engine.cljc#L1828)), the way that [query](https://github.com/cerner/clara-rules/blob/0.20.0/src/main/clojure/clara/rules/engine.cljc#L1973-L1988) is implemented it currently expects the caller to provide the parameters that the query expects, without explicitly verifying that they have.

This can lead to odd behavior, as without all parameters provided to the query it will not be able to [find the correct tokens in memory](https://github.com/cerner/clara-rules/blob/0.20.0/src/main/clojure/clara/rules/engine.cljc#L1978)

With this issue, I want to:
1. Add validation that that the parameters provided to the query contain keys only stipulated in the parameters of the query itself otherwise fail
2. Update the documentation[1][2] around the usage of query to be more explicit on the fact that the parameters to a query align to that of the bindings in the query constraints

[1]: https://github.com/cerner/clara-rules/blob/0.20.0/src/main/clojure/clara/rules.cljc#L47-L57
[2]: http://www.clara-rules.org/docs/queries/

PR #455, opened 2020-05-10
[Issue454]: Add validation that queries provide the correct parameters

EthanEChristian:
A PR for #454


EthanEChristian:
todos: 
I need to add clearer explanations on both the clara site and the rule namespace as well, about how queries and the contract of queries are defined.

I am adding this review, to make sure everyone is ok with the changes prior to changing all the doc.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ This is a history of changes to clara-rules.
 # 0.21.0 SNAPSHOT
 * Add names to anonymous functions generated by rule compilation, see [issue 261](https://github.com/cerner/clara-rules/issues/261) and [issue 291](https://github.com/cerner/clara-rules/issues/291)
 * Add alpha node types, see [issue 237](https://github.com/cerner/clara-rules/issues/237)
+* BREAKING: Validate parameters provided to queries exist on query and compilation. see [issue 454](https://github.com/cerner/clara-rules/issues/454)
```
mrrodriguez, line 6: It may be better to not just call it breaking. It is in a sense, but previously you’d get no result basically. Perhaps mention? Or is the issue link enough I guess?

src/main/clojure/clara/rules/engine.cljc
```diff
@@ -1986,6 +1986,9 @@
     (let [query-node (get-in rulebase [:query-nodes query])]
       (when (= nil query-node)
         (platform/throw-error (str "The query " query " is invalid or not included in the rule base.")))
+      (when-not (= (into #{} (keys params)) ;; nil params should be equivalent to #{}
+                   (:param-keys query-node))
+        (platform/throw-error (str "The query " query " was not provided with the correct parameters, expected: " (:param-keys query-node) ", provided: " (set (keys params)))))
```
mrrodriguez, line 1992: Nit : Really long line?

src/test/clojure/clara/test_compiler.clj
```diff
@@ -55,4 +57,22 @@
...
+  (let [;; (defquery a-query
+        ;;   [:?b]
+        ;;   [?c <- ::a-fact-type])
+        query {:lhs [{:type ::a-fact-type
+                      :constraints []
+                      :args []
+                      :fact-binding :?c}]
+               :params #{:?b}
+               :name "a-query"}]
+    (try (com/mk-session [[query]])
```
mrrodriguez, line 73: Why did you need to use the compiler ns internal one for this instead of the top-level clara.rules/mk-session?

src/test/clojure/clara/test_compiler.clj
```diff
@@ -55,4 +57,22 @@
...
+         (catch ExceptionInfo exc
```
WilliamParker, line 74: It looks like you could use [assert-ex-data](https://github.com/cerner/clara-rules/blob/0.20.0/src/main/clojure/clara/tools/testing_utils.cljc#L192) here to reduce brittleness in case this exception was ever caught and wrapped.

src/test/clojure/clara/test_compiler.clj
WilliamParker: +1 I'd suggest using the user-facing function just to have a very slightly higher-level test.  Very minor point though IMO.

WilliamParker:
Re todos: +1 to clarifying the doc, but if the current state is that we're accepting query calls that won't actually return data properly I think throwing an exception on that is fine even prior to that - i.e. clarifying doc is a good idea but I personally don't think it is a blocker to merging this.

WilliamParker:
@EthanEChristian just wondering what your plans are with this one, if you're busy I could perhaps help with getting this over the finish line e.g. updating docs as it is quite close I think.

CHANGELOG.md
WilliamParker: +1 to the changes in https://github.com/cerner/clara-rules/pull/455/commits/b559d93351f34cd058a05435b3a32dafa926bcb9

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -1292,15 +1292,26 @@
...
+                     :bindings ancestor-bindings})
+          ;; else its a query node and we need to validate that the query has at least the bindings
+          ;; specified in the parameters
+          (if (every? ancestor-bindings (:params production))
+            (add-node beta-graph
+                      parent-ids
+                      (create-id-fn)
+                      {:node-type :query
+                       :query production})
+            (throw (ex-info "QueryNode does not contain bindings specified in parameters."
```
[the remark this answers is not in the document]
WilliamParker: +1 to the changes in https://github.com/cerner/clara-rules/pull/455/commits/b559d93351f34cd058a05435b3a32dafa926bcb9

src/test/clojure/clara/test_compiler.clj
WilliamParker: +1 to the changes in https://github.com/cerner/clara-rules/pull/455/commits/b559d93351f34cd058a05435b3a32dafa926bcb9

src/main/clojure/clara/rules/engine.cljc
WilliamParker: Looks like this was resolved by the changes in https://github.com/cerner/clara-rules/pull/455/commits/b559d93351f34cd058a05435b3a32dafa926bcb9

src/test/clojure/clara/test_compiler.clj
WilliamParker: +1 to the changes in https://github.com/cerner/clara-rules/pull/455/commits/b559d93351f34cd058a05435b3a32dafa926bcb9

I'm marking this as resolved @mrrodriguez feel free to change if you disagree.

WilliamParker:
+1 to merging this after the changes in https://github.com/cerner/clara-rules/pull/455/commits/b559d93351f34cd058a05435b3a32dafa926bcb9

PR #458, opened 2020-06-29
Issue 275: Max-cycles listener using a listener-based approach

WilliamParker:


src/main/clojure/clara/tools/loop_detector.cljc
```diff
@@ -0,0 +1,102 @@
...
+
+   on-limit-fn is a 0-arg function that is invoked exactly once when this limit is exceeded.  It can either be
+   a user-provided function or a keyword that indicates a built-in function to use.  Currently supported keywords are:
+
+   :throw-exception - This throws an exception when the limit is reached.  If tracing is enabled, the exception will include
+   the trace.
+
+   :standard-out-warning - This prints a warning to standard out."
+
+  [session max-cycles on-limit-fn]
```
WilliamParker, line 94: I envision ultimately using this as part of a general "development mode" that includes a sensible default for the max-cycles.

src/main/clojure/clara/tools/loop_detector.cljc
```diff
@@ -0,0 +1,102 @@
...
+  []
+  (println "Reached maximum activation group transitions threshhold; an infinite loop is suspected"))
+
+(defn ->invoke-once-fn
+  [wrapped-fn]
+  (let [warned (atom false)]
+    (fn []
+      (when-not @warned
+        (reset! warned true)
+        (wrapped-fn)))))
```
WilliamParker, line 68: This is new from the last PR.  This allows us to restrict user-provided limit functions to a single invocation, unlike the previous one where they would be invoked many times once the limit is exceeded.

WilliamParker:
@mrrodriguez @EthanEChristian were you wanting to look over this before merging?  Happy to wait on this if you need more time and would like to do so but since this is tooling rather than something with direct functional impact and I think we resolved what this should look like in previous reviews both in terms of design and code style ( https://github.com/cerner/clara-rules/pull/425 and https://github.com/cerner/clara-rules/pull/447 ) I'm comfortable with merging.  Let me know though if I'm perceiving a consensus where one doesn't exist. :)

src/main/clojure/clara/tools/loop_detector.cljc
```diff
@@ -0,0 +1,102 @@
...
+
+  [session max-cycles on-limit-fn]
+  
+  (eng/with-listener
+    session
+    (CyclicalRuleListener.
+     (atom 0)
+     max-cycles
+     (->invoke-once-fn (or (on-error-fns on-limit-fn)
+                           on-limit-fn)))))
```
EthanEChristian, line 102: Are we concerned if a user provided `on-limit-fn` as **nil**?

seems like this could sneakily bite someone as it would be lying in wait until a max loop was encountered. Then we would thunk **nil**. Not sure its a big deal, but we could try and prevent it.

EthanEChristian:
Looks good to me @WilliamParker

src/main/clojure/clara/tools/loop_detector.cljc
WilliamParker: Good catch @EthanEChristian , I addressed this in https://github.com/cerner/clara-rules/pull/458/commits/84bcc7051f3ae52575e663ae28758812171f0c72 (as I think you noticed from another comment)

src/main/clojure/clara/tools/loop_detector.cljc
WilliamParker: I replaced this with a delay in https://github.com/cerner/clara-rules/pull/458/commits/26d967ebbd659a669207334b84281a925663ec64

I also noticed that there was a bug where the atom with the cycles count could end up being shared between sessions.  I fixed this and added a test for the scenario.

PR #463, opened 2021-01-06
Add support to specify query binding arguments as symbols

k13gomez:
This pull request adds support to specify query binding arguments as symbols instead of only keywords so that defquery syntax looks closer to function definition syntax.

Example from test case added:
```clojure
(defquery temperature-below-value-using-symbol-arg
  [?value]
  [:temperature [{value :value}] (< value ?value)])
```


EthanEChristian:
Overall i don't see any issues with updating the query contract like this, but perhaps @mrrodriguez or @WilliamParker might.

I do however think it would be nice to update the [doc](https://github.com/cerner/clara-site/blob/c94557b20e9b9b5b80c682b76b90b6597287b505/_docs/queries.md) to include the expectation that the arg of a query is either a keyword or symbol.

k13gomez:
I can open a separate PR to update the [doc](https://github.com/cerner/clara-site/blob/c94557b20e9b9b5b80c682b76b90b6597287b505/_docs/queries.md) as well.

EthanEChristian (approved):
I will try and keep a look out for the PR to clara-site

EthanEChristian (changes requested):
Oops just noticed the test failures

k13gomez:
Yes, I did too, I'm working on fixing them, will let you know when it's good to go.

src/main/clojure/clara/rules/dsl.clj
```diff
@@ -277,7 +277,7 @@
 
          query {:lhs (list 'quote (mapv #(resolve-vars % (destructure-syms %))
                                         conditions))
-                :params (set params)}
+                :params (set (map keyword params))}
```
WilliamParker, line 280: I don't mind allowing symbols as query params, but I suggest that it be restricted to keywords or symbols, not anything that keyword can coerce e.g. a string or something else that it will return nil on and cause problems later.  We probably should have a test validating that this fails as well.

WilliamParker:
This test fails:

````
(deftest test-defquery
  (let [session (-> (mk-session [cold-query])
                    (insert (->Temperature 15 "MCI"))
                    (insert (->Temperature 20 "MCI")) ; Test multiple items in result.
                    (insert (->Temperature 10 "ORD"))
                    (insert (->Temperature 35 "BOS"))
                    (insert (->Temperature 80 "BOS"))
                    fire-rules)]


    ;; Query by location.
    (is (= #{{:?l "BOS" :?t 35}}
           (set (query session cold-query '?l "BOS"))))

    (is (= #{{:?l "MCI" :?t 15} {:?l "MCI" :?t 20}}
           (set (query session cold-query '?l "MCI"))))

    (is (= #{{:?l "ORD" :?t 10}}
           (set (query session cold-query '?l "ORD"))))))
````

That is, in the query operation params are still expected to be keywords.  My snap response right now is that this divergence is probably fine but will give it a bit more thought.

File: src/main/clojure/clara/rules/dsl.clj (hunk 1 above)
Comment (k13gomez) [reply src/main/clojure/clara/rules/dsl.clj]: I've addressed this and also added corresponding tests to ensure the new coerce function works as intended, including a failure test.

**Comment by k13gomez** [comment]
@WilliamParker @EthanEChristian I think I've addressed all the feedback, tests are passing, and all changes have been squashed together.

**Review by EthanEChristian — APPROVED** [review]

**Comment by mrrodriguez** [comment]
> This test fails:
> 
> ```
> (deftest test-defquery
>   (let [session (-> (mk-session [cold-query])
>                     (insert (->Temperature 15 "MCI"))
>                     (insert (->Temperature 20 "MCI")) ; Test multiple items in result.
>                     (insert (->Temperature 10 "ORD"))
>                     (insert (->Temperature 35 "BOS"))
>                     (insert (->Temperature 80 "BOS"))
>                     fire-rules)]
> 
> 
>     ;; Query by location.
>     (is (= #{{:?l "BOS" :?t 35}}
>            (set (query session cold-query '?l "BOS"))))
> 
>     (is (= #{{:?l "MCI" :?t 15} {:?l "MCI" :?t 20}}
>            (set (query session cold-query '?l "MCI"))))
> 
>     (is (= #{{:?l "ORD" :?t 10}}
>            (set (query session cold-query '?l "ORD"))))))
> ```
> 
> That is, in the query operation params are still expected to be keywords. My snap response right now is that this divergence is probably fine but will give it a bit more thought.

I think I'd rather see the output _always_ be keywords. It is a data structure at that point and we should use consistent types there. In general, it'd pretty odd to use symbol-keyed maps in clj anyways, so it'd also be fairly non-idiomatic.

The symbols used in the actual definition of the query are more of a syntactic convenience to look closer to how one may expect things to be from a DSL perspective. I don't think the output needs to directly mirror the syntax used as input.

File: src/main/clojure/clara/rules.cljc [hunk 2]
```diff
@@ -48,13 +48,16 @@
   "Runs the given query with the optional given parameters against the session.
    The optional parameters should be in map form. For example, a query call might be:
 
-   (query session get-by-last-name :last-name \"Jones\")
+   (query session get-by-last-name :?last-name \"Jones\")
```
mrrodriguez, line 51: This was just incorrect docs correct? ie. not a change specific to this PR changes?

mrrodriguez (approved):
https://github.com/cerner/clara-rules/pull/463#issuecomment-759556550 I believe this is how it is implemented (always returning kw keys for output)

So this all looks fine to me now.

src/main/clojure/clara/rules.cljc
k13gomez: yes, the docs were incorrect, so I fixed it.

k13gomez: I also considered that with a few changes we could make it possible to pass parameters as :last-name and have the keyword be converted to `:?last-name`, but that would probably add additional complexity that's just not needed. If that's something useful we can add it in the future.

EthanEChristian:
@WilliamParker
Unless you have additional comments, i am going to try and get this merged pretty soon.

k13gomez:
@EthanEChristian I've opened the docs update, please let me know if you have any suggestions: https://github.com/cerner/clara-site/pull/38

WilliamParker:
I missed this, but basically this and the docs change look fine, +1.

PR #466, opened 2021-03-01
convert mk-session to a function since it does not need to be a macro

k13gomez:
Having mk-session be a function makes it more easy to plug into several session building scenarios, including creating middleware for it to add custom caching or behavior.


k13gomez:
Closing because I just realized `(ns-name *ns*)` only works inside a macro and one can always just instrument `clara.rules.compiler/mk-session` instead.

Issue #238, opened 2016-11-22
Duplicate result bindings in rules should cause an exception


WilliamParker:
The following rule compiles and a session with it executes:

````
(defrule sample-should-fail-rule
                    [?binding <- Cold]
                    [?binding <- Hot]
                    =>
                    (insert! (->LousyWeather)))
````

This repetition of result bindings between different conditions seems fundamentally nonsensical to me.  I found a case in our codebase where we had rules like this being automatically generated.  Obviously this is a defect in our code's use of Clara, but ideally Clara wouldn't accept rules like this even if users mistakenly create them.

devn:
+1

WilliamParker:
As noted on https://github.com/cerner/clara-rules/pull/467#issuecomment-788699735 it looks like the engine did have behaviour, that while arguably undesirable, isn't exactly nonsensical - equality between the duplicated bindings does seem to be enforced.  In terms of linting, I think any enhancements can be added in the clj-kondo hooks added in https://github.com/cerner/clara-rules/pull/482 ; a separate linting layer seems superfluous in my opinion.  I'm closing this issue and the PR https://github.com/cerner/clara-rules/pull/467 but we can discuss if there is disagreement.  In any case the PR has been stale for over 2 years now.

PR #467, opened 2021-03-01
Issue 238: Throw exception when a rule has duplicate result bindings

WilliamParker:


src/test/clojure/clara/test_linter.clj
```diff
@@ -0,0 +1,16 @@
+(ns clara.test-linter
+  (:require [clara.rules :refer :all]
+            [clojure.test :refer :all]
+            [clara.rules.testfacts :refer :all]
+            [clara.rules.dsl :as dsl]
+            [clara.tools.testing-utils :refer [assert-ex-data]])
+  (:import [clara.rules.testfacts
+            Cold Hot]))
+
```
WilliamParker, line 9: I'll add tests with more scenarios e.g. more than one duplicate binding and queries if we like this approach.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -2098,6 +2099,7 @@
                           (mapcat #(if (satisfies? IRuleSource %)
                                      (load-rules %)
                                      %))
+                          linter/validate-productions
```
WilliamParker, line 2102: I thought this should be done in the compiler so that it would catch users providing rules outside the DSL but this could be checked there as well.  Thoughts?

src/main/clojure/clara/linter.cljc
```diff
@@ -0,0 +1,28 @@
+(ns clara.linter
+  (:require [clojure.string :as clj-str]))
+
+
```
WilliamParker, line 4: This doesn't yet handle nested conditions - I'd like to handle that as well but it doesn't necessarily need to be in a first pass IMO.

src/main/clojure/clara/linter.cljc
```diff
@@ -0,0 +1,28 @@
...
+(defn validate-duplicate-result-bindings
+  [productions]
+  (doseq [p productions
+          :let [fact-binding-frequencies (frequencies (into
+                                                       []
+                                                       (comp (filter :fact-binding)
```
mrrodriguez, line 10: Could use `keep` instead of the composition I believe.

src/main/clojure/clara/linter.cljc
```diff
@@ -0,0 +1,28 @@
+(ns clara.linter
+  (:require [clojure.string :as clj-str]))
```
mrrodriguez, line 2: We seem a bit inconsistent with how we alias this ns in this project, but typically it's `string`:

```
clojure/clara/tools/inspect.cljc
21:    [clojure.string :as str])

clojure/clara/rules/engine.cljc
6:            [clojure.string :as string]

clojure/clara/rules/dsl.clj
6:            [clojure.string :as string]

clojure/clara/rules/compiler.clj
8:            [clojure.string :as string]
```

src/main/clojure/clara/rules/compiler.clj
mrrodriguez: Would it have previously been broken to have duplicate bindings or is the addition of this linting a breaking change to existing rules?

I think typically I'd expect linting to be something you can opt in & out of.

That said, I think if it's going to happen somewhere in some automated way, the compiler makes sense to me - instead of dsl.

mrrodriguez (commented):
I think you have a bit more you planned to do. First pass seems ok aside from a few concerns.

WilliamParker:
``` Would it have previously been broken to have duplicate bindings or is the addition of this linting a breaking change to existing rules?
```

It does seem that the engine doesn't crash in this circumstance.  The question is whether it is useful behaviour.  Some samples I tried:

````

(defrule sample-should-fail-rule
         [?binding <- Cold]
         [?binding <- Hot]
         =>
         (insert! (->LousyWeather)))

(defrule sample-run-rule
         [?binding <- Cold]
         [?binding <- Cold]
         =>
         (insert! (->LousyWeather))) 

(defrule should-not-match-rule
         [?binding <- Cold]
         [?binding <- Cold (not (identical? this ?binding))]
         =>
         (insert! (->LousyWeather))) 


(-> (mk-session [sample-run-rule lousy-weather-query]) clara.tools.tracing/with-tracing (insert (->Cold :cold)) fire-rules (query lousy-weather-query))
({:?weather #clara.rules.testfacts.LousyWeather{}})

(-> (mk-session [sample-should-fail-rule lousy-weather-query]) clara.tools.tracing/with-tracing (insert (->Cold :cold) (->Hot :hot)) fire-rules (query lousy-weather-query))
()

(-> (mk-session [should-not-match-rule lousy-weather-query]) clara.tools.tracing/with-tracing (insert (->Cold :cold) (->Hot :hot)) fire-rules (query lousy-weather-query))
()
````

I've always thought of rules as being the Cartesian join of different sets, which would make statements like this either nonsensical or always false (and thus not useful).  That said, so far the examples I've tried seem to be always false i.e. the rule doesn't fire, although I'm fairly confident this wasn't by design.  I suspect it might still be possible to break that behaviour but haven't yet found an example.

If we did want to support the behaviour of having rules like this always not match but run we should probably explicitly test that.  I'm still not sure it is useful behaviour though.

File: src/main/clojure/clara/rules/compiler.clj (hunk 2 above)
Comment (WilliamParker) [reply src/main/clojure/clara/rules/compiler.clj:2102]: Maybe "validation" would be a better name if we end up throwing an exception as here.  That seems more in keeping with checks for correctness as opposed to checks for questionable but valid code, which is what I'd expect from a linter.

File: src/main/clojure/clara/linter.cljc [hunk 6]
```diff
@@ -0,0 +1,28 @@
...
+                cardinalities-set (-> fact-binding-frequencies vals set)]]
+    (when-not (every? #(<= % 1) cardinalities-set)
+      (let [duplicate-bindings (->> fact-binding-frequencies
+                                    (filter (fn [[k v]] (> v 1)))
+                                    (map first))
+            duplicates (->> duplicate-bindings
+                            (map name)
+                            (clj-str/join ","))]
+        (throw (ex-info (str "Duplicate fact-binding in rule or query " (:name p) ", binding names: " duplicates)
+                        {:production p
```
EthanEChristian, line 22: For larger productions this might be a little verbose, would it make more sense to have the production name?

EthanEChristian:
Probably a weird scenario but what is the behavior of:
```
(defrule some-rule
  [:or [?b <- SomeFact] [?b <- SomeOtherFact]]
  =>
  (<do something sketchy with ?b>))
```

Im not even sure what the behavior is today, but i will try and post back

WilliamParker:
Closing as commented on https://github.com/cerner/clara-rules/issues/238

Issue #473, opened 2022-01-18
Automate releases to clojars from release tags


k13gomez:
I'm working on a small PR to automate clojars releases from release tags, release process would be the same as right now (but automated), see steps below:
1. prepare for release by updating changelog and updating release version, commit to `main` branch.
2. create tag to match release version, push tag.
3. build runs for release tag and gets automatically deployed to clojars.
4. prepare for development by updating snapshot version, commit to `main` branch.

k13gomez:
After merging 474 and creating a tag the automated deploy was not triggered, I suspect something wrong with the regex pattern, so I will open another PR when I have a chance to test this on a test repository.

k13gomez:
> > After merging 474 and creating a tag the automated deploy was not triggered (I had to do a manual deploy), I suspect something wrong with the regex pattern, so I will open another PR when I have a chance to test this on a test repository.
> 
> I suppose the auto created tag will need removed?

The tag is still created manually, that part is fine, and since the deploy job did not trigger (but I still wanted to release 0.21.2) I did a manual push; just the auto deploy needs more work, further changes will be needed in the travis config file.

I opened a corresponding PR for the 0.21.2 release docs here: https://github.com/cerner/clara-site/pull/40

k13gomez:
I've opened a new PR which should address the problems with the deploy step.

PR #474, opened 2022-01-18
feat: automate clojars releases for release tags with travis-ci

k13gomez:
Discussed in #473

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -89,4 +89,8 @@
                               [:name "Ryan Brush"]
                               [:url "http://www.clara-rules.org"]]]
   :deploy-repositories [["snapshots" {:url "https://oss.sonatype.org/content/repositories/snapshots/"
-                                      :creds :gpg}]])
+                                      :creds :gpg}]
+                        ["clojars" {:url "https://repo.clojars.org"
+                                    :username :env/clojars_username
+                                    :password :env/clojars_password
+                                    :sign-releases false}]])
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,13 @@ script:
   - lein test
   - lein test :generative
   - lein with-profile dev,recent-clj test
+deploy:
+  provider: script
+  script: lein deploy clojars
+  on:
+    tags: true
+    all_branches: false
+    condition: $TRAVIS_TAG =~ ^(\d+)\.(\d+)\.(\d+)(-.+)?$
 # Workaround for https://github.com/travis-ci/travis-ci/issues/4691
 sudo: required
 jdk:


k13gomez:
This might break a few things while I work on it, just noticed build status didn't update, but this is the overall approach.

project.clj
```diff
@@ -89,4 +89,8 @@
                               [:name "Ryan Brush"]
                               [:url "http://www.clara-rules.org"]]]
   :deploy-repositories [["snapshots" {:url "https://oss.sonatype.org/content/repositories/snapshots/"
-                                      :creds :gpg}]])
+                                      :creds :gpg}]
+                        ["clojars" {:url "https://repo.clojars.org"
+                                    :username :env/clojars_username
+                                    :password :env/clojars_password
+                                    :sign-releases false}]])
```
WilliamParker, line 96: I suppose this isn't completely ideal, but since we (at least I) haven't published the GPG key I used to manually deploy in reality it isn't much different from what we've done before.  Also in practice it would be quite obvious if releases that didn't have a corresponding tag were created and since nobody has ever asked about signature on the jars I'm guessing it isn't checked.  So basically I think this is OK.

project.clj
```diff
@@ -89,4 +89,8 @@
                               [:name "Ryan Brush"]
                               [:url "http://www.clara-rules.org"]]]
   :deploy-repositories [["snapshots" {:url "https://oss.sonatype.org/content/repositories/snapshots/"
-                                      :creds :gpg}]])
+                                      :creds :gpg}]
+                        ["clojars" {:url "https://repo.clojars.org"
+                                    :username :env/clojars_username
```
WilliamParker, line 94: From some quick googling it looks like we'll need these variables to be added using the instructions at https://docs.travis-ci.com/user/environment-variables/#encrypting-environment-variables.  Is that what you're thinking?  I could have missed something of course, I haven't looked at the travis docs in much depth to be honest.

k13gomez: @WilliamParker yes, I have already added the variables from the build configuration in Travis CI console

project.clj
k13gomez: I thought this would be fine as well, if someone has an issue with jar signing in the future we can open a new issue and find a way to get GPG working on the CI environment but I don't think that will be the case.

k13gomez:
@mrrodriguez @EthanEChristian do you have any concerns? Think we can get this approved/merged to start doing releases more easily?

mrrodriguez (approved):
I do not see any issues with this. The jar signing situation is interesting, but sounds like something we can address later if needed.

EthanEChristian (approved):
I don't see any issues with this addition

Issue #264, opened 2017-02-09
Provide better exception when unable to resolve class while deserializing rulebase


EthanEChristian:
When a rulebase or session is serialized with a newer version of rules that may contain new records, if you try and deserialize the rulebase with older version of the rules file that doesn't contain these classes.

In the case described above while deserializing there is a null pointer thrown when trying to resolve the constructor symbol in the [read-record](https://github.com/cerner/clara-rules/blob/master/src/main/clojure/clara/rules/durability/fressian.clj#L135) function.

In this scenario i would expect the deserialization to fail, but provide a better exception when the constructor for the missing record.

Typically this sort of error would effect deserializing a session, however can effect the rulebase when a custom-fact-type function is provided and a record is used as the type in a condition for a rule. ex.

```
(defprotocol IFactType
  (fact-type [x]))

(defprotocol IAncestors
  (custom-ancestors [fact-type]))

(extend-type Object
  IFactType
  (fact-type [x] (type x))
  IAncestors
  (custom-ancestors [x]  (ancestors x)))

(defrecord CustomFact [fact-type value]
  IFactType
  (fact-type [x] fact-type))

(defrecord FactType [name]
  IAncestors
  (custom-ancestors [x]  #{CustomFact}))

(defrecord Outcome [final-value])

(def matching-fact (->FactType "Fact-1"))

(defrule some-rule
  [matching-fact [{:keys [value]}] (= ?value value)]
  =>
  (insert! (->Outcome ?value)))

(defquery get-outcomes
  []
  [Outcome (= ?final-value final-value)])

(def session (-> (mk-session [some-rule get-outcomes]
                             :fact-type-fn fact-type
                             :ancestors-fn custom-ancestors)
               (insert (->CustomFact (->FactType "Fact-1") 23))
               (fire-rules)))
```

WilliamParker:
Note that this error could occur with various scenarios that lead to the record map constructor function not being resolvable.  In this case, my understanding from the issue and offline conversations with @EthanEChristian is that it occurred from serializing with one version of an application consuming Clara and deserializing with a newer version of that application, but it could also occur if, say, the namespace needed simply wasn't required even if the needed namespace and version were on the classpath.  Ideally we'd handle the nil in such a way that a meaningful exception saying something like "Couldn't resolve the needed record constructor function map->YourRecordNameHere."  However, since deserialization involves lots of calls to this record handler any error checking will need to be done in a very performance-sensitive way.

[...]

PR #476, opened 2022-07-19
#264:Provide better exception when unable to resolve class while dese…

EthanEChristian:
…rializing rulebase

https://github.com/cerner/clara-rules/issues/264

### DIFF
--- a/src/main/clojure/clara/rules/durability/fressian.clj
+++ b/src/main/clojure/clara/rules/durability/fressian.clj
@@ -132,7 +132,11 @@
   ([^Reader rdr]
    (read-record rdr nil))
   ([^Reader rdr add-fn]
-   (let [builder (-> (.readObject rdr) resolve deref)
+   (let [try-resolve #(if-let [clazz (resolve %)]
+                        clazz
+                        (throw (ex-info (str "Unable to resolve record symbol: '" % "'")
+                                        {:record-class %})))
+         builder (-> (.readObject rdr) try-resolve deref)
          build-map (.readObject rdr)
          m (read-meta rdr)]
      (cond-> (builder build-map)


src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -132,7 +132,11 @@
   ([^Reader rdr]
    (read-record rdr nil))
   ([^Reader rdr add-fn]
-   (let [builder (-> (.readObject rdr) resolve deref)
+   (let [try-resolve #(if-let [clazz (resolve %)]
```
EthanEChristian, line 135: could be named `resolve-or-fail`

mrrodriguez: That’s used elsewhere I think?

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -132,7 +132,11 @@
...
+                        clazz
+                        (throw (ex-info (str "Unable to resolve record symbol: '" % "'")
```
mrrodriguez, line 137:
```suggestion
                        (throw (ex-info (str "Unable to resolve fact type symbol: '" % "'")
```

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -132,7 +132,11 @@
...
+                                        {:record-class %})))
```
mrrodriguez, line 138:
```suggestion
                                        {:fact-type-class %})))
```

mrrodriguez (commented):
I have minor suggestions.

mrrodriguez (approved):

Zylox (approved):

EthanEChristian:
I plan on merging this this weekend to allow for a little additional time for @WilliamParker if he is able to.

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -132,7 +132,11 @@
   ([^Reader rdr]
    (read-record rdr nil))
   ([^Reader rdr add-fn]
-   (let [builder (-> (.readObject rdr) resolve deref)
+   (let [try-resolve #(if-let [clazz (resolve %)]
+                        clazz
```
WilliamParker, line 136: I don't think this is a class, but rather a Clojure Var for the record constructor function.  From the implementation before this PR:

````
(defn write-record
  "Same as write-with-meta, but with Clojure record support.  The type of the record will
   be preserved."
  [^Writer w tag rec]
  (let [m (meta rec)]
    (.writeTag w tag 3)
    (.writeObject w (record-map-constructor-name rec) true)
    (write-map w rec)
    (if m
      (.writeObject w m)
      (.writeNull w))))
````

Then looking at the current read implementation:

```
(defn read-record
  "Same as read-with-meta, but with Clojure record support.  The type of the record will
   be preserved."
  ([^Reader rdr]
   (read-record rdr nil))
  ([^Reader rdr add-fn]
   (let [builder (-> (.readObject rdr) resolve deref)
         build-map (.readObject rdr)
         m (read-meta rdr)]
     (cond-> (builder build-map)
       m (with-meta m)
       add-fn add-fn))))
````
If the resolved value was a class it wouldn't be possible to invoke it as a Clojure function - note the ```(builder build-map)``` call.

Assuming I'm not missing something, this makes testing easier if you're so inclined, though for this sort of error handling I don't think it is critical.  You could just use [ns-unmap](https://clojuredocs.org/clojure.core/ns-unmap) before deserializing.

File: src/main/clojure/clara/rules/durability/fressian.clj [hunk 5]
```diff
@@ -132,7 +132,11 @@
...
+                        (throw (ex-info (str "Unable to resolve fact type symbol: '" % "'")
+                                        {:fact-type-class %})))
```
WilliamParker, line 138: I don't think this is specific to facts?  It is being used for all Clojure records from what I can see here (admittedly it has been a while since I read through this): https://github.com/cerner/clara-rules/blob/0.21.2/src/main/clojure/clara/rules/durability/fressian.clj#L386

src/main/clojure/clara/rules/durability/fressian.clj
mrrodriguez: This is a good point. So the suggestion (besides testing) is that the resolved symbol isn't a class at all and the message should avoid being worded suggesting that it is. Also, we shouldn't call it `clazz`, but perhaps just `resolved` or `resolved-var` or something like that? It is notable that `deref` is called on it a step later, so it definition can't be a class.

src/main/clojure/clara/rules/durability/fressian.clj
mrrodriguez: I think you are right. It really is a failure to resolve a record builder fn.

src/main/clojure/clara/rules/durability/fressian.clj
EthanEChristian: not in this repo, but it is around ;)

EthanEChristian (commented):
correct terminology

src/main/clojure/clara/rules/durability/fressian.clj
EthanEChristian: corrected:
https://github.com/cerner/clara-rules/pull/476/commits/5252ca54994b2a521897bce2a6d8a2af5b223edf

src/main/clojure/clara/rules/durability/fressian.clj
EthanEChristian: corrected:
https://github.com/cerner/clara-rules/pull/476/commits/5252ca54994b2a521897bce2a6d8a2af5b223edf

EthanEChristian: todo: 
add a test

src/main/clojure/clara/rules/durability/fressian.clj
```diff
@@ -132,7 +132,11 @@
   ([^Reader rdr]
    (read-record rdr nil))
   ([^Reader rdr add-fn]
-   (let [builder (-> (.readObject rdr) resolve deref)
+   (let [try-resolve #(if-let [resolved (resolve %)]
+                        resolved
+                        (throw (ex-info (str "Unable to resolve fact type symbol: '" % "'")
```
WilliamParker, line 137: I don't think this is necessarily a fact, although if the Fressian handlers are being brought in by an IWorkingMemorySerializer implementation they usually will be - the Clara engine nodes have their own handlers anyway: https://github.com/cerner/clara-rules/blob/0.21.2/src/main/clojure/clara/rules/durability/fressian.clj#L393 So this seems OK either way as "fact type symbol" or "record constructor symbol".

WilliamParker (approved):

PR #477, opened 2023-03-15
mk-session docstring: document sources

holyjak:
It wasn't clear from the docstrings that mk-session actually takes rule sources, and in what form. It's better to include the information from http://www.clara-rules.org/docs/ruleforms/ right here.

### DIFF
--- a/src/main/clojure/clara/rules.cljc
+++ b/src/main/clojure/clara/rules.cljc
@@ -311,14 +311,17 @@
   (defmacro mk-session
      "Creates a new session using the given rule sources. The resulting session
       is immutable, and can be used with insert, retract, fire-rules, and query functions.
-
-      If no sources are provided, it will attempt to load rules from the caller's namespace,
-      which is determined by reading Clojure's *ns* var.
+      
+      `sources-and-args` can start with 0+ sources, each being either a namespace symbol or
+      a sequence of rules. If no sources are provided, it will attempt to load rules from the 
+      caller's namespace, which is determined by reading Clojure's *ns* var.
+    
+      Examples: `(mk-session)`, `(mk-session 'my.ns)`, `(mk-session [my-rule])`, `(mk-session 'ns1 [rule2] :cache false)`
 
       This will use rules defined with defrule, queries defined with defquery, and sequences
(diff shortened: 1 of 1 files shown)


EthanEChristian (approved):
Thanks holyjak, the contract on this macro is a bit fiddly and clarification was certainly needed.

src/main/clojure/clara/rules.cljc
```diff
@@ -311,14 +311,17 @@
   (defmacro mk-session
      "Creates a new session using the given rule sources. The resulting session
       is immutable, and can be used with insert, retract, fire-rules, and query functions.
-
-      If no sources are provided, it will attempt to load rules from the caller's namespace,
-      which is determined by reading Clojure's *ns* var.
+      
+      `sources-and-args` can start with 0+ sources, each being either a namespace symbol or
```
mrrodriguez, line 315: I think this is an improvement to the doc string overall.

I do think it's a bit more strict in the doc string than the actual fn is though.

The concept of a "rule source" is protocol-based in clara. The caller could give other data types that represent rule sources.

There is a clara.rules.compiler/IRuleSource protocol that has a few built-in extensions to it for things like clojure.lang.Symbol [as can be seen here](https://github.com/cerner/clara-rules/blob/0.21.2/src/main/clojure/clara/rules.cljc#L274). People can extend it to other things, eg. clojure.lang.Var types if they wanted to (I've done this before). If what is given does not satisfy this protocol, then it is assumed to indeed be a sequence of rule/query (aka "production") data structures. This logic [can be observed here](https://github.com/cerner/clara-rules/blob/0.21.2/src/main/clojure/clara/rules/compiler.clj#L2097).

WilliamParker (approved):
I agree with @mrrodriguez 's comments on IRuleSource implementors being valid alternative arguments and also that this is a net improvement on the documentation.  I also agree with renaming "args" to the more descriptive "sources-and-args".

@holyjak if you want to add that here, feel free, but otherwise I'm inclined to merge this soon unless there are objections.  I'm happy to follow up with a short note about ISourceRule in the docs afterward.  Thanks for pointing out the missing info!

PR #481, opened 2023-05-13
convert to using github actions instead of travis ci

k13gomez:
This PR is to move from travis-ci and start to use github actions instead; its simpler, free, and just works.


mrrodriguez (commented):
Good call!

PR #482, opened 2023-05-13
add clj-kondo support

k13gomez:
I've implemented a clj-kondo hook to lint clara rules, this makes it easier to detect problems during development as well as add clojure-lsp based support for clara rules very easily.

details are briefly discussed in a clj-kondo issue: https://github.com/clj-kondo/clj-kondo/issues/2059

### DIFF
--- a/.clj-kondo/config.edn
+++ b/.clj-kondo/config.edn
@@ -1,0 +1,1 @@
+{:config-paths ["../clj-kondo/clj-kondo.exports/clara/rules"]}
--- a/clj-kondo/clj-kondo.exports/clara/rules/hooks/clara_rules.clj
+++ b/clj-kondo/clj-kondo.exports/clara/rules/hooks/clara_rules.clj
@@ -62,7 +62,7 @@
                    (seq (:children node))
                    (concat token-seq (extract-arg-tokens (:children node)))
 
-                   :else token-seq)) [] node-seq)
+                   :else token-seq)) [(api/token-node 'this)] node-seq)
        (set)
        (sort-by node-value)))
 
@@ -79,7 +79,7 @@
           (symbol? (node-value fact-node))
           [(api/vector-node (vec (extract-arg-tokens condition))) condition]
 
-          :else [(api/vector-node []) condition])
+          :else [(api/vector-node [(api/token-node 'this)]) condition])
         args-binding-set (set (map node-value (:children production-args)))
         prev-bindings-set (->> (mapcat (comp :children first) prev-bindings)
                                (filter binding-node?)
@@ -225,7 +225,7 @@
                                 (vec production-output))))
         fn-node (api/list-node
                   (list
-                    (api/token-node 'clojure.core/fn)
+                    (api/token-node 'fn)
                     input-args
                     production-result))
         new-node (api/map-node
@@ -280,7 +280,7 @@
                               (api/vector-node
                                 (vec production-output))))
         fn-node (api/list-node
-                  (cond-> (list (api/token-node 'clojure.core/fn))
+                  (cond-> (list (api/token-node 'fn))
                     production-docs (concat [production-docs])
                     :always (concat [input-args])
                     production-opts (concat [production-opts])
@@ -330,7 +330,7 @@
                               body-node))
         fn-node (api/list-node
                    (list
-                     (api/token-node 'clojure.core/fn)
+                     (api/token-node 'fn)
                      input-args
                      production-result))
         new-node (api/map-node
@@ -378,7 +378,7 @@
                                 (vec production-output))
                               body-seq))
         fn-node (api/list-node
-                  (cond-> (list (api/token-node 'clojure.core/fn))
+                  (cond-> (list (api/token-node 'fn))
                     production-docs (concat [production-docs])
                     :always (concat [input-args])
                     production-opts (concat [production-opts])
@@ -390,4 +390,49 @@
                        (api/map-node
                          [(api/keyword-node :production) fn-node])))
                    merge {:clj-kondo/ignore [:clojure-lsp/unused-public-var]})]
+    {:node new-node}))
+
+
+(defn analyze-def-rules-test-macro
+  [{:keys [:node]}]
+  (let [[test-name test-params & test-body] (rest (:children node))
+        {:keys [rules
+                queries
+                sessions]} (->> (:children test-params)
+                                (partition 2)
+                                (map (juxt (comp api/sexpr first) last))
+                                (into {}))
+        rules-seq (for [[name-node rule-node] (partition 2 (:children rules))]
+                    [name-node (analyze-parse-rule-macro
+                                 {:node (api/list-node
+                                          (list*
+                                            (api/token-node 'clara.rules.dsl/parse-rule)
+                                            (:children rule-node)))})])
+        query-seq (for [[name-node query-node] (partition 2 (:children queries))]
+                    [name-node (analyze-parse-query-macro
+                                 {:node (api/list-node
+                                          (list*
+                                            (api/token-node 'clara.rules.dsl/parse-query)
+                                            (:children query-node)))})])
+        session-seq (for [[name-node productions-node options-node] (partition 3 (:children sessions))
+                          :let [args-seq (concat (:children productions-node)
+                                                 (:children options-node))]]
+                      [name-node (api/list-node
+                                   (list
+                                     (api/token-node 'clara.rules.compiler/mk-session)
+                                     (api/vector-node
+                                       (vec args-seq))))])
+        args-seq (->> (concat rules-seq query-seq session-seq)
+                      (apply concat))
+        wrap-body (api/list-node
+                    (list*
+                      (api/token-node 'let)
+                      (api/vector-node
+                        (vec args-seq))
+                      test-body))
+        new-node (api/list-node
+                   (list
+                     (api/token-node 'clojure.test/deftest)
+                     test-name
+                     wrap-body))]
     {:node new-node}))
--- a/clj-kondo/clj-kondo.exports/clara/rules/hooks/clara_rules.clj
+++ b/clj-kondo/clj-kondo.exports/clara/rules/hooks/clara_rules.clj
@@ -1,0 +1,393 @@
+(ns hooks.clara-rules
+  (:require [clj-kondo.hooks-api :as api]
+            [clojure.string :as str]
+            [clojure.set :as set]))
+
+(defn node-value
+  [node]
+  (when node
+    (api/sexpr node)))
+
+(defn node-type
+  [node]
+  (when node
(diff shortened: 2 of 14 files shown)


project.clj
```diff
@@ -20,15 +21,19 @@
...
+            "clj-kondo-lint" ["do" ["clj-kondo-deps"] ["clj-kondo" "--lint" "src/main" "--fail-level" "error"]]}
   :codox {:namespaces [clara.rules clara.rules.dsl clara.rules.accumulators
                        clara.rules.listener clara.rules.durability
                        clara.tools.inspect clara.tools.tracing
                        clara.tools.fact-graph]
           :metadata {:doc/format :markdown}}
   :javadoc-opts {:package-names "clara.rules"}
   :source-paths ["src/main/clojure"]
-  :resource-paths []
+  :resource-paths ["clj-kondo"]
```
mrrodriguez, line 36: This seems like it should only be in the dev profile and not the top level where it’ll become part of the deployed jar.

project.clj
```diff
@@ -20,15 +21,19 @@
             [lein-cljsbuild "1.1.7" :exclusions [org.clojure/clojure
                                                  org.clojure/clojurescript]]
             [lein-figwheel "0.5.14" :exclusions [org.clojure/clojure
-                                                 org.clojure/clojurescript]]]
+                                                 org.clojure/clojurescript]]
+            [com.github.clj-kondo/lein-clj-kondo "0.2.4" :exclusions [org.clojure/clojure
+                                                                      org.clojure/clojurescript]]]
+  :aliases {"clj-kondo-deps" ["clj-kondo" "--copy-configs" "--dependencies" "--parallel" "--lint" "$classpath"]
+            "clj-kondo-lint" ["do" ["clj-kondo-deps"] ["clj-kondo" "--lint" "src/main" "--fail-level" "error"]]}
```
mrrodriguez, line 28: Why not lint tests too?

project.clj
```diff
@@ -1,12 +1,13 @@
-(defproject com.cerner/clara-rules "0.22.0-SNAPSHOT"
+(defproject com.cerner/clara-rules "0.21.3-SNAPSHOT"
```
mrrodriguez, line 1: This is an interesting version bump. Curious the rationale.

.clj-kondo/config.edn
```diff
@@ -0,0 +1 @@
+{:config-paths ["../clj-kondo/clj-kondo.exports/clara/rules"]}
```
mrrodriguez, line 1: Why not put the config inline here instead of the indirection?

project.clj
```diff
@@ -20,15 +21,19 @@
             [lein-cljsbuild "1.1.7" :exclusions [org.clojure/clojure
                                                  org.clojure/clojurescript]]
             [lein-figwheel "0.5.14" :exclusions [org.clojure/clojure
-                                                 org.clojure/clojurescript]]]
+                                                 org.clojure/clojurescript]]
+            [com.github.clj-kondo/lein-clj-kondo "0.2.4" :exclusions [org.clojure/clojure
```
mrrodriguez, line 25: I’ve never used a plugin for this. I’ve just seen the main function called directly. Interesting.

project.clj
```diff
@@ -20,15 +21,19 @@
...
+                                                                      org.clojure/clojurescript]]]
+  :aliases {"clj-kondo-deps" ["clj-kondo" "--copy-configs" "--dependencies" "--parallel" "--lint" "$classpath"]
```
mrrodriguez, line 27: It’s not clear to me what this alias is for.

clj-kondo/clj-kondo.exports/clara/rules/hooks/clara_rules.clj
```diff
@@ -0,0 +1,393 @@
+(ns hooks.clara-rules
```
mrrodriguez, line 1: Wow. I’m surprised to see these detailed hooks written instead of just ignoring the forms. Nice.

project.clj
k13gomez: This was done on purpose, in order to export the hooks as built-in with the jar they need to be bundled, and I didn't want to bundle `"resources"` because the JS is being built there and those are excluded on purpose.

Exporting a configuration in a dependency:
https://github.com/clj-kondo/clj-kondo/blob/master/doc/config.md#exporting

project.clj
k13gomez: I think at the time I did this the lint support was relatively small, but it grew, will change back.

.clj-kondo/config.edn
k13gomez: This is inline with export/import functionality, the main point of this PR is not just to lint clara-rules (the project) itself but to add support for linting out of the box for the library. This is how the exported lint hooks are integrated with the project, I followed the export/import guide: https://github.com/clj-kondo/clj-kondo/blob/master/doc/config.md#exporting

project.clj
k13gomez: yeah, I've also only used the main function directly, but since `clara-rules` still uses `leiningen` this was the easiest way to get it setup for CI.

project.clj
k13gomez: this alias takes care of copying the lint dependencies of clara itself and populating the cache with the linted dependencies (for development and CI) before linting clara-rules itself.

clj-kondo/clj-kondo.exports/clara/rules/hooks/clara_rules.clj
k13gomez: Yeah the main value being added with this is not ignoring the forms but actually transforming them to forms that the linter can understand and give you meaningful unresolved-symbol/unresolved-var etc warnings/errors.

project.clj
k13gomez: I'm going to spend some time trying to lint the tests (they have some errors unrelated to clara itself I think, just the result of having developed the code-base without linting it.

k13gomez: @mrrodriguez  alright, so the problem was primarily that I had not added support for `def-rules-test`, I have done that now and everything is being linted with no errors.

clj-kondo/clj-kondo.exports/clara/rules/hooks/clara_rules.clj
mrrodriguez: Yeah I didn’t realize the full scope of this PR until your explanations. Nice. 

I think this makes sense.

k13gomez:
I've tested everything with a SNAPSHOT release, will merge tonight and do a 0.22.0 release to clojars.

clj-kondo/clj-kondo.exports/clara/rules/hooks/clara_rules.clj
WilliamParker: Agreed, this looks pretty extensive and I'll be curious to take a closer look (I've been away for a bit and not looking at this) but it looks good.  To be clear, I don't want to block on my looking at it.

src/main/clojure/clara/rules/durability.clj
```diff
@@ -339,7 +339,8 @@
                         (mem/->RuleOrderedActivation (.-node-id act)
                                                      (index-token internal-seen seen (.-token act))
                                                      (.-activation act)
-                                                     (.-rule-load-order act)))
+                                                     (.-rule-load-order act)
+                                                     false))
```
WilliamParker, line 343: I guess this was broken then?  Going off of my immediate recollection I think this wouldn't matter as long as a session wasn't serialized with activations waiting, which they normally probably aren't (but fixing the bug would still be good).  Sorry, I'll take a closer look within the next few days but a bit pressed for time tonight.  If it is a durability bugfix as well we should update the changelog accordingly.  There might be times when we could have this be true (again, would need to look more closely than I can at the moment) but in any case a value of false should be functionally safe.  Perhaps we should add a test, not necessarily in this PR.

src/test/clojure/clara/test_rules.clj
```diff
@@ -1281,9 +1284,9 @@
                                           [:or
                                            [Cold (= ?temp temperature)
                                             (< temperature 10)]
-                                           [Windspeed (= ?loc location)
+                                           [WindSpeed (= ?loc location)
```
WilliamParker, line 1287: This test was broken then I guess?

k13gomez: Yes, the linter caught that this was broken, the rule in question supposed to fail anyways because it is for `test-unbound-bindings` but the binding was supposed to be wrong, not also the fact type.

src/main/clojure/clara/rules/durability.clj
k13gomez: I think I need to update the log anyways to reflect that clj-kondo support is baked in anyways, thanks for the reminder. The linter caught that the arity of this call was wrong, I think it sort of works anyways because the field in question is a volatile field and the compiler must be treating it differently.

k13gomez: @WilliamParker I'm going to leave this up a bit longer in case you want to review it some more, based on how I understand this works I'm not sure how it ever worked, but somehow it did, I assume the `use-token-identity?` was set to `nil` or `false` by default, but if I try to make this call with only 4 args I get either of these errors:
```
CompilerException java.lang.IllegalArgumentException: No matching ctor found for class clara.rules.memory.RuleOrderedActivation
OR
ArityException Wrong number of args (4) passed to: memory/eval4632/->RuleOrderedActivation--4636
```

clj-kondo/clj-kondo.exports/clara/rules/hooks/clara_rules.clj
k13gomez: Sounds good, I'm happy to make some further improvements later then if you find any issues, but based on dog-fooding this on my own library of rules it is working very nicely.

Matthew-Ielusic:
This is really cool!  But, um, I tried executing these hooks on a legacy codebase and got a bunch of `warning: unused binding this` linter results - I think something in the rules' implementation is introducing a binding that ultimately goes unused (H[ere is a link to a slack thread with slightly more details.](https://clojurians.slack.com/archives/C08TC9JCS/p1684532624461789)) Does you have any advice?

k13gomez:
@Matthew-Ielusic can you open an issue with a minimal example to reproduce it? I have some examples of my own codebase and will be working on a fix but more test cases is always better.

Matthew-Ielusic:
@k13gomez Sure, I'll take some time to identify the smallest example that produces the warning and then open an issue.

k13gomez:
actually I believe I've found the issue @Matthew-Ielusic and opened a PR: https://github.com/cerner/clara-rules/pull/484 but if you do open that issue later we can ensure there aren't other cases of this happening. I'll get this merged soon and do a minor release.

PR #484, opened 2023-05-22
fix lint warning when this binding is not used

k13gomez:
reported here https://github.com/cerner/clara-rules/pull/482 that sometimes there is a lint warning, turns out this was due to having a unused binding when `this` is used in args.


k13gomez:
Warnings count went down, from:

https://github.com/cerner/clara-rules/actions/runs/4972789275/jobs/8898217294
```
linting took 16181ms, errors: 0, warnings: 1127
```
to:
https://github.com/cerner/clara-rules/actions/runs/5041788662/jobs/9041795191
```
linting took 20710ms, errors: 0, warnings: 974
```

Matthew-Ielusic:
Yes, this does solve the issue I raised in https://github.com/cerner/clara-rules/pull/482, thank you!

PR #485, opened 2023-05-31
separate compile-test-handler from test-handler and add env

k13gomez:
- separates compile-test-handler from compile-test, this separates this functionality similar to how it was prior to compile-test being changed into it's current form and helps with advanced profiling techniques using alter-var-root to instrument compiled nodes.
- add env support to compile-test which was noted as something to fix in the future and was fairly simple to add

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 This is a history of changes to clara-rules.
 
+# 0.23.0
+* extract clara.rules.compiler/compile-test-handler from clara.rules.compiler/compile-test
+* add support for `env` inside of test expressions
+
 # 0.22.1
 * fix incorrent lint warning triggered when this binding is not used in clj-kondo hooks
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 This is a history of changes to clara-rules.
 
-# 0.23.0
+# 0.23.0-SNAPSHOT
 * extract clara.rules.compiler/compile-test-handler from clara.rules.compiler/compile-test
 * add support for `env` inside of test expressions
 
--- a/src/main/clojure/clara/macros.clj
+++ b/src/main/clojure/clara/macros.clj
@@ -131,7 +131,8 @@
           :test
           `(eng/->TestNode
             ~id
-            ~(com/compile-test id (:constraints condition))
+            ~(:env beta-node)
+            ~(com/compile-test id (:constraints condition) (:env beta-node))
             ~(gen-beta-network child-ids beta-graph all-bindings))
 
           :accumulator
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -1555,7 +1555,7 @@
    expr-fn-lookup :- schema/NodeFnLookup
    new-bindings :- #{sc/Keyword}]
 
-  (let [{:keys [condition production query join-bindings]} beta-node
+  (let [{:keys [condition production query join-bindings env]} beta-node
 
         condition (if (symbol? condition)
                     (.loadClass (clojure.lang.RT/makeClassLoader) (name condition))
@@ -1609,6 +1609,7 @@
       :test
       (eng/->TestNode
         id
+        env
         (compiled-expr-fn id :test-expr)
         children)
 
--- a/src/main/clojure/clara/rules/compiler.clj
+++ b/src/main/clojure/clara/rules/compiler.clj
@@ -403,7 +403,7 @@
         fn-name (mk-node-fn-name "AlphaNode" node-id "AE")]
 
     `(fn ~fn-name [~(add-meta '?__fact__ type)
-          ~destructured-env] ;; TODO: add destructured environment parameter...
+                   ~destructured-env]
        (let [~@assignments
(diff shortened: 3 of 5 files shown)


WilliamParker (commented):
I don't see any functional problems here, although we should probably have a test.  Although for that matter, I looked for tests where the env is used for other condition types and didn't find any.  Unless I'm missing them somewhere (there are a lot of tests, I'd like to have another look later) perhaps we should have an issue to add those tests.

src/main/clojure/clara/rules/compiler.clj
```diff
@@ -414,19 +414,25 @@
...
 
-    `(let [handler# (fn ~fn-name [~'?__token__]
-                      (let [~@assignments]
-                        (and ~@constraints)))]
-       {:handler handler#
-        :constraints '~constraints})))
+(defn compile-test [node-id constraints env]
+  (let [test-handler (compile-test-handler node-id constraints env)]
+    `(array-map :handler ~test-handler
+                :constraints '~constraints)))
```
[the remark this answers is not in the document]
k13gomez: it was previously wrapped in a `let` binding call, the problem steps from the fact that there is schema validation happening with these expressions that require the output of the `compile-*` functions to be a s-expr, returning a map literal breaks that contract, and this seemed like an easy compromise. Functionally it works the same way as it did before minus the let-binding layer.

k13gomez:
This is a work in-progress, I am adding some tests and ensuring they pass. One thing I have noticed as well is that there are other areas where `env` is actually not passed through and an empty map is passed, I am considering if I should add more extensive tests for uses of `env` and then fix the remaining places where `env` is passed as `{}`

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 This is a history of changes to clara-rules.
 
+# 0.23.0
```
WilliamParker, line 3: Minor: Perhaps change to 0.23.0-SNAPSHOT and then back to 0.23.0 when the release is cut so that anyone reading this isn't confused into thinking 0.23.0 has been released when it hasn't.  This probably doesn't matter if you're planning to release 0.23.0 upon merging this though.

src/main/clojure/clara/rules/compiler.clj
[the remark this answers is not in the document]
WilliamParker: OK, makes sense.

src/main/clojure/clara/macros.clj
```diff
@@ -131,7 +131,8 @@
           :test
           `(eng/->TestNode
             ~id
-            ~(com/compile-test id (:constraints condition))
+            ~(:env beta-node)
+            ~(com/compile-test id (:constraints condition) (:env beta-node))
```
WilliamParker, line 135: I think this would have fixed the ClojureScript compiler error I see in the build that failed? https://github.com/cerner/clara-rules/actions/runs/5136943790/jobs/9244352594

src/test/clojure/clara/test_rules.clj
```diff
@@ -1330,7 +1330,23 @@
...
+
+(deftest test-destructured-test-env-binding
+  (let [rule-output-env (atom nil)
+        rule {:name "clara.test-destructured-binding/test-destructured-test-env-binding"
+              :env {:rule-output rule-output-env} ; Rule environment so we can check its output.
+              :lhs '[{:args [[e a v]]
+                      :type :foo
+                      :constraints [(= e ?entity) (= v ?value)]}
+                     {:constraints [(= ?entity 1) (reset! rule-output ?value)]}]
+              :rhs '(println @rule-output)}]
```
WilliamParker, line 1343: I think we can just have a RHS here that doesn't really do anything.   For example an operation like
```(inc 1)``` or anything similar.  In order to resemble real rules it is probably better to have something to execute though rather than say ```(do nil)``` although that would work too.

 This adds some clutter to the logs of the test build output.

From https://github.com/cerner/clara-rules/actions/runs/5152272254/jobs/9278212444 :
````
lein test clara.test-rules
42
42
````

WilliamParker (approved):
Not sure if this still a draft @k13gomez - I left some minor comments but nothing very major.

WilliamParker (approved):

src/main/clojure/clara/macros.clj
k13gomez: Yes, I had missed this in my initial pass.

src/test/clojure/clara/test_rules.clj
k13gomez: Makes sense, will make this change.

CHANGELOG.md
k13gomez: Will do since I might wait until I also merge #487 to do a release.

k13gomez:
@WilliamParker I also noticed there are actually other places where `env` is not passed through, I am planning to open a separate PR for that, but not in a rush to do it, the places where env isn't passed through are these:
- https://github.com/cerner/clara-rules/blob/main/src/main/clojure/clara/rules/engine.cljc#L673
- https://github.com/cerner/clara-rules/blob/main/src/main/clojure/clara/rules/engine.cljc#L1382

An empty map is passed in both of these cases, I'd like your view on whether this is worth fixing or if it was done in this way by design.

PR #488, opened 2023-06-11
Issue 284: Include faulty expression in compilation exception for mul…

WilliamParker:
…ti-child negations


k13gomez:
Nice, I'm also planning to start a backlog of linter enhancements and add support to catch many of these issues there, instead of at run/compile time.

WilliamParker:
@k13gomez Yes, it would be nice to have this in clj-kondo too, as you say it would catch errors earlier.  However that won't cover cases where Clara is used as a compilation target, for example in more targeted industry-specific DSLs, so this is still useful I think.  I wasn't actually setting out to improve errors specifically though, I was just starting at the beginning of the issue list and working forward to see what we can close and this looked easy to just implement.

k13gomez:
This looks good, I can merge it and include in next release; do you want to add an entry to the change log?

WilliamParker:
Merged with a changelog entry, all done for you to release at your convenience @k13gomez .

PR #491, opened 2024-04-06
Address CVE-2024-22871 and cleanup older patterns

EthanEChristian:


.gitignore
```diff
@@ -17,3 +17,6 @@ figwheel_server.log
 *.iml
 .clj-kondo
 .lsp
+node_modules/*
+package.json
```
mrrodriguez, line 21: May not be a bad idea to check both the package*.json files into gh instead for stability. But won't block on it.

mrrodriguez (approved):
Looks good. I have only 1 smaller comment. The compiler changes are a bit to wrap my head around, but I think I followed along. Perhaps a bit of inline explanation there would be nice.

The removal of the use of an unnecessary atom for the `compile-constraints` is quite nice.

svirl (approved):
Hello from a first-time reviewer. 
Liked the shadowing bit; one less atom is nice.

.gitignore
EthanEChristian: Seems reasonable:
https://github.com/cerner/clara-rules/pull/491/commits/4d0e1dcfaf040da31d4d5c99f63300c2da225412

EthanEChristian:
> Looks good. I have only 1 smaller comment. The compiler changes are a bit to wrap my head around, but I think I followed along. Perhaps a bit of inline explanation there would be nice.
> 
> The removal of the use of an unnecessary atom for the `compile-constraints` is quite nice.

I added some comments here:
https://github.com/cerner/clara-rules/pull/491/commits/fcfc4b1a934d00a82a156ee5ac72f35143657df1

I didn't think commenting on the changes to the TestNode record itself made sense documented inline, so i will cover them here.

Previously the testNode compilation was evaluating a structure like:
```
{:test (fn [] ... ) 
 :constraints (....)}
```
This was odd as the constraints didn't need to be evaluated from what i could see, and were actually adding to bloat in the generated classes, which in turn actually breaks the assertion here:
 https://github.com/cerner/clara-rules/blob/3d3c776dd19f1fc80d50ae13c9d1fc51acc40da0/src/main/clojure/clara/rules/compiler.clj#L1971-L2000
  in the sense that the additional evaluated hash-map was generating its own Static variables within the class files, meaning that the number of functions that could be eval'd in a batch was greatly diminished for some use-cases.

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
```
mrrodriguez, line 2: Not that important since we aren't publishing this for any production use, but I'd typically think it should be a "devDependencies" entry.

mrrodriguez (approved):
Extra comments look good. Nothing blocking from me.

package.json
EthanEChristian: shows my lack of experience outside of the clojure side of things, updated here:
https://github.com/cerner/clara-rules/pull/491/commits/e0baa0547673355230cb7d655e34b17b59a412c3

mrrodriguez (approved):

